PR #116, opened 2023-05-25
Add a `set_seed` function

### LINKED ISSUE #56: Add the possibility to seed the generations
Generative models return samples from a distribution; if no seed is set they should return different results each time they are called. The current caching implementation however does not take the random state into account. It is only based on the other inputs, so two successive runs will return the same value even though no seed has been set. This behavior is unintuitive and should be corrected.

Instead, we should add a mechanism so users can set the seed globally. If the seed is set the cache will be hit at the next run, otherwise a new result will be returned. This is however made slightly complicated by the fact that it is not possible to pass a seed value to some API providers like OpenAI and Anthropic, so they will generate a different value each time they are called when the temperature is different from 0.

I have thought of a possible workaround. We can manually define cache keys in the form:

```
random_seed + model_name + call # + argument values
```

- The `random_seed` can be set globally by the user with `outlines.seed(42)`, otherwise it is drawn at random when requested. This ensures that consecutive runs will not hit the cache when no seed is set.
- `model_name` is here to ensure that we're not hitting the cache if we've changed the model used for generation between two runs.
- `argument_values` so the cache is not hit when the same model is called twice with different arguments.
- `call #` so we can cache calls to closed APIs with non zero temperature and  repeated calls within the same program execution to yield different results.

brandonwillard:
This PR adds a basic `set_seed` function.  It doesn't currently handle the caching mentioned in #56 (more on that later).

- [ ] Test other API providers

Closes #56

### DIFF
--- a/outlines/__init__.py
+++ b/outlines/__init__.py
@@ -1,10 +1,38 @@
 """Outlines is a Generative Model Programming Framework."""
-from outlines.caching import clear_cache, disable_cache, get_cache
+import random
+from contextlib import suppress
+
+import numpy as np
+
+from outlines.caching import clear_cache, disable_cache, enable_cache, get_cache
 from outlines.text import prompt
 
 __all__ = [
     "clear_cache",
     "disable_cache",
     "get_cache",
+    "enable_cache",
     "prompt",
 ]
+
+
+def set_seed(seed: int):
+    """Set a global seed for output generation.
+
+    This highly undesirable approach is taken from
+    https://github.com/huggingface/transformers/blob/118e9810687dd713b6be07af79e80eeb1d916908/src/transformers/trainer_utils.py#L83
+    """
+    random.seed(seed)
+    np.random.seed(seed)
+
+    with suppress(ImportError):
+        import torch
+
+        torch.manual_seed(seed)
+        torch.cuda.manual_seed_all(seed)
+        # ^^ safe to call this function even if cuda is not available
+
+    with suppress(ImportError):
+        import tensorflow
+
+        tensorflow.random.set_seed(seed)
--- a/outlines/caching.py
+++ b/outlines/caching.py
@@ -56,3 +56,11 @@ def clear_cache():
     """Erase the cache completely."""
     global memory
     memory.storage.clear()
+
+
+def enable_cache():
+    """Enable the cache."""
+    global memory
+
+    if isinstance(memory, NoCache):
+        memory = Cache(serializer=JSONSerializer(), storage=LocalFileStorage(cache_dir))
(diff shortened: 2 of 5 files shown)


outlines/__init__.py
```diff
@@ -1,10 +1,38 @@
...
+        import torch
+
+        torch.manual_seed(seed)
+        torch.cuda.manual_seed_all(seed)
+        # ^^ safe to call this function even if cuda is not available
+
+    with suppress(ImportError):
+        import tensorflow
+
+        tensorflow.random.set_seed(seed)
```
brandonwillard, line 38: Importing these just in case a user uses a provider backed by them seems awfully wasteful.  We could delay seeding until we know exactly which provider and backend is being used, but that sounds like it could get complicated.

outlines/caching.py
```diff
@@ -56,3 +56,11 @@ def clear_cache():
...
     global memory
     memory.storage.clear()
+
+
+def enable_cache():
+    """Enable the cache."""
+    global memory
+
+    if isinstance(memory, NoCache):
+        memory = Cache(serializer=JSONSerializer(), storage=LocalFileStorage(cache_dir))
```
brandonwillard, line 66: I left this in from when I was attempting to test the seeding with and without caching in `tests/models/test_hf_transformers.py`; however, it looks like the caching approach might be restricting this sort of dynamic on/off functionality, because the cached [`call_model_generate_method`](https://github.com/normal-computing/outlines/blob/2af0b5ab05182ae86fdd931dd463b1c6dd986136/outlines/models/hf_transformers.py#L65) is constructed at module load time.  If we construct `call_model_generate_method` or apply the `cache` decorator to it [when completers `call` functions are constructed](https://github.com/normal-computing/outlines/blob/2af0b5ab05182ae86fdd931dd463b1c6dd986136/outlines/models/hf_transformers.py#L49), we might be able to remove that restriction.

outlines/__init__.py
```diff
@@ -1,10 +1,38 @@
...
+    np.random.seed(seed)
+
+    with suppress(ImportError):
+        import torch
+
+        torch.manual_seed(seed)
+        torch.cuda.manual_seed_all(seed)
+        # ^^ safe to call this function even if cuda is not available
+
+    with suppress(ImportError):
```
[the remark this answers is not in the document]
brandonwillard: It was in the reference `set_seed` implementation linked above.

rlouf:
I think the new `Sequence` solves this problem, no?

brandonwillard:
> I think the new `Sequence` solves this problem, no?

Yeah, in large part it should.

PR #120, opened 2023-05-26
Vectorize function calls

rlouf:
In this PR I introduce a `outlines.vectorize` function that works similarly to `numpy.vectorize`, and allows to turn functions into functions that take arrays as an input. The major differences with `numpy.vectorize` are:

1. The possibility to batch the execution. Launching too many concurrent calls at the same time, doing inference locally with a batch size too large may not be acceptable and thus we need to batch the parallel calls.
2. It will execute coroutines concurrently for better performance.

This function will be used internally for model calls, so we can pass arrays of arbitrary shapes as inputs and get an array of answers. Usual broadcasting rules apply:

```python

complete = models.text_completion.openai("text-davinci-003")

prompts = np.array([["Hi", "Hello"], ["Bye", "Ciao"]])
answer = complete(prompts)
print(answer.shape)
# (2, 2)

answer_samples = complete(prompts, num_samples=10)
print(answer_samples.shape)
# (2, 2, 10)
```

I made the choice here to execute the coroutines in a new event loop before returning. This allows to keep the synchronous user-facing API for now. Of course, this restricts the potential speedups we get by using `asyncio` and we will eventually need to move towards an explicitly async API. But that would be too big of a change for a single PR.

- [x] Vectorize scalar functions and coroutines (OpenAI API, simple tools)
- [x] Vectorize functions with an arbitrary signature (HF models)
- [x] Vectorize coroutines with an arbitrary signature (OpenAI with several samples)
- [x] Execute coroutines in a new event loop before returning
- [x] Replace the OpenAI calls with async ones and use `vectorize`
- [x] Use `vectorize` for HF model calls

### DIFF
--- a/outlines/base.py
+++ b/outlines/base.py
@@ -1,8 +1,15 @@
 import asyncio
 import functools
 import inspect
+from typing import Callable, Optional
 
 import numpy as np
+from numpy.lib.function_base import (
+    _calculate_shapes,
+    _parse_gufunc_signature,
+    _parse_input_dimensions,
+    _update_dim_sizes,
+)
 
 
 class vectorize:
@@ -19,17 +26,19 @@ class vectorize:
 
     """
 
-    def __init__(self, func, signature=None):
+    def __init__(self, func: Callable, signature: Optional[str] = None):
         self.func = func
         self.signature = signature
         self.is_coroutine_fn = inspect.iscoroutinefunction(func)
 
         functools.update_wrapper(self, func)
 
-        if self.signature is not None:
-            raise NotImplementedError(
-                "Vectorization of non-scalar functions is not implemented yet."
-            )
+        if signature is not None:
+            # Parse the signature string into a Python data structure.
+            # For instance "(m),(s)->(s,m)" becomes `([(m,),(s,)],[(s,m)])`.
+            self._in_and_out_core_dimensions = _parse_gufunc_signature(signature)
+        else:
+            self._in_and_out_core_dimensions = None
 
     def __call__(self, *args, **kwargs):
         """Call the vectorized function."""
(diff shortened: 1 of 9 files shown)


outlines/base.py
```diff
@@ -0,0 +1,204 @@
...
+                )
+
+            outputs = _update_arrays_type(outputs, results)
+
+            for output, result in zip(outputs, results):
+                output[index] = result
+
+        return outputs[0] if nout == 1 else outputs
+
+    def vectorize_call(self, broadcast_shape, flat_args, flat_kwargs):
```
rlouf, line 153: We don't need to duplicate the code between `vectorize_call` and `vectorize_call_coroutine`

outlines/base.py
```diff
@@ -0,0 +1,204 @@
...
+                    np.asarray(x).reshape(broadcast_shape).squeeze()
+                    for x in zip(*outputs)
+                ]
+            )
+        else:
+            outputs = np.asarray(outputs).reshape(broadcast_shape).squeeze()
+
+        return outputs
+
+    def call_with_signature(self, *args, **kwargs):
```
rlouf, line 99: We should document this function for future contributors as this is not completely trivial

outlines/base.py
```diff
@@ -0,0 +1,204 @@
...
+
+            if n_results == 1:
+                results = (results,)
+
+            if nout != n_results:
+                raise ValueError(
+                    f"wrong number of outputs from the function, expected {nout}, got {n_results}"
+                )
+
+            if outputs is None:
```
rlouf, line 138: Perform these operations outside of the loop so we can run coroutine concurrently

rlouf:
Update: Vectorization is currently implemented, although it needs more documentation and tests to make the PR mergeable.

outlines/base.py
```diff
@@ -0,0 +1,163 @@
...
+
+            outputs = await asyncio.gather(*tasks)
+
+            return outputs
+
+        loop = asyncio.new_event_loop()
+        try:
+            outputs = loop.run_until_complete(create_and_gather_tasks())
+        finally:
+            loop.close()
```
brandonwillard, line 245: Same consideration here.

outlines/base.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+
+    def __init__(self, func, signature=None):
+        self.func = func
+        self.signature = signature
+        self.is_coroutine_fn = inspect.iscoroutinefunction(func)
+
+        self.__name__ = func.__name__
+        if hasattr(func, "__doc__"):
+            self.__doc__ = func.__doc__
```
brandonwillard, line 35: This makes me wonder if some `functools.wraps`-based approach would work for this as well.

brandonwillard (approved):
Looks like we could use some tests for OpenAI, but I imagine that's not particularly straightforward.

Aside from that, just some questions/comments; otherwise, looks good to me.

outlines/base.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Thunks have no arguments and can thus be called directly.
+
+        """
+        if self.is_coroutine_fn:
+            loop = asyncio.new_event_loop()
+            try:
+                outputs = loop.run_until_complete(self.func())
+            finally:
+                loop.close()
```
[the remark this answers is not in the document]
rlouf: I think the function that returns the new sequence should be async, but I didn't want to break the API just yet.

rlouf:
> Looks like we could use some tests for OpenAI, but I imagine that's not particularly straightforward.

We need to refactor the code to make it easier to test the code that relies on the OpenAI API. That's the thing that annoys me the most currently.

rlouf:
We might want a slightly different behavior for HuggingFace models: in this case we're better off flattening the input arrays so inference can be run in batch. This does not make this PR any less useful, but probably means that we'll need a different mechanism for these models. Will open an issue once this is merged.

outlines/base.py
rlouf: Replaced these lines with `functools.update_wrapper(self, func)`.

PR #122, opened 2023-05-30
Add HuggingFace embedding consistent with cloud service

dpsimpson:
I've added HuggingFace models to the embeddings options in outlines.

Some things:

- There would be less code if we just used `sentence_transformers`, but this is consistent with what happens in `cloud` right now (a light re-write of that code. Ben's argument for doing manual pooling is to reduce dependency bloat.)
- It adds a `batch_size` parameter to the generated function, in the case where you want to do batching. The default processes the input all in one batch. Would it be better for the default to do batches of 1?
- I haven't tested it with a gpu, but it should be lowered properly if a gpu is available 
- I've written some tests that all pass. ~But for some reason, a lot of the other tests are failing on my machine.~ It turns out I was using Python 3.11, which explains all of this.
- It's been through the pre-commit circus.

lmk what you think needs changing etc.

### DIFF
--- a/outlines/models/hf_transformers.py
+++ b/outlines/models/hf_transformers.py
@@ -465,7 +465,7 @@ def HuggingFaceEmbeddings(model_name: str):
         sentences
             The strings to be embedded
         batch_size
-        The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
+            The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
 
 
         Returns
--- a/outlines/models/hf_transformers.py
+++ b/outlines/models/hf_transformers.py
@@ -462,7 +462,7 @@ def HuggingFaceEmbeddings(model_name: str):
 
         Parameters
         ----------
-        sentences: List[str] | str
+        sentences
             The strings to be embedded
         batch_size: Optional[int]
         The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
--- a/outlines/models/hf_transformers.py
+++ b/outlines/models/hf_transformers.py
@@ -1,5 +1,5 @@
 """Integration with HuggingFace's `transformers` library."""
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Tuple, Union
 
 from outlines.caching import cache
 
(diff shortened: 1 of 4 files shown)


outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,113 @@ def create_type_constraint(
...
+    -------
+    A function that will call run the HuggingFace embedding with the given parameters when
+    passed a prompt. It will use a GPU if there is one available.
+
+    """
+
+    def mean_pooling(model_output, attention_mask):
+        """Pools together the outputs from a sentence embedding model to generate a vector embedding.
+
+        Code modified from https://github.com/normal-computing/cloud/blob/main/embedding-services/main.py
```
brandonwillard, line 449:
```suggestion
```

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,113 @@ def create_type_constraint(
...
+        The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
+
+
+        Returns
+        -------
+        A function that will call run the HuggingFace embedding with the given parameters when
+        passed a prompt. It will use a GPU if there is one available.
+
+        Code modified from https://github.com/normal-computing/cloud/blob/main/embedding-services/main.py
+
```
brandonwillard, line 481:
```suggestion
```

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+        )
+
+    def get_embedding(
+        sentences: Union[List[str], str], *, batch_size: Optional[int] = None
+    ):
+        """Generates embeddings from mean-pooled sentence embeddings from HuggingFace sentence transformers.
+
+        Parameters
+        ----------
+        sentences: List[str] | str
```
rlouf, line 465:
```suggestion
        sentences
```

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+            The strings to be embedded
+        batch_size: Optional[int]
```
rlouf, line 467:
```suggestion
        batch_size
```

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+        The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
```
rlouf, line 468:
```suggestion
            The batch size. If it is not provided, or if a negative value is given, the embeddings will be run as a single batch.
```

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+        embeddings_list = []
+
+        for i in range(0, len(sentences), batch_size):
+            batch_sentences = sentences[i : i + batch_size]
+            encoded_input = tokenizer(
+                batch_sentences,
+                padding=True,
+                truncation=True,
+                return_tensors="pt",
+            )
```
rlouf, line 507: What is the point of batching? Memory constraints?

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+                model_output, encoded_input["attention_mask"]
+            )
+            batch_embeddings = torch.nn.functional.normalize(
+                batch_embeddings, p=2, dim=1
+            )
+
+            embeddings_list.append(batch_embeddings)
+
+        embeddings_tensor = torch.cat(embeddings_list, dim=0)
+        return embeddings_tensor.cpu().numpy()[:n_sentences,]  # check slicing here
```
rlouf, line 525: We should write a test to check the slicing. (what would be after `n_sentences`?)

outlines/models/hf_transformers.py
```diff
@@ -421,3 +421,107 @@ def create_type_constraint(
...
+
+
+        Returns
+        -------
+        A function that will call run the HuggingFace embedding with the given parameters when
+        passed a prompt. It will use a GPU if there is one available.
+
+        """
+        import torch
+        from transformers import AutoModel, AutoTokenizer
```
rlouf, line 478: Is there any kind of check on `model_name` that we can enforce? For text generation this is done on the `transformers` size by calling `AutoModelForCausalLM`. Is there a similar model class in `transformers` that checks that the model is appropriate for the given task?

Maybe even something as simple as making sure that the model name starts with `sentence-transfomers/`.

tests/models/test_hf_transformers.py
```diff
@@ -76,3 +81,27 @@ def test_is_in_multiple_samples():
...
+    model = HuggingFaceEmbeddings(EMBED_MODEL)
+    out = model(["This is a test sentence", "Another test sentence"])
+    assert isinstance(out, np.ndarray)
+    assert out.shape[0] == 2
+    assert out.shape[1] == 384
+
+
+def test_batch_embeddings():
+    model = HuggingFaceEmbeddings(EMBED_MODEL)
+    out = model(["This is a test sentence", "Another test sentence"], batch_size=4)
```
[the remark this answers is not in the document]
dpsimpson: It’s checking that the batching returns the right shape even after padding.

outlines/models/hf_transformers.py
dpsimpson: Also parallelism. Faster to reduce number of sends to gpu

brosand:
> > * There would be less code if we just used `sentence_transformers`, but this is consistent with what happens in `cloud` right now (a light re-write of that code. Ben's argument for doing manual pooling is to reduce dependency bloat.)
> 
> Not an expert, but it looks like the `sentence-transformer` repos contains much more code than just a pooling function. With what models is the code in this PR compatible?

This code is specially built for the sgpt implementation. I would actually vote for moving to `sentence-transformers in the long-term, so I think it's reasonable to do it here (assuming there aren't any reproducibility problems

> > * It adds a `batch_size` parameter to the generated function, in the case where you want to do batching. The default processes the input all in one batch. Would it be better for the default to do batches of 1?
> 
> Has Ben found that chunking batches was necessary? At what scale?

Batch size really depends on hardware and the model that is being used. I think it's probably fine to set a default batch size of 128 or larger since most embedding models are pretty small, but I wouldn't really want to commit to any numbers without more information.

PR #124, opened 2023-06-01
Create masks from regex

rlouf:
This PR adds a function that creates a mask from strings that represent a regex. I also add a aliases for a few frequent use cases:

- floats
- integers
- set of characters

Masks are returned as boolean `np.ndarray`, and can thus be combined using `|` (logical or) and `&` (logical and):

```python
>>> import numpy as np
>>> a = np.array([True, False, True])
>>> b = np.array([True, True, False])
>>> a & b
array([ True, False, False])
>>> a | b
array([True, True, True])
```

I originally used `torch.tensor`s but decided to use `numpy.ndarray`s in the end. While this could allow supporting models written with different frameworks, I am still wondering if the conversion to and from the frameworks would slow down the generation due to copying the content of the arrays.

In the Numpy-> Framework direction, `jax.numpy.asarray` [does not  copy](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.asarray.html) the input is a `numpy.ndarray`, and [neither does](https://pytorch.org/docs/stable/generated/torch.from_numpy.html) `torch.from_numpy`. So we should be able to return NumPy arrays here without adding any overhead.

### DIFF
--- a/outlines/text/masks.py
+++ b/outlines/text/masks.py
@@ -1,0 +1,73 @@
+import re
+from typing import Dict, Iterable
+
+import numpy as np
+
+__all__ = [
+    "create_char_set_mask",
+    "create_float_mask",
+    "create_int_mask",
+    "create_mask_from_regex",
+]
+
+
+def create_mask_from_regex(vocabulary: Dict[str, int], regex: str) -> np.ndarray:
+    """Create a token mask from a regex.
+
+    Parameters
+    ----------
+    vocabulary
+        A dictionary that contains a tokenizer's vocabulary as a map
+        between tokens and their ids.
+    regex
+        The regex that tokens need to respect.
+
+    """
+    program = re.compile(regex)
+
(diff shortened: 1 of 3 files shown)


outlines/text/masks.py
```diff
@@ -0,0 +1,71 @@
...
+
+    mask = np.zeros(len(vocabulary), dtype=np.bool_)
+    for token, token_id in vocabulary.items():
+        if program.match(token) is not None:
+            mask[token_id] = True
+
+    return mask
+
+
+def create_int_mask(vocabulary: Dict[str, int]) -> np.ndarray:
```
rlouf, line 36: Benchmark against using `isdigit()`

rlouf: More generally all `int_mask`, `float_mask` and `char_set` masks should be benchmarked against their equivalent Python implementation.

brandonwillard (approved):

PR #129, opened 2023-06-04
Add retry mechanism for openai API calls

### LINKED ISSUE #115: Add retry mechanism for OpenAI API calls
API calls fail more often than they should; we should add a retry mechanism.

harshkumarchourasia:
Add retrying mechanism for openai api calls. 

Sample use cases
1. `complete = models.text_completion.openai("text-davinci-003", **{"wait_exponential_multiplier":1000, "wait_exponential_max":10000, "stop_max_attempt_number":5})`
2. `complete = models.text_completion.openai("text-davinci-003", **{"wait_random_min":1000, "wait_random_max":10000, "stop_max_attempt_number":5})`

Closes #115

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -184,8 +184,8 @@ def OpenAIEmbeddings(model_name: str):
     """
 
     @functools.partial(outlines.vectorize, signature="()->(s)")
-    def generate(query: str) -> np.ndarray:
-        api_response = call_embeddings_api(model_name, query)
+    async def generate(query: str) -> np.ndarray:
+        api_response = await call_embeddings_api(model_name, query)
         response = api_response["data"][0]["embedding"]
         return np.array(response)
 
@@ -213,8 +213,11 @@ def OpenAIImageGeneration(model_name: str = "", size: str = "512x512"):
 
     """
 
+    def generate(prompt: str, samples: int = 1):
+        return generate_base(prompt, samples)
+
     @functools.partial(outlines.vectorize, signature="(),()->(s)")
-    async def generate(prompt: str, samples: int = 1) -> PILImage:
+    async def generate_base(prompt: str, samples: int) -> PILImage:
         api_response = await call_image_generation_api(prompt, size, samples)
 
         images = []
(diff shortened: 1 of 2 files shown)


brandonwillard (changes requested):
Don't forget to set up `pre-commit` locally.  Also, `retrying` needs to be added to the `dependencies` in `pyproject.toml`.

arunpatro:
I would advise using https://pypi.org/project/backoff/ instead of retying. Backoff seems to have more stars, and I have seen multiple people using it like the official tree-of-thoughts too

brandonwillard:
> I would advise using https://pypi.org/project/backoff/ instead of retying. Backoff seems to have more stars, and I have seen multiple people using it like the official tree-of-thoughts too
> 
> See: https://github.com/ysymyth/tree-of-thought-llm/blob/9f0f986dc30604b83f8d07203cf09b745ada36aa/models.py#L18

Is there a technical/functional/support reason one of these libraries should be used over the other?

arunpatro:
Nothing too technical as of my current knowledge. Just some thoughts based on seeing various implementations of `retrying`. I found another simple and intuitive library - https://github.com/invl/retry (600 stars).  One reason to prefer this one or backoff (2000 stars) is that `retrying` has only 29 stars. 

I feel using backoff or `retry` can create a more intuitive api extension, I can also try creating a PR to compare.

brandonwillard:
> Nothing too technical as of my current knowledge. Just some thoughts based on seeing various implementations of `retrying`. I found another simple and intuitive library - https://github.com/invl/retry (600 stars). One reason to prefer this one or backoff (2000 stars) is that `retrying` has only 29 stars.
> 
> I feel using backoff or `retry` can create a more intuitive api extension, I can also try creating a PR to compare.

Assuming stars as a proxy for something like quality or maturity of the libraries, and that the changes are trivial, they might as well be made here.

rlouf:
I've always thought [tenacity](https://github.com/jd/tenacity) (a fork of `retrying`) was the de facto standard for this, and it seems to be very similar to `backoff`. If there is a technical reason to prefer `backoff`, let's use that. Otherwise we can stick with `tenacity`.

brandonwillard (approved):

PR #130, opened 2023-06-05
Implement Tree of Thoughts using Outlines

arunpatro:
This is a simple implementation of ToT using Outlines for prompt management. Code is inspired from the official implementation from https://github.com/ysymyth/tree-of-thought-llm.


rlouf:
Great! You will have to install [pre-commit](https://pre-commit.com) and run it locally before pushing.

rlouf:
Also please rebase when you're pulling the commits from main instead of merging to keep a clean git history.

arunpatro:
Closing this as this is a little outdated. Will raise another PR with simpler structure, pydantic for Node classes, etc that can represent any generic problem.

PR #131, opened 2023-06-06
Add basic parser-driven masking utilities

brandonwillard:
This PR introduces utilities for parser-driven masking: i.e. a parser is run alongside a language model's (LM) token sampling step and used to filter for grammar-valid next tokens.  `lark` was chosen as the underlying parser library because I have previous experience using it, its design is fairly straightforward, it's pure Python, and it supports arbitrary EBNF-like grammars.

The current approach is very experimental and is essentially driven by some patches to `lark`'s LALR parser.  Those patches allow for incremental additions to a lexed/parsed string, and, more importantly, they allow for partial token matches during scanning.  Partial tokens arise when the string being scanned cuts off without completely constructing/finalizing a token.  

For example, the source string `"def fo"` would normally be scanned into two tokens `def` and `fo`; however, if the next LM-sampled token is `"o("`, the complete sampled source string would be `"def foo("`, and the parsed result would be the tokens `def`, `foo`, `(`.  These two scan (or parse) results conflict, because the first one ends by assuming that `fo` is a completed name token, but the next result&mdash;when starting from the beginning&mdash;contains no such token.  In other words, we would need to backtrack in order to find the "correct" name token (i.e. `foo`).

The approach currently used in this PR doesn't immediately accept (i.e. advance the parse state) scan results from tokens with ambiguities like that.

- [ ] Create a demo that uses the new `Sequence` functionality.
- [x] Use the partial token matching information to improve next-token filtering.
- [ ] We could use a better partial matching approach.  The current one does partial matching for only the first terminal regex in a regex that's the _union_ of many terminals, then it follows up with a loop that checks each terminal regex in that union.
- [ ] See if we can expand regex and multi-character terminal symbols in a context-free grammar (CFG) and create more "granular" CFGs that accept characters at a time.  This would make the generated lexers and parsers immediately compatible with single-character-at-a-time parsing and provide a more general and less hacky solution to our partial matching/parsing problem.  It would also provide a foundation for the pre-computing described below.
    The basic idea is that we can get finite state machines (FSMs) for all the (actually regular) regex and multi-character terminal symbols, and those FSMs can be converted to symbols and rules that themselves can be added to the parent CFG.
- [ ] Look into pre-computing token-based subsets of a vocabulary (e.g. subsets of the vocabulary that are valid [partial] `NAME`/`DEF`/etc. token matches).  
    In general, all the vocabulary tokens/strings could possibly be grouped and classified in such a way that the "run-time" parsing only needs to be performed on much smaller sets.  
    For example, a vocabulary token like `"foo("` could be parsed offline to determine that the `"foo"` part is/could be a [partial] `NAME` token, and the `"("` a subsequent `LPAR` token, of course.  With that information we could create a subset of the vocabulary that is `NAME`-prefixed, then, when the current parse state only permits a transition to a `NAME` token, we can reduce the filtered/checked vocabulary to only that subset (or simply reduce/refine the support in rejection sampling scenarios).
    [Here's a Gist outlining the idea](https://gist.github.com/brandonwillard/4567b124596835c0dec31df7cd094da8).  A `lark` grammar that relies less on non-standard/regular regex features, like [this SQL one](https://github.com/zbrookle/sql_to_ibis/blob/0e9226da42065940ce21439d490f9fcacadc7f92/sql_to_ibis/grammar/sql.lark), could actually work with the very naive approach implemented in that Gist.  Again, we want as much coverage as we can get out of the box, but, if people are willing to tweak a grammar, we could pre-process vocabularies right now.
- [ ] Do something better for the indentation-tracking patch.
- [ ] Can we use tree-sitter instead?  I tried out the Python bindings, but they appeared to be lacking wrt. grammar information and lexer/scanner exposure.

### DIFF
--- a/examples/parsing.py
+++ b/examples/parsing.py
@@ -1,0 +1,108 @@
+"""An example illustrating parser-based masking."""
+import math
+import time
+
+import torch
+from lark import Lark
+from lark.indenter import DedentError
+from lark.lexer import UnexpectedCharacters, UnexpectedToken
+from transformers import (
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    LogitsProcessor,
+    LogitsProcessorList,
+    set_seed,
+)
+
+from outlines.text.parsing import PartialPythonIndenter, copy_parser_state, parse_to_end
+
+revision = None
+checkpoint = "Salesforce/codegen-350M-mono"
+device = "cuda"
+
+tokenizer = AutoTokenizer.from_pretrained(checkpoint)
+
+model = AutoModelForCausalLM.from_pretrained(
+    checkpoint, trust_remote_code=True, revision=revision
+).to(device)
+
+input_text = "def "
+inputs = tokenizer.encode(input_text, return_tensors="pt").to(device)
+
+
+class ParserLogitsProcessor(LogitsProcessor):
+    """Bias invalid token scores according to a running parse state."""
+
+    def __init__(self):
+        pyparser = Lark.open_from_package(
+            "lark",
+            "python.lark",
+            ["grammars"],
+            parser="lalr",
+            postlex=PartialPythonIndenter(),
+            start="file_input",
+        )
+        ip = pyparser.parse_interactive("")
+        self.parser_state = ip.parser_state
+        self.states_stack = [self.parser_state]
+        self.token_seq = None
+        self.token_idx = 0
+
+    def __call__(
+        self, input_ids: torch.LongTensor, scores: torch.FloatTensor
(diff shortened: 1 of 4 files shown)


rlouf (approved):

brandonwillard:
The iterative parsing has been usable for a while (e.g. see the full Python demo) and the tools for partial parsing, pre-parsing, and indexing/caching vocabulary tokens are in place.  I'll merge this as-is so that we can split the remaining work (e.g. bullet points in the description) into separate issues/PRs and focus on them independently.

PR #136, opened 2023-06-06
Add a flag to support various models

arunpatro:
Without `trust_remote_code` set to `true`, we cannot run models like https://huggingface.co/mosaicml/mpt-7b and falcon. HF mentions to add this flag. See here -> https://huggingface.co/docs/transformers/custom_models#using-a-model-with-custom-code


arunpatro:
I accidentally added another file

PR #137, opened 2023-06-07
Pass keyword arguments to `AutoModelForCausalLM`

arunpatro:
Without `trust_remote_code` set to `true`, we cannot run models like https://huggingface.co/mosaicml/mpt-7b and falcon. HF mentions to add this flag. See here -> https://huggingface.co/docs/transformers/custom_models#using-a-model-with-custom-code

### DIFF
--- a/outlines/models/hf_transformers.py
+++ b/outlines/models/hf_transformers.py
@@ -16,6 +16,7 @@ def HuggingFaceCompletion(
     model_name: str,
     max_tokens: Optional[int] = None,
     temperature: Optional[float] = None,
+    **model_kwargs,
 ) -> Callable:
     """Create a function that will call the `generate` method of a `transformers` model.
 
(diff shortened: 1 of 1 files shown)


rlouf:
Should we allow users to pass this keyword argument when initialising the model and use whatever default the `transformers` library uses? This looks like a legit security concern.

arunpatro:
> Should we allow users to pass this keyword argument when initialising the model and use whatever default the `transformers` library uses? This looks like a legit security concern.

I agree with the security concern, but users would definitely want to run latest models like falcon, etc. We can raise a meaningful error too. 

There are other keywords that are relevant which can be passed to the hf model like quantization flags. We should support these features and more `kwargs` because local inference is preffered by community and also because apple silicon is a beast.

rlouf:
> I agree with the security concern, but users would definitely want to run latest models like falcon, etc. We can raise a meaningful error too. 

Unless we have a good reason to not do so we should use the same defaults as the `transformers` library. For [this blog post](https://huggingface.co/blog/falcon):

```python
import transformers
import torch

model = "tiiuae/falcon-7b-instruct"

tokenizer = AutoTokenizer.from_pretrained(model)
pipeline = transformers.pipeline(
    "text-generation",
    model=model,
    tokenizer=tokenizer,
    torch_dtype=torch.bfloat16,
    trust_remote_code=True,
    device_map="auto",
)
```

Beyond the security concerns (which should be taken seriously) the interface should aim to be non-surprising for users of the `transformers` library. 

> There are other keywords that are relevant which can be passed to the hf model like quantization flags. We should support these features and more `kwargs` because local inference is preffered by community and also because apple silicon is a beast. 

We definitely should. I think it shouldn't take a lot of extra work to do it in this PR, at least for the most used options.

outlines/models/hf_transformers.py
```diff
@@ -93,7 +93,7 @@ def call_model_generate_method(
     prompt = list(prompt)
 
     tokenizer = AutoTokenizer.from_pretrained(model_name, padding_size="left")
-    model = AutoModelForCausalLM.from_pretrained(model_name)
+    model = AutoModelForCausalLM.from_pretrained(model_name, trust_remote_code=True)
```
rlouf, line 96: We probably shouldn't set this to `True` by default

rlouf (changes requested):

outlines/models/hf_transformers.py
dgerlanc: This 100% shouldn't be set to `True` by default. As Rémi said, we want to respect the library defaults here.

arunpatro:
What about allowing users to do something like `outlines.trust_remote_code = True` at the top like we use for disabling cache?

rlouf:
The best option imo is to use`**kwargs` for passthrough arguments.

rlouf:
Being implemented in #139, closing.

PR #139, opened 2023-06-09
Add `Transformers` model and `Completion` sequence generation

rlouf:
## Some notes on sequence generation

If we note $\boldsymbol{t} = (t_1, \dots, t_n)$ a sequence of $n$ tokens that represent a string, then we can define the next token $t_{n+1}$ in the sequence as the following random variable:

$$
\begin{align*}
\boldsymbol{\alpha} &= \mathrm{LM}(\boldsymbol{t}, \boldsymbol{\theta})\\
t_{n+1} &\sim \mathrm{Categorical}({\boldsymbol{\alpha}})
\end{align*}
$$

where $\boldsymbol{\theta}$ is the set of trained parameters. This random variable's support is the entire token vocabulary $\mathcal{V}$ (~ $10^3$ to $10^4$ tokens). In the following, the $\mathrm{LM}$ function will typically refer to a deep neural network trained on next-token-completion tasks, but it does not need to be and can be as simple as a function that returns a constant vector with equal values.

This random variable is the basic building block that allows us to sample text sequences. Representing these random variables explicitly is thus the first step in an effort to refactor the generation process in Outlines to make it more flexible, to allows a larger variety of sequence-generating processes and ways to sample them.

Using the same model call we can define many other random variables by manipulating the output logits $\boldsymbol{\alpha}$, randomly or deterministically. A particularly interesting example is when one applies a boolean mask $m$ to restrict the support of the distribution like so:

$$
\begin{align*}
\boldsymbol{\alpha} &= \mathrm{LM}(\boldsymbol{t}, \boldsymbol{\theta})\\
\tilde{\boldsymbol{\alpha}} &= m \odot \boldsymbol{\alpha}\\
t_{n+1} &\sim \mathrm{Categorical}({\tilde{\boldsymbol{\alpha}}})
\end{align*}
$$

This mask can encode some a priori knowledge about what the support should be. For instance:
- We want the random variable to represent a digit;
- We want to exclude the `<EOS>` (end-of-sequence) token from the support;
- We want the random variable to match the `[a-zA-Z]` regular expression.

We can summarize the above in the following notation:

$$
t_{n+1} \sim \mathrm{Token}_m(\boldsymbol{t})
$$

where $\mathrm{Token}_m$ is a random variable with support $\mathcal{V} \backslash \left\\{ i | m_i = 0 \right\\}$ and parametrized by $\boldsymbol{t}$.

## In practice

Generating a text sequence using language models requires:
- A language model that returns next-token logits given an input sequence;
- A model that describes how a sequence is to be generated

### Models

In this PR we introduce the `Transformers` object which is responsible of initializing and calling the model. This implicitly defines the interface that models have with the rest of the library (a parent class will be added). In particular:

- Models accept a NumPy array of token ids and return a NumPy array that contains the logits. The NumPy interface simplifies the rest of the library, and we make the assumption that transferring these arrays back-and-forth between the CPU and the device is not the limiting factor in performance.
- Caching is handled at the model level. We need to cache and persist input ids to logit calls (for repeating call of the same workflow during development), and the model's k-v cache for repeated calls with overlapping prompts.

The k-v cache will require some careful thought, and will likely have to be customized to the particular class of models for which it is implemented (`transformers`, `llama.cpp`, etc.). My initial though is to build a trie that we query each time the model is called; this cache is attached to the model instance.

Another difficulty arises when the workflow uses different local models, since we cannot hold all of the models' weights in memory. We may need a process that supervises the models so that when a model is called we know whether it is currently loaded in memory.

We don't necessarily need to solve both these problems now, and they can be turned into issues.

### Sequences

Sequences are objects that represent a sequence-generation model; when called with a prompt or a list of prompt they generate the sequence(s). Here we implement the simplest possible sequence, completion until an EOS token is found. The proposed API is as follows:

```python
import outlines.text as text

completion = text.completion(model)("say something")
new_completion = text.completion(model)(completion)
```

When sampling generations we can simply return a (list of) string(s), but for more advanced generation mechanisms we will need to return a state object that contains both the completion and the corresponding probabilities. This also means that we will need to unpack/repack the state when calling python functions in the middle of a chain.

### Local model and API calls

Models based on API calls are less flexible; for instance we cannot shape the proposals completely (when we can) and don't have access to the logits. They have a different interface and will thus need to inherit from a different base class.

This also means that the `Sequence` implementation needs to be different; to make this transparent to the user we implement `completion` function which dispatches to `Completion` when the logits are available, and to a custom implementation when it is not. It can possibly fail when some generation constraints cannot be applied when calling the API.


rlouf:
Left to implement:

## Models

- [ ] Adding cache to the `__call__` method and persist to disk
- [ ] Add cache for kv values
- [ ] Abstract tokenizer calls to unify interface and add `TransformersTokenizer` class

## Completion

- [ ] Add `Sequence` parent class, which contains the `step` and `__call__` implementations
- [ ] Test `vectorized_choice`
- [ ] Test `Completion` with mock model and tokenizers
- [ ] Test `is_finished` in isolation

thomasahle:
> The k-v cache will require some careful thought, and will likely have to be customized to the particular class of models for which it is implemented (transformers, llama.cpp, etc.). My initial though is to build a trie that we query each time the model is called; this cache is attached to the model instance.

It might be easier to allow the user to handle the caching. That way the model doesn't have to save all the text it has ever been prompted with. E.g.
```
prompt = "Here's a bunch of examples"
_text, kv_cache = model.generate(prompt, tokens=0, return_kv=True)  # Don't generating any text, just return the kv-cache
output1 = model.generate("", context=kv_cache)  # Use kv-cache from previous generation
output2 = model.generate("", context=kv_cache)  # Use kv-cache from original generation
```

I think this is a bit similar to how Hugging Face handles kv-caches right now.

It would also allow using a kv_cache that has been "learned" by fine-tuning, rather than representing an actual prefix.

At some point we'd like to give the model access to a vector db of "external" key-value pairs. I wonder if you are interested in having such a feature in outlines as well.

rlouf:
> It might be easier to allow the user to handle the caching. That way the model doesn't have to save all the text it has ever been prompted with. E.g.

It will be easier across generation calls, I agree. In which case I'd just return a state that contains more than the completion, or a tuple `(state, extra)` by default. We'd need to do this to return extra info about the generation process, e.g. with SMC sampling. To be refined but something like:

```python
import outlines.text as text

completion, extra = text.completion(model)(prompt)
_ = text.completion(model, kv_cache=extra.kv_cache)(completion)
```

We will still need to manage the k-v cache internally when sampling a single sequence, however.

> I think this is a bit similar to how Hugging Face handles kv-caches right now.

Not sure HF returns the kv cache with the `generate` method, not that it really matters.

> It would also allow using a kv_cache that has been "learned" by fine-tuning, rather than representing an actual prefix.
> At some point we'd like to give the model access to a vector db of "external" key-value pairs. I wonder if you are interested in having such a feature in outlines as well.

All valid use cases which justify allowing external cache management.

I won't implement the k-v cache in this PR, which is getting too big already, but will open an issue instead. I will copy over these comments and add a reference to that [vLLM](https://github.com/vllm-project/vllm) paper as well.

rlouf:
> Left to implement:
> 
> ## Models
> * [ ]  Add cache to the `__call__` method and persist to disk
> * [ ]  Add cache for kv values

This is almost ready for review. I just want to do some extra cleanup, in particular the `__call__` code could be a tad cleaner. I won't implement the other model-related items, and will open issues instead. Same for the re-implementation of `stop_at`,  `Choice`, `Integer`, `Float`, the OpenAI integration and SMC sampling.

outlines/models/transformers.py
```diff
@@ -0,0 +1,101 @@
...
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+            self.pad_token_id = self.eos_token_id
+        else:
+            self.pad_token_id = self.tokenizer.pad_token_id
+            self.pad_token = self.tokenizer.pad_token
+
+    def encode(
+        self, prompt: Union[str, List[str]]
+    ) -> Tuple[NDArray[np.int64], NDArray[np.int64]]:
+        output = self.tokenizer(prompt, padding=True, return_tensors="np")
```
brandonwillard, line 84: Pass-through arguments here, too?

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,238 @@
...
+        num_tokens = token_ids.shape[-1]
+        new_token_ids = np.empty(batch_shape + (num_tokens + 1,), dtype=np.int64)
+
+        token_ids_finished = token_ids[is_finished]
+        batch_shape_finished = token_ids_finished.shape[:-1]
+        token_ids_finished = np.concatenate(
+            [
+                token_ids_finished,
+                self.model.tokenizer.pad_token
+                * np.ones(batch_shape_finished + (1,), dtype=np.int64),
```
brandonwillard, line 137: Can we use the broadcasting functions here instead?  Might remove a temporary array creation if the `ones` are only used for broadcasting.

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,238 @@
...
+):
+    """Vectorized implementation of `np.random.choice`.
+
+    `np.random.choice` does not support arrays of probability. This implements
+    the equivalent of this function where the `p` argument can be a matrix.
+
+    Note
+    ----
+    `searchsorted` might be more efficient here since the number of elements
+    can be quite large.
```
brandonwillard, line 217: That's [what we used in Aesara](https://github.com/aesara-devs/aesara/blob/2368ed3e7f14801785280038094ac99773e876a6/aesara/tensor/random/basic.py#LL1970C26-L1970C26).

brandonwillard (commented):
Just some minor comments for now.  Still looking over it, though.

outlines/text/sequences/sequence.py
rlouf: Made the change. Note that despite the smaller complexity `searchsorted` is slower for small arrays, but here the vocabulary is large and I found appreciable speedups for size ~ 10k+.

outlines/text/sequences/sequence.py
rlouf: Also changed the implementation in `expand_attention_mask`

outlines/models/transformers.py
rlouf: We can use pass-through arguments here too, although it makes no sense to override the values of `padding` and `return_tensor`.

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,243 @@
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+class Sequence:
```
rlouf, line 7: Wondering if `SequenceGenerator` might be a better name. We might want to introduce a `Sequence` state object later on to track quantities such as last output, logprobs and the KV cache.

rlouf: In the same vein, should we rename the `sequences` directory to `generation`?  Calling `text.generation.completion` is more explicit than `text.sequences.completion` imo.

outlines/models/tokenizer.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Tuple, Union
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+class Tokenizer:
```
rlouf, line 7: Should this subclass `Protocol`? I'm not sure what the difference is with this implementation.

outlines/models/transformers.py
```diff
@@ -0,0 +1,104 @@
...
+from outlines.models.tokenizer import Tokenizer
+
+if TYPE_CHECKING:
+    from transformers import PreTrainedModel, PreTrainedTokenizer
+
+
+__all__ = ["transformers"]
+
+
+class Transformers:
```
rlouf, line 16: Should we add a model protocol, or wait until we have several model implementations?

outlines/text/sequences/sequence.py
rlouf: Oh and should this be an abstract base class?

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,243 @@
...
+            ],
+            axis=-1,
+        )
+
+        new_token_ids[~is_finished] = token_ids_unfinished
+        new_token_ids[is_finished] = token_ids_finished
+
+        return new_token_ids
+
+    def __call__(
```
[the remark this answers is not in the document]
brandonwillard: Yes; I think we should almost always allow fine-grained control over the RNG states in use.

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+vsearchsorted = np.vectorize(np.searchsorted, otypes=[int], signature="(n),()->()")
+
+
+def vectorized_random_choice(
+    items: NDArray[np.int64],
+    p: NDArray[np.float64],
+    samples: int = 1,
+):
```
brandonwillard, line 212: This could use an RNG state parameter, as hinted at above.

outlines/text/sequences/sequence.py
brandonwillard: > Wondering if `SequenceGenerator` might be a better name. We might want to introduce a `Sequence` state object later on to track quantities such as last output, logprobs and the KV cache.

Sounds like we might want to make these actual Python generators.

rlouf: The issue I see with making these Python generators, is that there are cases in which we need to post-process the generated string. For instance when you want to stop generation after a given string has been generated, you generally don't want said string to be included (see [here](https://github.com/normal-computing/outlines/blob/7901224b3a663ed9863a89a2d08f0744c6184781/examples/pick_odd_one_out.py#L37) for instance).

brandonwillard: > The issue I see with making these Python generators, is that there are cases in which we need to post-process the generated string. For instance when you want to stop generation after a given string has been generated, you generally don't want said string to be included (see [here](https://github.com/normal-computing/outlines/blob/7901224b3a663ed9863a89a2d08f0744c6184781/examples/pick_odd_one_out.py#L37) for instance).

Could we formulate this in a more functional/"stream" processing way, so that the kind of post-processing you describe is applied to the stream/generator returned to the user?

rlouf: It's worth trying

rlouf:
I added integration tests for `Completion` using a `transformers` model, and allowed fine-grained control over the RNG states in `Sequence`. In my opinion, making `Sequence` a python generator requires more thinking, and can always be done later once this PR is merged.

I am also unsure about the naming, but this can be changed before the next release once we're at feature parity with the current API. Addressing #162 may also impact the naming down the line.

This is ready for another review.

arunpatro:
Hi all, 

I couldn't find documentation for this code. Is there any way to generate the docs? Nevertheless, I played around with this branch. 

```
import outlines.text as text
import outlines.text.sequences as sequences

model_name = "gpt2"
llm = models.transformers(model_name, device="cuda")

outs=sequences.completion(llm, max_tokens=200)("Say something")

print(outs)
```

I am trying to figure out the right way to use the API correctly. This snippet shows the model is indeed loaded in the GPU, but a 0% gpu-util. And its over 3mins but no output. I observe a non-zero gpu-util on MPS, but still takes forever.

rlouf:
> I couldn't find documentation for this code. Is there any way to generate the docs?

It's not documented yet, API might still change.

> I find a large difference in native hf generation vs this api generation. Is it because of the constant device-cpu tensor-NDArray conversions? 

Could be, or I'm doing something wrong with the arrays. Given the size of token id arrays I didn't think it would make a huge difference, but I will need benchmark this.

> Can we also provide a warning message if no. of tokens is not specified, this would not terminate the while True loop, if a stopping_criterion is also not mentioned. 

The default stopping criterion is `EOS`, which I think GPT2 does not have? It may be a good idea to set a default value for `max_tokens` or a warning as you suggested.

rlouf:
I timed Outlines and `transformers` using the following code. I include encoding/decoding for a fair comparison:

```python
def test_time_outlines():
    import time
    from outlines.text.sequences.completion import completion

    now = time.time()
    sequence = completion(model, max_tokens=100)("This is a prompt")
    print(f"Outlines: {time.time()-now:.2f}")

def test_time_hf():
    import time

    now = time.time()
    model = AutoModelForCausalLM.from_pretrained("gpt2")
    tokenizer = AutoTokenizer.from_pretrained("gpt2")
    inputs = tokenizer(["This is a prompt"], return_tensors="pt")
    out = model.generate(**inputs, do_sample=True, max_new_tokens=100)
    sequence = tokenizer.batch_decode(out)
    print(f"HuggingFace: {time.time()-now:.2f}")
```

On CPU it is a wash, and I am yet to try this on GPU. Is that where you saw the difference? Could you post the code you used and some measurements?

outlines/models/transformers.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(
+        self,
+        model: "PreTrainedModel",
+        tokenizer: "PreTrainedTokenizer",
+        device: Optional[str] = None,
+    ):
+        import torch
+
+        if device is None:
+            if torch.cuda.is_available():
```
rlouf, line 28: I don't think we should do this automatically

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+        else:
+            batch_shape = token_ids.shape[:-1]
+            is_finished = np.zeros(batch_shape, dtype=np.bool_)
+
+        while True:
+            num_generated_tokens = token_ids.shape[-1] - num_prompt_tokens
+            if np.all(is_finished) or num_generated_tokens == self.max_tokens:
+                break
+
+            token_ids_unfinished = token_ids[~is_finished]
```
rlouf, line 197: Boolean indexing introduces a lot of copying at each step, wondering how inefficient that might be.

outlines/text/sequences/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+        pad tokens to the finished sequences.
+
+        """
+        batch_shape = token_ids.shape[:-1]
+        num_tokens = token_ids.shape[-1]
+        new_token_ids = np.empty(batch_shape + (num_tokens + 1,), dtype=np.int64)
+
+        token_ids_finished = token_ids[is_finished]
+        batch_shape_finished = token_ids_finished.shape[:-1]
+        token_ids_finished = np.concatenate(
```
rlouf, line 142: cf comment on Boolean indexing below. Padding could be performed once at the end of generation.

rlouf:
> What is the intuition here? How can outlines be faster than native, considering we do a lot of work on top? Are we avoiding extra boilerplate code that HF has inside their `generate` function? 

We do _less_ things than native in `completion`, I don't use the `generate` method models but call their forward method to get the logits and then do the sampling myself. Several benefits:

1. To draw several samples, `transformers` duplicates the input as many times as there are samples and then does the forward pass with this batched input. `outlines` run the forward pass with the single input and then draws several samples using the logits;
2. When doing inference in batch `transformers` keeps feeding finished sequences to the model. Outlines only feeds the unfinished sequences.

arunpatro:
I see, that makes sense. 

> The reason I decouple the model call / sampling this way is to be able to support different model providers and sampling methods (like SMC) in the future.

I like this design decision,  of de-coupling the model and the generation process. 

Question: Why does outlines code not load the model in GPU automatically? According to the code I see, if `device=None`, it should be auto inferred right? Instead of `device=None`, maybe we can keep the default as `device='auto'`, which would imply the meaning correctly. 

I also ran your test cases and I can confirm that outlines is faster on cuda for `gpt2` however I cannot say the same for any other model.

```python
# model_name = "gpt2"
model_name = "togethercomputer/RedPajama-INCITE-Instruct-3B-v1"
prompt = "This is a prompt"

def test_time_outlines():
    import time

    from outlines.text.sequences.completion import completion

    now = time.time()
    model = models.transformers(model_name)
    sequence = completion(model, max_tokens=100)(prompt, samples=10)
    print(f"Outlines: {time.time()-now:.2f}")


def test_time_hf():
    import time

    now = time.time()
    model = AutoModelForCausalLM.from_pretrained(model_name)
    inputs = tokenizer([prompt], return_tensors="pt")
    out = model.generate(**inputs, do_sample=True, max_new_tokens=100, num_return_sequences=10, use_cache=False)
    sequence = tokenizer.batch_decode(out)
    print(f"HuggingFace: {time.time()-now:.2f}")
    
  
def test_time_outlines_cuda():
    import time

    from outlines.text.sequences.completion import completion

    now = time.time()
    model = models.transformers(model_name, 'cuda')
    sequence = completion(model, max_tokens=100)(prompt, samples=10)
    print(f"Outlines: {time.time()-now:.2f}")

 
def test_time_hf_cuda():
    import time

    now = time.time()
    model = AutoModelForCausalLM.from_pretrained(model_name).to('cuda')
    inputs = tokenizer([prompt], return_tensors="pt")['input_ids'].to('cuda')
    out = model.generate(inputs, do_sample=True, max_new_tokens=100, num_return_sequences=10, use_cache=False)
    sequence = tokenizer.batch_decode(out)
    print(f"HuggingFace CUDA: {time.time()-now:.2f}")
```

The times show that HF is faster than Outlines on both cuda and cpu (for `use_cache=False`)

HuggingFace CUDA: 31.83
Outlines CUDA: 47.30
HuggingFace: 88.90
Outlines: 230.86

rlouf:
Thanks! Are those average times over several runs? I'm asking because with any other model than GPT2 the sequence is likely to terminate before `max_tokens` is reached (when `EOS` is found).

A possibility is that with a bigger vocabulary we're paying for the memory transfer, which may be avoided by delegating next-token-sampling to the model class as well. This way we're only transferring a few tokens instead of the full logits. The reason for keeping NumPy is to keep as much generality as we can if we want to include llamacpp, or even JAX or TF `transformers` models.

arunpatro:
> Thanks! Are those average times over several runs? I'm asking because with any other model than GPT2 the sequence is likely to terminate before `max_tokens` is reached (when `EOS` is found).

No, these are not average numbers, but I ran this multiple times, and its similar. 

> PS: do you need an A100 for a 3B model?

Not really, but I am experimenting with upto 13B models, and sometimes I need a lot of GPU-RAM cuz of batchsizes (vLLM requirements, etc)

rlouf:
> Question 2: Why are we moving back and forth torch and numpy arrays? I modified this branch to only use torch.tensors and it can be drop in replacement (except MPS devices because `torch.searchsorted` is not implemented for MPS)

I have thought more about this, and there's probably no good reason to not be using PyTorch as the default. This would allow to keep a strict separation between model calls and logit manipulation without moving memory around. Arrays output by llamacpp can be converted into `torch.Tensor` at virtually no cost, and apparently so can JAX and TF arrays (although that's slightly more complicated).

arunpatro:
> Because HF models are loaded on cpu by default, and you need to explicitly move them to another device. I'm following their conventions as much as I can. Do you think we should do differently?

No, this is good. We should stick to HF wherever we can for defaults. The correct and fastest way to load model to device, is to use `config.device_map = 'auto'` which HF also expects as `**kwargs`.

outlines/text/generate/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+        probabilities and then sample one or several new tokens from a
+        categorical distribution parametrized by these probabilities.
+
+        Parameters
+        ----------
+        rng
+            NumPy random number Generator instance
+        token_ids
+            The token ids passed as an input to the model, of shape `batch_shape
+            + (num_tokens,)`.  The generated token is a random variable
```
brandonwillard, line 54: Is `num_tokens` supposed to be `samples`?

outlines/text/generate/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+        contains the completion (input token ids and generated token ids) and an
+        array of shape `new_batch_shape + (vocab_size,)` that contains the next
+        token probabilities. `new_batch_shape` is computed by removing dimensions
+        of size one in `(samples,) + batch_shape`.
+
+        """
+        num_input_dims = token_ids.ndim
+        probs = self.model(token_ids, attention_mask)
+
+        # Sample `samples` new tokens
```
brandonwillard, line 71:
```suggestion
        # Sample `samples`-many new tokens
```

outlines/text/generate/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+        vocab_size = probs.shape[-1]
+        next_token_ids = vectorized_random_choice(
+            rng, np.arange(vocab_size), probs, samples
```
brandonwillard, line 74: We might want to consider how we can avoid creating an array like `np.arange(vocab_size)` if possible.  If `vectorized_random_choice` is only indexing that array, then we should be able to in at least this case, no?

brandonwillard (approved):
Minor questions and comments; otherwise, I'm good with moving forward on this.

outlines/text/generate/sequence.py
```diff
@@ -0,0 +1,259 @@
...
+
+            num_batch_dims = token_ids.ndim - 1
+            repetitions = (samples,) + (1,) * num_batch_dims
+            attention_mask = np.tile(attention_mask, repetitions)
+            attention_mask = self.expand_attention_mask(attention_mask)
+        else:
+            batch_shape = token_ids.shape[:-1]
+            is_finished = np.zeros(batch_shape, dtype=np.bool_)
+
+        while True:
```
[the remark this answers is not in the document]
rlouf: Yeah afaik everything is covered in that function.

outlines/text/generate/sequence.py
rlouf: The naming is unclear. `samples` = `num_sampled_sequences` and `num_tokens` = number of tokens in the sequences. I'll try to make this clearer.

outlines/text/generate/sequence.py
rlouf: Good catch!

PR #140, opened 2023-06-09
Add `JSON` generation method

rlouf:
In this PR we add a new generation method that can generate valid JSON following a [Pydantic mode](https://docs.pydantic.dev/latest/#pydantic-examples) or [JSON Schema](https://json-schema.org/) specification.

We first parse the (possibly nested) JSON schema into a generation schedule, which is a list of strings that represent the JSON structure, and annotations that correspond to the field specifications. The following annotations are currently supported, with a minimal set of functionalities:

- `string`
- `int`
- `float`
- `bool`
- `Union[type1, type2]`
- `enum`
- `List[type]`
- `List[Union[type1, type2]]`
- `Optional[type]`
- Nested schema
- `maxLength` for strings
- `minLength` for strings 

The tests demonstrate how to build a schedule from Pydantic models, using the models' `schema_json` method. 

We also add an `outlines.text.generate.json` function that parses a JSON schema, turns it into a regex and instantiates the `Regex` class.

### DIFF
--- a/.github/workflows/build_documentation.yml
+++ b/.github/workflows/build_documentation.yml
@@ -14,10 +14,10 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Set up Python 3.9
+      - name: Set up Python 3.10
         uses: actions/setup-python@v1
         with:
-          python-version: 3.9
+          python-version: "3.10"
 
       - name: Build the documentation with Sphinx
         run: |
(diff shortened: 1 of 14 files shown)


havietisov:
off topic : would be interesting to see constrained generation of HTML

rlouf:
> off topic : would be interesting to see constrained generation of HTML

Coming with #178

brandonwillard (approved):

PR #166, opened 2023-06-29
Add `Integer` sequence generator

### LINKED ISSUE #153: Add an `Integer` generation method
We can steer the generation process so we only output integers. I propose to implement an `Integer` subclass of `Sequence` that uses [masking](https://github.com/normal-computing/outlines/blob/a4662218232df8264d1b9eb93867362ed8da2e54/outlines/text/masks.py#L36) to restrict the generated tokens to integers. We will need to add a `create_proposal` method to `Sequence` that applies the mask to the logits generated by the model.

We can also add constraints on the generated integers:

* Add an upper or lower bound on the number of digits generated
* Add an upper or lower bound on the value of the integer

We will probably need [SMC sampling](https://arxiv.org/abs/2306.03081v1) to implement the constraints.

rlouf:
In this PR I introduce a sequence generator that only outputs integers. Closes #153.

The generator can currently generate sequences that have leading zeros, which are then trimmed during post-processing. I am not sure what we should do about it. On one hand we can ignore it (maybe the LLM has seen this during training), on the other hand we could use the pre-processing approach described in #131.

### DIFF
--- a/outlines/text/generate/integer.py
+++ b/outlines/text/generate/integer.py
@@ -1,0 +1,96 @@
+import math
+from typing import List, Optional, Tuple
+
+import interegular
+import torch
+
+from outlines.text.generate.continuation import Continuation
+from outlines.text.parsing import find_partial_matches, map_partial_states_to_vocab
+
+
+class Integer(Continuation):
+    """Represents a integer generation model.
+
+    `Integer` instances are constrained generation models that only
+    generate integer values. Leading zeros are fobidden. EOS tokens
+    are only allowed after at least one digit has been generated.
+
+    >>> import outlines.text as text
+    >>> sequence = text.generate.integer(model)("Return an integer between 0 and 10")
+
+    """
+
+    def __init__(self, model, max_tokens: Optional[int]):
+        super().__init__(model, max_tokens)
+
+        vocabulary = model.tokenizer.vocabulary
+        sorted_vocabulary = [
+            k for k, v in sorted(vocabulary.items(), key=lambda kv: kv[1])
+        ]
+
+        int_regex_string = r"(0|[1-9][0-9]+)"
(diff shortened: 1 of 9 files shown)


rlouf:
The code now makes use of the tools introduced in #131 and the follow-up #170 to pre-compute the masks and advance the parsing state.

outlines/text/generate/integer.py
```diff
@@ -0,0 +1,72 @@
...
+            partial_matches = find_partial_matches(self.int_regex_fsm, sampled_sequence)
+            pmatch = max(
+                partial_matches, key=lambda x: x[0] if x[0] is not None else -1
+            )
+            self.pstate: Tuple[str, int] = (self.pstate[0], pmatch[1][-1])
+        else:
+            self.pstate = ("INT", self.int_regex_fsm.initial)
+
+        next_support = self.pstate_to_vocab[self.pstate]
+        mask = -float("Inf") * np.ones(len(self.vocabulary), dtype=np.int64)
```
brandonwillard, line 65: What about `[torch|np].full_like(..., -math.inf)`?  (N.B. I'm not sure about the performance of either&mdash;`full_like` just comes to mind.)

brandonwillard (approved):
Minor comments/questions; looks good otherwise.

rlouf:
The integer generation is working, although we currently cannot handle the end-of-sequence (EOS) token properly. To generate integers the regex we'd like the generated text to match `(0|[1-9][0-9]+)(<EOS>)` where `<EOS>` is the EOS token. The corresponding finite-state machine contains the transitions `"<" -> "E" -> "O" -> "S" -> ">"`, but we can be in a situation where for instance `<EOS` is in the vocabulary but `>` is not, stalling the generation.

Since EOS is always a final state we shouldn't have too much trouble modifying the alphabet/map of the FSM to get the desired behavior.

PR #172, opened 2023-07-06
Generalize parse state cloning

brandonwillard:
This PR makes some simple generalizations to our custom `lark` parse state cloning.  It also adds SQL-guided generation as an extension to the existing LLM example.

N.B. Don't expect the example model (i.e. [`codegen-350M-mono`](https://huggingface.co/Salesforce/codegen-350M-mono)) to actually produce meaningful/interesting SQL; it was trained primarily/exclusively on Python code.  Also, the example SQL grammar could have issues of its own.

### DIFF
--- a/examples/parsing.py
+++ b/examples/parsing.py
@@ -1,6 +1,7 @@
 """An example illustrating parser-based masking."""
 import math
 import time
+import urllib.request
 
 import torch
 from lark import Lark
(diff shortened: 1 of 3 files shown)


examples/parsing.py
```diff
@@ -26,23 +27,33 @@
     checkpoint, trust_remote_code=True, revision=revision
 ).to(device)
 
-input_text = "def "
-inputs = tokenizer.encode(input_text, return_tensors="pt").to(device)
+sql_grammar_url = "https://github.com/zbrookle/sql_to_ibis/raw/0e9226da42065940ce21439d490f9fcacadc7f92/sql_to_ibis/grammar/sql.lark"
```
rlouf, line 30: What do you think about maintaining some of the grammars we need?

brandonwillard: Yeah, we can add them with the appropriate licenses, of course.

PR #175, opened 2023-07-07
Add `Regex` generation method

rlouf:
This trivially generalizes #166 by defining a `Regex` class that can be initialized with any regex string. The `integer` function now instantiates `Regex` with the appropriate regex string.

- [x] General `Regex` class
- [x] Function `integer` to initialize `Regex` with a regex that only matches integers
- [x] Function `float` to initialize `Regex` with a regex that only matches integers
- [x] Handle `EOS` tokens for open-ended sequences


tests/text/generate/test_integer.py
```diff
@@ -68,3 +69,15 @@ def test_integer_proposal():
...
+    result = generator.create_proposal(torch.tensor([[4, 0], [2, 2]]), logits)
+    assert torch.equal(
+        result,
+        torch.tensor(
+            [
+                [1.0, -math.inf, -math.inf, -math.inf, -math.inf, -math.inf],
+                [1.0, 1.0, 1.0, -math.inf, 1.0, -math.inf],
+            ]
+        ),
+    )
```
brandonwillard, line 83: This test demonstrates the EOS and padding functionality by only allowing EOS to be sampled for the sequence that has already terminated with EOS (i.e. the first sequence).  It's set to use EOS whenever `model.tokenizer.pad_token` isn't available.  I'm not sure if that's the best approach, but it should be easy to change regardless.

brandonwillard (commented):
@rlouf, I added the EOS handling and padding.

brandonwillard (commented):
FYI: The current error in CI looks like a "determinism" problem with the numbering/order of the FSM states.

tests/text/generate/test_integer.py
rlouf: We need to generate `eos_token_id` in all cases here. When found the `is_finished` method will mark the current sequence as finished, and the `__call__` method will then stop sampling new tokens for this sequence and insert the `pad_token_id`s.

rlouf:
I would like to add a test to define the behavior when there is no possible match in the vocabulary. Then it should be ready to merge.

rlouf:
Now the `Regex` class raises an exception when the vocabulary does not allow to build sequences that match the input regex. Ready for review.

brandonwillard (approved):
@rlouf, I just pushed a change that makes the FSM start from the previous state it was in: i.e. it avoids decoding and rerunning the FSM for the entire sequence on each iteration.  Tell me if that looks fine; otherwise, everything else looks good to me.

brandonwillard:
Agh, looks like the generated token sequences change shape when using `transformers`!

outlines/text/generate/regex.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # TODO: This check might be a little too strict, because I think that
+        # while some states are made unreachable by a vocabulary (and will not
+        # be present in the following set difference), there could still be
+        # paths to terminal states emanating from the states that are reachable.
+        states_with_transition = {x[1] for x in pstate_to_vocab.keys()}
+        if len(self.regex_fsm.states.difference(states_with_transition)) > 0:
+            raise ValueError(
+                "The vocabulary does not allow us to build a sequence that matches the input regex"
+            )
```
brandonwillard, line 55: We should look into this.  Perhaps as a follow-up issue.

outlines/text/masks.py
```diff
@@ -35,14 +35,16 @@ def create_mask_from_regex(vocabulary: Dict[str, int], regex: str) -> torch.Bool
...
 
     return mask
 
 
 def create_float_mask(vocabulary: Dict[str, int]) -> torch.BoolTensor:
     """Create a mask to generate floating point numbers."""
-    mask = create_mask_from_regex(vocabulary, r"^(([0-9]+)?([.]([0-9]*)?)?|[.][0-9]+)$")
+    mask = create_mask_from_regex(
+        vocabulary, r"([+-]?((0|[1-9]+)([.][0-9]*)?)|([.][0-9]+))"
+    )
```
brandonwillard, line 47: The tests were getting a little flaky in CI, so I had to add these updates.

brandonwillard (approved):
OK, I think it's good to go now.

outlines/text/generate/regex.py
rlouf: Good point. I opened #184 to track this.

outlines/text/generate/regex.py
```diff
@@ -0,0 +1,160 @@
...
+        if generated_token_ids.shape[-1] > 0:
+            new_pstates = []
+            for token_seq, (_, last_fsm_state, last_token_idx) in zip(
+                generated_token_ids,
+                self.pstates,
+            ):
+                # Get the tokens we haven't already processed
+                readable_tokens = token_seq[last_token_idx:]
+                # excluding any EOS tokens
+                not_eos_mask = [
```
rlouf, line 92: You should never get a sequence with an `EOS` token here. Those are filtered out in `Sequence.__call__`.

rlouf:
I only have one minor comment regarding `EOS` tokens. `Sequence.__call__` filters finished sequences, and `Regex` inherits from `Continuation` which marks a sequence as finished when an `EOS` token is found. We can open a follow-up issue for this.

outlines/text/generate/regex.py
brandonwillard: If you start it out with a sequence like `[[10, 2, 0, 0]]`, you would only want to process the first two.  That's what it should be doing.

rlouf: When would that happen if you cannot get sequences with `0` by design?

brandonwillard: I believe I had to add it for the tests at one point, but that might not longer be true.

PR #178, opened 2023-07-10
Use FSMs for scanning during grammar-guided generation

### LINKED ISSUE #170: Pre-parse vocabulary strings in order to more efficiently determine the next non-zero probability tokens
Look into parsing a vocabulary "off-line" in order to construct maps from partial parse states to subsets of the vocabulary that can continue from those parse states (e.g. subsets of the vocabulary that are valid [partial] `NAME`/`DEF`/etc. token matches).  With these maps all the vocabulary tokens/strings could possibly be grouped and classified in such a way that either no "run-time" parsing is necessary, run-time results can be cached/memoized based on the parse state stack, and/or the run-time parsing will only need to be performed on much smaller sets.

For example, a vocabulary token like `"foo("` could be parsed offline to determine that the `"foo"` part is/could be a [partial] `NAME` token, and the `"("` a subsequent `LPAR` token, of course.  With that information we could create a subset of the vocabulary that is `NAME`-prefixed, then, when the current parse state only permits a transition to a `NAME` token, we can reduce the filtered/checked vocabulary to only that subset (or simply reduce/refine the support in rejection sampling scenarios).

[Here's a Gist outlining the idea](https://gist.github.com/brandonwillard/4567b124596835c0dec31df7cd094da8).  A `lark` grammar that relies less on non-standard/regular regex features, like [this SQL one](https://github.com/zbrookle/sql_to_ibis/blob/0e9226da42065940ce21439d490f9fcacadc7f92/sql_to_ibis/grammar/sql.lark), could actually work with the very naive approach implemented in that Gist.  Again, we want as much coverage as we can get out of the box, but, if people are willing to tweak a grammar, we could pre-process vocabularies right now.

N.B. We are already able to do all this for a single regex/finite state machine (FSM).  This issue is specific to the more general context-free grammar (CFG) case.

brandonwillard:
This PR is the first step to closing #170 using the approach described in our paper ["Efficient Guided Generation for Large Language Models"](https://arxiv.org/abs/2307.09702).

In other words, the changes in this PR allow us to create "index" `dict`s for full grammars.  By "index" we mean a `dict` that maps partial parse states to subsets of a vocabulary that are valid continuations of the parse state.

NOTE: The indexing work is being split off into its own PR.

TODO:
- [ ] Use a more efficient DFA implementation. 
- [ ] Use DFA minimization on the parse state FSMs.
    This will remove a large number of unnecessary states and decrease the overall cost of constructing indices.
- [x] We need to do something about ignored tokens in `terminal_symbol_scan`.
    They are currently being _completely_ ignored, which incorrectly excludes vocabulary strings consisting of only ignored tokens (e.g. `" "` in a Python grammar).
- [x] Convert the terminal symbol FSM states used to compute the indices to their corresponding parser state (aggregate/unioned) FSM states.

### DIFF
--- a/outlines/text/parsing.py
+++ b/outlines/text/parsing.py
@@ -4,7 +4,7 @@ from typing import Any, Callable, DefaultDict, Dict, Iterable, Optional, Set, Tu
 
 import interegular
 import regex
-from interegular.fsm import FSM, anything_else
+from interegular.fsm import FSM, Alphabet, anything_else
 from interegular.patterns import Unsupported
 from lark import Lark, Token
 from lark.common import LexerConf, ParserConf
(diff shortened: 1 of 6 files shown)


outlines/text/parsing.py
```diff
@@ -332,17 +368,17 @@ def _partial_match(
         if not terminated and state == fsm.initial:
             return None, None
 
-        return None if not terminated else i, accepted_states
```
rlouf, line 335: I'm surprised I did not see any bug related to this

dpsimpson:
It is I, Daniel, smasher of for loops, obsfucator of code here to tell you that you can replace the for loop [here](https://github.com/brandonwillard/outlines/blob/1b804fe16864a1e23906457de4093308d024a597/examples/parsing.py#L108) (aka the one starting at line 108 of /examples/parsing.py with the probably more efficient, if somewhat less clear
(assuming that `next_vocab` is a list of integers, otherwise add some code to make it so)
```
ids = torch.tensor(next_vocab, dtype = torch.uint8).unsqueeze(0) # same number of dims as mask
mask = mask.scatter_(1, ids, torch.zeros_like(ids) # trailing _ is the in-place version
```

Is this a sensible change for an example? No. Will it be faster when `next_vocab` is big and you're on a cuda device? It should be.

samuela:
OOC how does this work and the corresponding paper relate to the "[Parsing with Derivatives](https://matt.might.net/papers/might2011derivatives.pdf)" from @mattmight et al?

brandonwillard:
> OOC how does this work and the corresponding paper relate to the "[Parsing with Derivatives](https://matt.might.net/papers/might2011derivatives.pdf)" from @mattmight et al?

We're aware of that work, but this PR does not use anything from it directly.  The work being done here is almost exclusively concerned with the adaptation of existing LALR(1) parsers and the use of the "indices" described in our technical paper.

brandonwillard:
I'm going to merge this after the tests pass and create follow-up issues for the remaining work mentioned in the description.  Doing so will fix the flaky FSM tests.

Wehzie:
I'm very interested in this PR. Will there be documentation detailing how a context free grammar can be specified for guided generation?

brandonwillard:
> I'm very interested in this PR. Will there be documentation detailing how a context free grammar can be specified for guided generation?

Yes, and hopefully very soon!  We have a development roadmap for integrating this into our `Sequence` framework and user interface and providing a performant sampling-based approach (to be followed by indexing and more).  The next steps on that path are taking place in #272 right now.

rlouf:
Working on it!

PR #183, opened 2023-07-12
Fix overview document and a type hint

mlelarge:
Very nice work! Thanks!
Just started playing with it and got an error with the Hello World example in the documentation. This small modification fixed the issue for me.

Another remark: experimenting in a Jupyter notebook does not work  because of a pb with asynchio, see https://pypi.org/project/nest-asyncio/
The solution proposed there worked for me:
```
import nest_asyncio
nest_asyncio.apply()
```
It would probably be nice to refer to it in the documentation.

I hope this helps.
Best,
Marc

### DIFF
--- a/docs/source/overview.rst
+++ b/docs/source/overview.rst
@@ -18,7 +18,7 @@ Here is a simple Outlines program that highlights some of its key features:
 
    hello_world = where_from("Hello world")
    foobar = where_from("Foo Bar")
-   answer = complete([hello_world, foobar], num_samples=3, stop_at=["."])
+   answer = complete([hello_world, foobar], samples=3, stop_at=["."])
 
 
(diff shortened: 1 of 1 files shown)


outlines/text/prompts.py
```diff
@@ -284,7 +284,7 @@ def get_schema_dict(model: Dict):
 
 
 @get_schema.register(type(BaseModel))
-def get_schema_pydantic(model: type[BaseModel]):
+def get_schema_pydantic(model: type(BaseModel)):
```
brandonwillard, line 287: @mlelarge, mind rebasing and removing this one?  If not, no worries; we can do it from here.

brandonwillard (changes requested):

outlines/text/prompts.py
mlelarge: Not sure to understand what you want me to do. Probably better if you take it over from here. Sorry...

rlouf:
Thank you for contributing @mlelarge!

PR #187, opened 2023-07-13
Stop generation with `Continuation` when a specific string was generated

### LINKED ISSUE #151: Stop generation with `Continuation` when a specific string has been generated
The current user interface allows user to pass a `stop_at` keyword argument that contains a list of string, so that the generation when either of the strings has been generated. I suggest to implement this by adding a `stop_at` keyword argument to `Completion` init method.

This will require to:

1. Modify the `is_finished` method which will return `True` when an EOS token or either of the strings has been generated.
2. Add a `postprocess` method to `Sequence` that takes the output string and removes the string responsible for stopping the generation.

rlouf:
Closes #151

### DIFF
--- a/tests/text/generate/test_integration_transfomers.py
+++ b/tests/text/generate/test_integration_transfomers.py
@@ -13,20 +13,24 @@ def test_transformers_integration_continuation():
 
     model_name = "hf-internal-testing/tiny-random-GPTJForCausalLM"
     model = models.transformers(model_name, device="cpu")
-    sequence = generate.continuation(model)("Write a short sentence", rng=rng)
+    sequence = generate.continuation(model)("Write a short sentence ", rng=rng)
     assert isinstance(sequence, str)
     assert model.tokenizer.eos_token not in sequence
 
     sequence = generate.continuation(model, max_tokens=10)(
-        "Write a short sentence", rng=rng
+        "Write a short sentence ", rng=rng
     )
     assert isinstance(sequence, str)
 
-    prompts = ["Write a short sentence", "And another one"]
+    prompts = ["Write a short sentence ", "And another one "]
     sequence = generate.continuation(model, max_tokens=10)(prompts, rng=rng)
     assert isinstance(sequence, list)
     assert len(sequence) == 2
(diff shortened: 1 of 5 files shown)


tests/text/generate/test_integration_transfomers.py
```diff
@@ -13,21 +13,25 @@ def test_transformers_integration_continuation():
...
     assert model.tokenizer.eos_token not in sequence
 
     sequence = generate.continuation(model, max_tokens=10)(
-        "Write a short sentence", rng=rng
+        "Write a short sentence ", rng=rng
     )
     assert isinstance(sequence, str)
 
-    prompts = ["Write a short sentence", "And another one"]
+    prompts = ["Write a short sentence ", "And another one "]
```
arunpatro, line 25: I am just curious why you added whitespace padding? According to `guidance`, its preferable to terminate prompts without any new space or line, because frequent tokens already come with a space before the word.

arunpatro:
Looks good to me.

arunpatro (approved):

tests/text/generate/test_integration_transfomers.py
rlouf: Good question, that came intuitively with GPT2 and numbers. I'll look at the vocabulary directly to see if that's actually the right thing to do.

brandonwillard (approved):

tests/text/generate/test_integration_transfomers.py
rlouf: I came back to this and looked at the vocabulary of the GPT2 tokenizer. It is true that most of the tokens begin with a space.

Your point highlights something that we need to be very careful about, and which might be incorrectly implemented in outlines.

For instance, the regex `[a-z]{3}` will allow "art" to be generated, but not " art". This could make it impossible to generate what would otherwise be the most probable completion. I need to dig more into this. I opened a discussion to keep track of my thinking on this.

arunpatro: You are right, in that token healing should be able to correct all these nuances.

PR #188, opened 2023-07-13
Generate a choice between different strings

### LINKED ISSUE #152: Add a `Choice` generation method
The current user interface of outlines provides a `is_in` keyword argument which constrains the generation to choose between different options. I propose to add a new `Choice` subclass to `Sequence` that implements this functionality.

I can think of two ways to implement this:

1. Create a token mask so only valid prefixes can be generated and update the mask as new tokens are generated until only one option remains (current method).
2. Call the model to compute the logprob corresponding to each choice, normalize them and sample the choice from this distribution.

We need a good argument to choose between these two implementations.

### LINKED ISSUE #184: Relax check for whether we can build a sequence that matches a regex
We can check whether a vocabulary allows to build a sequence that matches a regex when initializing `Regex`. The check is currently implemented as:

```python
states_with_transition = {x[1] for x in pstate_to_vocab.keys()}
if len(self.regex_fsm.states.difference(states_with_transition)) > 0:
    raise ValueError(
        "The vocabulary does not allow us to build a sequence that matches the input regex"
    )
```

However this check might be too strict. While some states might be unreachable with a dictionary there could still be a path to terminal states. We should  probably raise a warning when this happens, not an exception.

rlouf:
Closes #152. I use the `Regex` generation method using a `(choice_1|choice_2|...|choice_n)` regex. This method is equivalent to rejection sampling: draw a token and discard it if it is not on a path that generates either of the choices.

An alternative implementation would be to tokenize the choices, do a forward pass through the model with each and sample from the resulting categorical distribution. However, some strings can be built from different combinations of tokens and we would only be considering one for each stopping sequence.

Unsurprisingly, the check to verify that a sequence can be generated with the vocabulary added in #175 fails. We fix this by checking whether we can build terminal paths with the vocabulary. Closes #184.

### DIFF
--- a/outlines/text/parsing.py
+++ b/outlines/text/parsing.py
@@ -371,7 +371,9 @@ def map_partial_states_to_vocab(
         [str, Optional[int], Tuple[int, ...]], bool
     ] = lambda *args: True,
     final_state_string: Optional[str] = None,
-) -> DefaultDict[PartialParseState, Set[int]]:
+) -> Tuple[
+    DefaultDict[PartialParseState, Set[int]], Dict[str, DefaultDict[int, Set[int]]]
+]:
     """Construct a map from partial parse states to subsets of `vocabulary`.
 
     The subsets of `vocabulary` consist of elements that are accepted by--or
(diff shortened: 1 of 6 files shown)


outlines/text/parsing.py
```diff
@@ -383,24 +385,30 @@ def map_partial_states_to_vocab(
...
+    possible_paths = {}
     for symbol_name, fsm in terminals_to_fsms_map.items():
+        terminal_possible_paths = defaultdict(set)
         for i, vocab_string in enumerate(vocabulary):
             if vocab_string == final_state_string:
                 final_state_string_idx = i
 
             for end_idx, state_seq in find_partial_matches(fsm, vocab_string):
                 if partial_match_filter(vocab_string, end_idx, state_seq):
+                    terminal_possible_paths[state_seq[0]].add(state_seq[-1])
```
brandonwillard, line 400: Whether or not the last state is a final/accept state of the FSM will depend on `partial_match_filter`, no?  We might need extra handling for this, or we could track these things in that filter function, after adding the symbol name as an argument, of course.

tests/text/generate/test_regex.py
```diff
@@ -63,10 +63,10 @@ def test_regex_proposal(regex_string, valid_first_token, proposal):
     assert result.squeeze()[0] == 1
 
 
-def test_regex_no_valid_transition():
-    model = Model()
-    with pytest.raises(ValueError, match="The vocabulary does not allow"):
-        generate.regex(model, "aw")
+# def test_regex_no_valid_transition():
```
rlouf, line 66: I need to uncomment this

outlines/text/parsing.py
rlouf: The check would fail if the combination of the vocabulary and `partial_match_filter` didn't allow to reach a terminal state, no? Isn't that the desired behaviour?

PR #201, opened 2023-07-19
Align prompt with tokens

### LINKED ISSUE #161: Implement token healing
[Guidance](https://github.com/microsoft/guidance) implements a method called token healing, which consists in correcting for the shortcoming of modern encodings like BPE. See [this notebook](https://github.com/microsoft/guidance/blob/main/notebooks/token_healing.ipynb) for a thorough explanation of why this is necessary.

This consists in backtracking one or several tokens and start generation by imposing that we reproduce the text that corresponds to the removed tokens. This can be integrated in the `__call__` method of the `Sequence` class.

rlouf:
Closes #161


rlouf:
This PR is taking the wrong approach. We should instead use Regex-guided generation by taking the text of the boundary tokens as the beginning of the regex.

PR #204, opened 2023-07-26
Refactor OpenAI Models for parity with Transformers Continuation API

arunpatro:
I have started the refactor work based on our call. However I am kinda confused as how to proceed.

### DIFF
--- a/outlines/text/generate/continuation.py
+++ b/outlines/text/generate/continuation.py
@@ -1,7 +1,10 @@
+import functools
 from typing import List, Optional, Union
 
 import torch
 
+from outlines.models.openai import OpenAI
+from outlines.models.transformers import Transformers
 from outlines.text.generate.sequence import Sequence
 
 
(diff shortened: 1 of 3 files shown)


rlouf:
I updated the PR. I don't think it will work as is, but it should give you an idea of what the design should look like.

arunpatro:
This code works as intended. But pytest fails for `max_tokens` and `stop` becaues of the new api for continuation. However this code works:

```python
model = models.openai("gpt-4", temperature=0.7)
text.generate.continuation(model, max_tokens=100, stop=['tiger'])("This is a story about a tiger")
```

outlines/text/generate/continuation.py
```diff
@@ -88,10 +91,16 @@ def postprocess_completions(self, completions: List[str]) -> List[str]:
         return completions_without_stop
 
 
-def continuation(
+@functools.singledispatch
+def continuation(model, max_tokens):
```
rlouf, line 95: I believe `pytest` fails because this function needs to have the same signature as the other ones.

arunpatro: I don't think so. Looks like the `Model()` class defined in the test doesn't match the types we have created in this dispatch. So it throws a `NotImplementedError``

PR #207, opened 2023-08-01
add ctranslate2 models

arunpatro:
Inspired by Hamel's [benchmarks](https://hamel.dev/notes/llm/03_inference.html) on fast transfomer inference, I propose adding the [`ctranslate2`](https://github.com/OpenNMT/CTranslate2) library. 

This almost works, but fails for multiple samples. The way to sample multiple sequences natively in ctranslate2 is to use
```python
generator = ctranslate2.Generator("redpj", device="cuda")
ans = generator.generate_batch([tokens], beam_size=4, num_hypothesis=4)
```

First compile the HF model:
```
ct2-transformers-converter --model togethercomputer/RedPajama-INCITE-Instruct-3B-v1 --output_dir redpj --quantization bfloat16 
```
This code works: 
```python
model = models.ctranslate2("redpj", "togethercomputer/RedPajama-INCITE-Instruct-3B-v1", "cuda")

prompt = "Hey What's up? In the"
output = text.generate.continuation(model, max_tokens=250)(prompt)
print(output)
```

But this fails:
```python
output = text.generate.continuation(model, max_tokens=250)(prompt, samples=5)
```

### DIFF
--- a/outlines/models/ctranslate2.py
+++ b/outlines/models/ctranslate2.py
@@ -14,7 +14,7 @@ __all__ = ["ctranslate2"]
 
 
 class CTranslate2_Model:
-    """Represents a `transformers` model."""
+    """Represents a `ctranslate2` model."""
 
     def __init__(
         self,
@@ -64,7 +64,7 @@ class TransformersTokenizer(Tokenizer):
         kwargs["padding"] = True
         kwargs["return_tensors"] = "pt"
         output = self.tokenizer(prompt, **kwargs)
-        
+
         return output["input_ids"], output["attention_mask"]
 
     def decode(self, token_ids: torch.LongTensor) -> List[str]:
@@ -76,9 +76,11 @@ class TransformersTokenizer(Tokenizer):
         return string
 
 
-def ctranslate2(ctr2_model: str, tokenizer_name: str, device: Optional[str] = None, **model_kwargs):
+def ctranslate2(
+    ctr2_model: str, tokenizer_name: str, device: Optional[str] = None, **model_kwargs
+):
     import ctranslate2
-    
+
     model = ctranslate2.Generator(ctr2_model, device=device)
     tokenizer = TransformersTokenizer(tokenizer_name, **model_kwargs)
 
--- a/outlines/models/ctranslate2.py
+++ b/outlines/models/ctranslate2.py
@@ -2,8 +2,9 @@ import math
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 import torch
+import numpy as np
 
-from outlines.models.tokenizer import Tokenizer
+from outlines.models.transformers import TransformersTokenizer
 
 if TYPE_CHECKING:
     from transformers import PreTrainedTokenizer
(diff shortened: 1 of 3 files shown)


outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,87 @@
...
+        # `forward_batch` method of `Generator` accepts `tokens` in a list of list of str
+        tokens = [self.tokenizer.tokenizer.convert_ids_to_tokens(iids) for iids in input_ids]
+        logits = self.model.forward_batch(tokens, return_log_probs=True)
+        logits = torch.as_tensor(logits)
+        next_token_logits = logits[:, -1, :]
+
+        return next_token_logits
+
+
+class TransformersTokenizer(Tokenizer):
```
rlouf, line 41: We could import this from `outlines.models.transformers` instead of redefining it?

outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,87 @@
...
+        self.device = device if device is not None else "cpu"
+        self.model = model
+        self.tokenizer = tokenizer
+
+    def __call__(
+        self, input_ids: torch.LongTensor, attention_mask: torch.LongTensor
+    ) -> torch.FloatTensor:
+        # `forward_batch` method of `Generator` accepts `tokens` in a list of list of str
+        tokens = [self.tokenizer.tokenizer.convert_ids_to_tokens(iids) for iids in input_ids]
+        logits = self.model.forward_batch(tokens, return_log_probs=True)
```
rlouf, line 34: The slowness probably comes from the fact that we're not passing the KV cache from one step to the other. Is there any way we can get it as a returned value?

rlouf:
This is really cool, thank you for opening a PR. The slowness must come from the fact that we don't use the KV cache (to be fixed soon). Is there any way we can have the library return it at some point?

outlines/models/ctranslate2.py
arunpatro: Unable to find support for that. It would be good if they provided a HF Auto Class. See: https://github.com/OpenNMT/CTranslate2/issues/1391

arunpatro:
Ctranslate2 supports kv-caching in `model.generate` but doesn't look straightforward via the `model.forward`.

arunpatro:
We can perhaps close this PR, or draft and park it.

rlouf:
If this works but the only issue is the KV cache we can still merge it once the style issues have been fixed.

arunpatro:
Unable to understand how to fix the `pre-commit` error on github actions. On my local `pre-commit` passes completely. 
GH Actions error:
`outlines/models/ctranslate2.py:9: error: Cannot find implementation or library stub for module named "ctranslate2"  [import]`

arunpatro:
I also think that we should support `MLC` as more priority than `ctranslate2`, maybe because of more gh-stars or in general community support of MLC. See: `https://github.com/mlc-ai/mlc-llm`

outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,55 @@
...
+        logits = torch.as_tensor(logits)
+        next_token_logits = logits[:, -1, :]
+
+        return next_token_logits
+
+
+def ctranslate2(
+    ctr2_model: str, tokenizer_name: str, device: Optional[str] = None, **model_kwargs
+):
+    import ctranslate2
```
rlouf, line 50: Let's handle the `ImportError` exception here like I did for transformers in #240 and add a mention in the README.

outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,55 @@
...
+
+if TYPE_CHECKING:
+    from ctranslate2 import Generator  # type: ignore
+    from transformers import PreTrainedTokenizer
+
+
+__all__ = ["ctranslate2"]
+
+
+class CTranslate2_CausalLM:
```
rlouf, line 16:
```suggestion
class CTranslate2:
```

outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,55 @@
...
+
+def ctranslate2(
+    ctr2_model: str, tokenizer_name: str, device: Optional[str] = None, **model_kwargs
+):
+    import ctranslate2
+
+    model = ctranslate2.Generator(ctr2_model, device=device)
+    tokenizer = TransformersTokenizer(tokenizer_name, **model_kwargs)
+
+    return CTranslate2_CausalLM(model, tokenizer)
```
rlouf, line 55:
```suggestion
    return CTranslate2(model, tokenizer)
```

outlines/models/ctranslate2.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __call__(
+        self, input_ids: torch.LongTensor, attention_mask: torch.LongTensor
+    ) -> torch.FloatTensor:
+        # `forward_batch` method of `Generator` accepts `tokens` in a list of list of str
+
+        tokens = [
+            self.tokenizer.tokenizer.convert_ids_to_tokens(iids) for iids in input_ids
+        ]
+        logits = self.model.forward_batch(tokens, return_log_probs=True)
```
rlouf, line 37: `ctranslate2` models takes tokens as inputs instead of ids?

rlouf:
> I also think that we should support `MLC` as more priority than `ctranslate2`, maybe because of more gh-stars or in general community support of [MLC](https://github.com/mlc-ai/mlc-llm).

Makes sense, but it is important that we merge this PR before opening any other one.

outlines/models/ctranslate2.py
arunpatro: Yes, this is a peculiar decision.

PR #220, opened 2023-08-15
Fix ReAct example link path

lukestanley:
Just a fix for a copy pasta. @mlelarge


rlouf:
Thank you for spotting this and opening a PR 🙏

PR #228, opened 2023-08-15
Change default generate behavior from fixed seed to random seed

### LINKED ISSUE #221: Non-seeded generations by default?
Currently the generation is seeded by default.
I can get the exact same output when running the json example on README
```
{
  "name": "ranbelt",
  "age": 26,
  "armor": "chainmail",
  "weapon": "bow",
  "strength": 5
}
```

Not sure if this is the best default behavior. Users might expect LLMs to generate various of outputs.

mondaychen:
resolves #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,6 +195,8 @@ from pydantic import BaseModel, constr
 import outlines.models as models
 import outlines.text.generate as generate
 
+import torch
+
 
 class Weapon(str, Enum):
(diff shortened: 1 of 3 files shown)


rlouf:
Is this ready for review? The tests failed for an unrelated reason: `interegular`'s indexing is non-deterministic. This will be fixed in #178 but is not blocker here

PR #231, opened 2023-08-15
Fixed loading quantized model

minhalvp:
When loading a quantized model e.g.
```model = models.transformers("stabilityai/StableBeluga-7B", device_map="auto", load_in_4bit=True)```
you get an error:
```ValueError: `.to` is not supported for `4-bit` or `8-bit` models. Please use the model as it is, since the model has already been set to the correct devices and casted to the correct `dtype`.```
changed the code so that it only places model on device if the model does not have a quantization config

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -22,10 +22,10 @@ class Transformers:
         device: Optional[str] = None,
     ):
         self.device = device if device is not None else "cpu"
-        if isinstance(getattr(model.config, "quantization_config", None), dict):
-            self.model = model
-        else:
+        if getattr(model.config, "quantization_config", None) is None:
             self.model = model.to(device)
+        else:
+            self.model = model
         self.tokenizer = tokenizer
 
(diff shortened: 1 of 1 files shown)


rlouf:
Tests fail because not all configs have the `quantization_config` attribute.

minhalvp:
did as you suggested. ran the tests and it passed

zaqqwerty:
Does the new accelerate library fix this?  Saw some noise over in this issue thread https://github.com/huggingface/transformers/issues/24540#issuecomment-1611483473

rlouf:
We do not use `model.to` since #264, so I assume the problem is not there anymore? I'm going to test locally and close this PR if everything works fine.

rlouf:
Could you post a minimal example that reproduces the error, and the output?

brandonwillard:
Closing this for now, since I and others have been successfully using BitsAndBytes quantized models in the more recent versions of `outlines`

PR #237, opened 2023-08-16
Add logits method to Sequence class

SamDuffield:
Allows the user to access the next token logits for a given prompt (including with regex!)

```python
unguided_logits = generate.continuation(model, max_tokens=30).logits(prompt)
guided_logits = generate.regex(
    model,
    max_tokens=30,
).logits(prompt)
```

### DIFF
--- a/outlines/text/generate/sequence.py
+++ b/outlines/text/generate/sequence.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Union
+from typing import List, Optional, Tuple, Union
 
 import torch
 
@@ -79,7 +79,7 @@ class Sequence:
         token_ids: torch.LongTensor,
         attention_mask: torch.LongTensor,
         samples: int = 1,
-    ) -> torch.LongTensor:
+    ) -> Tuple[torch.LongTensor, torch.FloatTensor]:
         """Generate one or several tokens that complete the input sequence.
 
         The sampling step consists in using a model to generate next-token
@@ -103,7 +103,8 @@ class Sequence:
         -------
         A tuple with an array of shape `new_batch_shape + (num_tokens+1,)`that
         contains the completed sequences (input token ids and generated token
-        ids).
+        ids) and an array of shape `new_batch_shape + (vocab_size,)` that
+        contains the next token probabilities.
         `new_batch_shape` is computed by removing dimensions of size one in
         `(samples,) + batch_shape`.
 
@@ -134,7 +135,7 @@ class Sequence:
         token_ids = torch.atleast_2d(token_ids.squeeze())
         probs = torch.atleast_2d(probs.squeeze())
 
-        return token_ids
+        return token_ids, probs
 
     def expand_attention_mask(
         self, attention_mask: torch.LongTensor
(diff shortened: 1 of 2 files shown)


rlouf:
I think it would make more sense to return the logprob as part of the `step` method?

Then we could create a custom object that contains both the text output and the logprob. This might do for now, give me some time to think about it.

What are you using this for?

SamDuffield:
> I think it would make more sense to return the logprob as part of the `step` method?
> 
> Then we could create a custom object that contains both the text output and the logprob. This might do for now, give me some time to think about it.

If you include it in the `step` are you thinking to return the logit only for the sampled token or all of the logits. I think it would be useful for the user (at least me 😄 ) to have access to all of the logits if needed.


> What are you using this for?

It's just very useful to be able to analyse the next token logits to e.g. assess the confidence of the model on a True/False question for various questions (and this pairs very well with regex).

SamDuffield:
Ah I see what you mean, `step` already outputs the `probs` as an additional output.

It would be useful to have an additional method with a simple API like
`generate.continuation(model).next_token_logits(prompt)`
or
`generate.continuation(model).next_token_probs(prompt)`

I.e. returns the next token logits or probs without actually doing any sampling.

SamDuffield:
Ok I've thought about this a bit more.

IMO the best API would be to have a `.next_token_logits(prompt)` or `.next_token_probs(prompt)` function that allows the user to examine the distribution without sampling. I think it's good to have this separate from the `step` function which samples and appends the new token, which is a different use case.

With this new function it may also make sense to have `step` only return the `token_ids` and not `probs`. This could also avoid some confusion as the `token_ids` relate to the whole sequence and `probs` only the appended token.

Was there a specific reason to have `step` return `probs` as well?

WDYT? @rlouf @brandonwillard

rlouf:
`next_token_logits` sounds good

SamDuffield:
I'm now thinking that indeed the `probs` might be useful for SMC 😃 

Let's leave `step` returning `probs` for now

outlines/text/generate/sequence.py
```diff
@@ -40,6 +40,38 @@ def is_finished(self, token_ids: torch.LongTensor) -> torch.BoolTensor:
...
+            The input prompt.
+
+        Returns
+        -------
+        An array of shape `(batch_size, vocab_size)` containing the logits
+        (unnormalised log probabilities) for the next token generation.
+
+        """
+        token_ids, attention_mask = self.model.tokenizer.encode(prompt)
+        num_prompt_tokens = token_ids.shape[-1]
```
SamDuffield, line 68: I'm wondering if this is working as intended in the vectorised case.

Surely in the vectorised case, different prompts will have different lengths

SamDuffield: This was my bad, indeed the attention_mask handles it.

I was getting confused seeing different probs on different runs but it was because I was using GPT2 which has Dropout

dpsimpson:
(Also sorry @rlouf I must've mis-clicked. I don't think I meant to remove you as a reviewer)

PR #248, opened 2023-08-18
Add support for Llama2, Palm, Cohere, Replicate Models - using litellm

ishaan-jaff:
This PR adds support for models from all the above mentioned providers using https://github.com/BerriAI/litellm/

Here's a sample of how it's used:

```
from litellm import completion, acompletion

## set ENV variables
# ENV variables can be set in .env file, too. Example in .env.example
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"

messages = [{ "content": "Hello, how are you?","role": "user"}]

# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# async openai call
response = acompletion(model="gpt-3.5-turbo", messages=messages)

# llama2 call
model_name = "replicate/llama-2-70b-chat:2c1608e18606fad2812020dc541930f2d0495ce32eee50074220b87300bc16e1"
response = completion(model_name, messages)

# cohere call
response = completion("command-nightly", messages)

# anthropic call
response = completion(model="claude-instant-1", messages=messages)
```

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -374,8 +374,9 @@ async def call_chat_completion_api(
     num_samples: int,
 ):
     import openai
+    from litellm import acompletion
 
-    response = await openai.ChatCompletion.acreate(
+    response = await acompletion(
         model=model,
         messages=messages,
         max_tokens=max_tokens,
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,7 @@ module = [
     "jinja2",
     "joblib",
     "openai",
+    "litellm",
     "numpy.*",
     "perscache.*",
     "PIL",


ishaan-jaff:
@rlouf @brandonwillard can I please get a review on this PR ?😊

Happy to add more docs /tests if this initial PR looks good !

brandonwillard (changes requested):
We need to carefully consider adding external dependencies like this, especially ones with the kind of overlap that appears to exist between your project and ours.  For instance, we could easily have project direction and/or design interests that would reduce the utility of&mdash;or completely obviate&mdash;an addition like this, even in the near term.

If there's a means of adding an integration for your library that isn't as low-level and only requires an optional dependency, we could probably move forward with that.  Such a thing would necessarily involve some design-level work first, and I personally can't prioritize that at the moment.

PR #256, opened 2023-08-22
Remove unused `from typing import List` in `README.md`

alvarobartt:
Hi to whoever is reading this! 🤗 

## What's in this PR?

This is a simple PR to just remove an import in the `README.md` which is not needed.

Thanks!


zaqqwerty (approved):
LGTM, good catch!

PR #257, opened 2023-08-22
Fix some typos/language in `regex` and `sequence`

jrysana:
Please correct me if I'm wrong about any of the below! But I noticed a few more things that seem mismatched in comments.

- [x] Fixed a typo in `regex/create_proposal` implying that this method is only for integer generation, whereas it seems clear that this method is used for all schemas.
- [x] Fixed some typos where argument names were mismatched or nonexistent in the actual function code.
- [x] Fixed some typos for overall language (e.g. "used to compute" instead of "used to computes") and to clarify potentially confusing comments.
- [x] Fixed a broken link in `README.md`

### DIFF
--- a/outlines/text/generate/regex.py
+++ b/outlines/text/generate/regex.py
@@ -153,7 +153,7 @@ def regex(model, regex_string: str, max_tokens: Optional[int] = None):
     Parameters
     ----------
     model
-        The model to use to compute the next-token logits.
+        The language model to use to compute the next-token logits.
     regex_string
         The regular expression that generated expressions must match.
     max_tokens
@@ -173,7 +173,7 @@ def integer(model, max_tokens: Optional[int] = None):
     Parameters
     ----------
     model
-        The model to use to compute the next-token logits.
+        The language model to use to compute the next-token logits.
     max_tokens
         The maximum number of tokens to generate.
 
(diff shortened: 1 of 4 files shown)


outlines/text/generate/regex.py
```diff
@@ -210,16 +206,16 @@ def choice(model, choices: List[str], max_tokens: Optional[int] = None):
...
 
 def json(model, schema: Union[str, BaseModel], max_tokens: Optional[int] = None):
-    """Generate a text sequence that follows a JSON schema.
+    """Generate a text sequence that follows a JSON schema or Pydantic model.
 
     Parameters
     ---------
     model
-        The model to use to computes the next-token logits.
+        The model to use to compute the next-token logits.
```
jrysana, line 214: This is a nitpick, but I think these comments for `model` args might be more welcoming to some readers with less context, without losing clarity, if we said "the language model" / "the text transformer model" or something, instead of just "the model" - being that "model" is a very vague word and requires previous context to understand. For example, just in this file alone, there is a connotation of "model" as a text transformer model and "model" as a Pydantic data validation model.

zaqqwerty: Hey @jrysana, that sounds like a good idea to me 🙂  would you want to make those changes?

zaqqwerty (approved):
Hi @jrysana, thank you for the PR!  We appreciate your thorough reading of the code.  LGTM, lmk if you'd first like to update based on your suggestion before we merge.

outlines/text/generate/regex.py
jrysana: Sure, pushed @zaqqwerty

PR #265, opened 2023-09-02
Fix bug when partial match is empty

jccalvojackson:
`find_partial_matches` within `create_proposal` could return an empty set. This is not handled.

Also, `create_proposal` expects `self.pstates` to be non empty. If empty `torch.concatenate` fails.


brandonwillard:
Thanks for the patch!  Is there a MWE for this case that we can add as a test?

Issue #192, opened 2023-07-15
Filter vocabulary to avoid non-terminal paths


rlouf:
When a regex has more than one terminal state, it is possible that not all terminal states can be reached with a given vocabulary and that the FSM gets stuck in one of these non-terminal paths. We should therefore filter the vocabulary so it never gets stuck in such a path.

PR #272, opened 2023-09-06
Introduce Numba-based FSM utilities

### LINKED ISSUE #226: Slow Index Building
```python3
from typing import List
from enum import Enum
from pydantic import BaseModel, constr

import outlines.text.generate as generate
import torch

class Weapon(str, Enum):
    sword = "sword"
    axe = "axe"
    mace = "mace"
    spear = "spear"
    bow = "bow"
    crossbow = "crossbow"


class Armor(str, Enum):
    leather = "leather"
    chainmail = "chainmail"
    plate = "plate"


class Character(BaseModel):
    name: constr(max_length=10)
    age: int
    armor: Armor
    weapon: Weapon
    strength: int


model = models.transformers("gpt2", device=torch.device('cuda'))
sequence = generate.json(model, Character)("Give me a character description")
print(sequence)
# {
#   "name": "ranbelt",
#   "age": 26,
#   "armor": "chainmail",
#   "weapon": "bow",
#   "strength": 5
# }

parsed = Character.model_validate_json(sequence)
print(parsed)
```

The above code snippet runs very slowly on my machine. I have a 7900XTX and before I added `device=torch.device('cuda')` it was defaulting to CPU inference only. It seems to running on my GPU now, but the inference is still very slow (takes ~30 seconds to run the above). This could just be a ROCM thing, I'm not entirely sure. I've installed both Torch and Tensorflow with ROCM support. Any idea what might be going on?

### LINKED ISSUE #239: Cache masks computed during regex guidance
This has come up a few times, but we apparently haven't created an issue for it: we can keep the masks in the "index" we build instead of the vocab indices.  That will save on allocation time.

brandonwillard:
This PR introduces Numba JITed FSM utilities with 20x speed-ups over the current pure Python implementations.

It also introduces a more memory efficient "end-to-end" means of producing FSM indices.  I avoided implementing it this way originally because it involves multiple iterations through a vocabulary, but, in order to address some memory-related shortcomings of the CFG indexing approaches tested in #178, this might be the better approach for now.  It's a clear trade-off between processing and memory&mdash;now leaning toward processing&mdash;but, with the JIT speed-ups, it's reasonable.

Closes #226 (for now), closes #239, and should help with #192.

- [x] Tie this into the `Regex` implementation.
- [x] Map end states to tokens in the index, so that there's no need to re-walk the FSM after sampling.
    This was always how it was supposed to work, but our previous prototype didn't implement it.  Since we're updating/replacing that prototype, it might be best to add it now.
- [x] Cache computed masks.
- [x] Investigate Numba compilation and caching.
    We need to make sure that caching works exactly as expected (i.e. only once for all the index-building code).
- [ ] Test out some multi-threading approaches to the end-to-end indexing.
- [ ] Consider using `uint`s for the states, instead of `int64`.
- [x] Consider Python packaging/Numba AOT options.


AL-377:
will it help to speed up the "self.regex_fsm = regex_pattern.to_fsm().reduce()" in outlines 0.0.8，i found when the set the constrain field long like maxLength=1000, it takes very long in regex_fsm construction

rlouf:
Did you try with `0.0.9`?

Issue #276, opened 2023-09-11
Add speculative decoding


SamDuffield:
I'm keen on adding [speculative decoding](https://arxiv.org/abs/2211.17192) to outlines.

Is this something that is being worked on? Otherwise I would be happy to submit a PR but I'd need some advice on the API.

I think the following would be the most desirable so that it works with an arbitrary continuation (e.g. regex)
```python
import outlines.text as text

continuation = text.generate.continuation(model)
speculative_continuation = text.speculative_decode(continuation, draft_model)
output = speculative_continuation(prompt)
```

But I wouldn't be sure how to implement this since we'd likely want to modify the `__call__` method of the `continuation` instance.

brandonwillard:
Yes, I believe this is a feature worth adding.

There are some design questions that need to be answered first.  As you implied, a functional/compositional approach would be nice&mdash;one that allows easy mixing and matching of various logit manipulations (e.g. guided/constrained generation) and sampling approaches.

Feel free to open a draft PR that adds the functionality to a custom `Sequence` subclass.  The next-token sampling is performed in `Sequence.step`, so that's a good starting point.

SamDuffield:
Problem is `Sequence.step` generates a single token whereas speculative decoding generates a batch of sequential tokens and then rejects some of them. I don’t see a nice way of implementing this atm without overwriting `__call__` 😿

brandonwillard:
> Problem is `Sequence.step` generates a single token whereas speculative decoding generates a batch of sequential tokens and then rejects some of them. I don’t see a nice way of implementing this atm without overwriting `__call__` 😿

You can override `Sequence.__call__` to get something running; there's no problem with that.

PR #281, opened 2023-09-15
Add speculative decoding

SamDuffield:
First attempt at adding [speculative decoding](https://arxiv.org/abs/2211.17192)! Discussed in #276 

A few things to note

- It does not support models with different tokenizers, this feels pretty hard to achieve
- It does not currently support vectorised prompt inputs or multiple samples, this probably could be achieved but not sure how desirable since speculative decoding already uses up a lot of parallel resources
- I have tried to make it work for arbitrary `sequence` so including `regex` etc. But in a quite hacky way where I reproduce an instance of the received `sequence`. I think this could be greatly simplified/improved

### DIFF
--- a/outlines/text/generate/speculative_decoding.py
+++ b/outlines/text/generate/speculative_decoding.py
@@ -3,7 +3,7 @@ import inspect
 import torch
 
 from outlines.text.generate.continuation import Continuation
-from outlines.text.generate.sequence import Sequence
+from outlines.text.generate.sequence import Sequence, vectorized_random_choice
 
 
 def parallel_multi_step(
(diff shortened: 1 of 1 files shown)


SamDuffield:
I realise now that I have coded `parallel_multi_step` in a silly overkill way and all of the pseudo-prompts are not needed! For a transformer you can get logits for the whole previous sequence with just a single call to the model 🤯. Might need to go into the `hidden_state` of the transformer output to extract this and also would need to think about how to do so in a way that conforms to arbitrary outlines `sequences` (regex etc)

brandonwillard:
> * It does not support models with different tokenizers, this feels pretty hard to achieve
> * It does not currently support vectorised prompt inputs or multiple samples, this probably could be achieved but not sure how desirable since speculative decoding already uses up a lot of parallel resources

Yeah, don't worry about these for now.  We can follow up in later PRs.

brandonwillard:
Oh, don't forget to set up `pre-commit` locally!

brandonwillard (commented):
OK, so the funnest part of this is going to be testing!  Any ideas for that?

brandonwillard:
Just made a relevant update to the interface in https://github.com/outlines-dev/outlines/pull/318.

PR #287, opened 2023-09-18
Example to generate dating app profiles- combines prompt templating with JSON generation

veezbo:
Let me know if there's any additional information you'd like me to add.

I figure this practical example will be useful to demonstrate more complex JSON generation alongside prompt templating (which is more useful the more complex the JSON structure becomes).


rlouf:
Awesome example! I noticed a comment mentioning that `constr` makes the compilation slow. Would you mind opening an issue so we track this?#272 should help with this.

veezbo:
@rlouf Thanks! I went ahead and ran the pre-commit hook and it passes now. One note on that which I also made in a comment in the code: mypy seems to have some outstanding issues with constrained pydantic typing. See the issue here:  https://github.com/pydantic/pydantic/issues/975

Unfortunately, that means in order to pass the pre-commit hook I had to add a `type: ignore` comment in my code.

As a sanity check, I also ran the pre-commit hook on the [example](https://github.com/outlines-dev/outlines/blob/ff4ebb3d995909df15e0c4b98fa6c9e5475360d0/README.md?plain=1#L191) in the README but unfortunately this failed the pre-commit hook for the same reason and requires a `type: ignore` comment to pass.

I'll also create an issue soon `constr` increasing runtime issue and reference this PR.

brandonwillard (commented):
This is such a great example!

veezbo:
Thanks @brandonwillard! I'm a big fan of your guys' work on this, and see a lot of potential. 

I wanted to try JSON generation with a more complex JSON that would benefit from doing joint generation of multiple fields. Dating profiles (relevant to my day job) seemed to fit the bill.

PR #300, opened 2023-09-29
Use `device_map` value when `device` is unspecified

### LINKED ISSUE #299: `transformers.device = None` default sneakily overwrites `device_map='auto'`
I was trying to allow `accelerate` to load straight to GPU, and use multiple GPUs by passing `device_map='auto'`. But alas, the following line of code overrides it.

https://github.com/outlines-dev/outlines/blob/cae6a534f7b4aa8ea9b64aae2bd9f4335a5f24fc/outlines/models/transformers.py#L139

Now I have this, which is bad because `device='auto'` is an unconventional usage of `device`, which typically refers to `cpu/cuda:0` etc.

```python
    model = models.transformers(
        model_path,
        model_kwargs=dict(
            trust_remote_code=True,
            local_files_only=True,
            device_map='auto',
        ),
        device='auto',
    )
```

The above code saves me :+1:

brandonwillard:
Closes #299

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -159,9 +159,9 @@ def transformers(
     ----------
     model_name
         The name of the model as listed on Hugging Face's model page.
-    device_map
+    device
         The device(s) on which the model should be loaded. This overrides
-        the value passed for `device_map` in `model_kwargs`.
+        the `device_map` entry in `model_kwargs` when provided.
     model_kwargs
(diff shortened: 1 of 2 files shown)


RonanKMcGovern:
Could the docs be updated to indicate how to load with an auto device_map? I'm loading with outlines.models.transformers and it's loading to cpu by default

PR #307, opened 2023-10-01
Fix typos in model usage examples

Barahlush:
Corrected some typos in the documentation which were causing the example code to fail.


brandonwillard:
Much appreciated, @Barahlush!

PR #308, opened 2023-10-02
Remove unnecessary array allocations in generation process and enable caching

brandonwillard:
This PR removes some unnecessary array allocations during the generation process that affect scaling in max tokens and adds KV caching.

Perhaps the biggest non-cache-based change is that the method `Sequence.update_token_ids` has been removed; otherwise, the dimensions for arrays returned by `Sequence.step` are fixed (i.e. no squeezing).  This makes the dimensions in `Sequence.__call__` clearer and allows us to simplify the loop (e.g. no need to duplicate the steps in the auto-regression loop before starting the loop).

### DIFF
--- a/outlines/text/generate/sequence.py
+++ b/outlines/text/generate/sequence.py
@@ -1,13 +1,16 @@
 import math
-from typing import List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 import torch
+
+if TYPE_CHECKING:
+    from outlines.models.transformers import KVCacheType, Transformers
 
 
 class Sequence:
     """Represents a sequence generation method."""
 
-    def __init__(self, model, max_tokens: Optional[int] = None):
+    def __init__(self, model: "Transformers", max_tokens: Optional[int] = None):
         """Create a `Sequence` instance.
 
         Parameters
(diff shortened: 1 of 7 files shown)


outlines/text/generate/sequence.py
```diff
@@ -203,37 +139,53 @@ def __call__(
...
             is_finished[~is_finished] = self.is_finished(
-                updated_token_ids[:, num_prompt_tokens:]
+                token_ids[~is_finished][:, num_prompt_tokens:]
             ).flatten()
 
+        # TODO: It looks like batch shapes with more than one dimension aren't
+        # accepted by the decoder, so is it really that we support a one
+        # dimensional batch shape given by the number of prompts or
+        # (exclusively) `samples`?  If so, that would imply we can simplify the
+        # logic above and add an explicit check/`assert`.
```
brandonwillard, line 185: @rlouf, do we ever expect batch shapes greater than one dimension?  

We have a potential dimension given by `samples` and another given by the number of prompts.  This could be interpreted as a batch with shape `(samples, len(prompts))` or a batch wich shape `(samples * len(prompts),)` (i.e. flattened).  The tokenizer will only decode the latter (i.e. a list of lists), so really don't have a choice unless we update something to add support for more than two dimensional inputs in that decoder logic.

Issue #309, opened 2023-10-02
JSON schema used in generation should not require optional annotations


mattkindy:
In [outlines/text/json_schema.py#expand_json_schema](https://github.com/outlines-dev/outlines/blob/main/outlines/text/json_schema.py#L95), part of the regex generation process attempts to grab the `title` property from the current schema/sub-schema whenever it encounters a schema with `properties`:

https://github.com/outlines-dev/outlines/blob/5af26cb93905605451ffe3a2594a7c60316a7505/outlines/text/json_schema.py#L136

However, [the current JSON schema reference](https://json-schema.org/understanding-json-schema/reference/annotations#annotations) notes:

> JSON Schema includes a few keywords, that aren't strictly used for validation, but are used to describe parts of a schema. None of these "annotation" keywords are required, but they are encouraged for good practice, and can make your schema "self-documenting".

> The `title` and `description` keywords must be strings. A "title" will preferably be short, whereas a "description" will provide a more lengthy explanation about the purpose of the data described by the schema.

> The `default` keyword specifies a default value. This value is not used to fill in missing values during the validation process. Non-validation tools such as documentation generators or form generators may use this value to give hints to users about how to use a value. However, `default` is typically used to express that if a value is missing, then the value is semantically the same as if the value was present with the default value. The value of `default` should validate against the schema in which it resides, but that isn't required.

Here's a simple example where this fails:
```python
schema = """{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "type": "object",
    "properties": {
        "test": {
        "type": "integer"
        }
    },
    "required": [
        "test"
    ]
}"""
return generate.json(
    self.model,
    schema,
)(prompt)
```

```bash
File "/usr/local/lib/python3.11/site-packages/outlines/text/generate/regex.py", line 228, in json
regex_str = build_regex_from_schema(schema)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/outlines/text/json_schema.py", line 35, in build_regex_from_schema
schedule = build_schedule_from_schema(schema)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/outlines/text/json_schema.py", line 80, in build_schedule_from_schema
schema = expand_json_schema(schema, definitions)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #311, opened 2023-10-02
Fix JSON schema generation to accept schemas without 'title' property

mattkindy:
## Overview
In `expand_json_schema` -- used to denormalise the JSON schema -- only access and set the `title` property if present to avoid errors for schemas that have not set this optional property.

Fixes https://github.com/outlines-dev/outlines/issues/309


brandonwillard (commented):
This looks good, too; just needs to be rebased.

brandonwillard (approved):
Thanks again, @mattkindy!

Issue #310, opened 2023-10-02
JSON schema parsing doesn't handle `$ref` properly


mattkindy:
In [outlines/text/json_schema.py#expand_json_schema](https://github.com/outlines-dev/outlines/blob/main/outlines/text/json_schema.py#L95), during the regex generation process, when a `$ref` is encountered in a JSON schema,

https://github.com/outlines-dev/outlines/blob/5af26cb93905605451ffe3a2594a7c60316a7505/outlines/text/json_schema.py#L120

the `definitions` (corresponding to `$def`) object is searched:

https://github.com/outlines-dev/outlines/blob/5af26cb93905605451ffe3a2594a7c60316a7505/outlines/text/json_schema.py#L122

While `$def` <-> `$ref` correspondence is a reasonable practice, this doesn't support a number of other ways in which [$ref](https://json-schema.org/understanding-json-schema/structuring#dollarref) is used. Among other things, this prevents use of `$ref`s like `#/properties/myEquivalentProperty` (which is generated by tools like `zod`'s `zodToJsonSchema` when reusing a schema definition) or [recursive refs](https://json-schema.org/understanding-json-schema/structuring#recursion).

Here's an example where this fails:
```python
schema = """{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "type": "object",
    "properties": {
        "test": {
          "type": "integer"
        },
        "test2": {
          "$ref": "#/properties/test"
        }
    }
}"""
return generate.json(
    self.model,
    schema,
)(prompt)
```

```bash
File "/usr/local/lib/python3.11/site-packages/outlines/text/generate/regex.py", line 228, in json
regex_str = build_regex_from_schema(schema)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/outlines/text/json_schema.py", line 35, in build_regex_from_schema
schedule = build_schedule_from_schema(schema)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/outlines/text/json_schema.py", line 80, in build_schedule_from_schema
schema = expand_json_schema(schema, definitions)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/outlines/text/json_schema.py", line 120, in expand_json_schema
definitions[value["$ref"]], definitions
[...]

PR #312, opened 2023-10-02
Modify JSON schema $ref resolution strategy to support resolving values internal to the schema

mattkindy:
## Overview
Previously, only `$ref` that were directed to `#/$defs` could be resolved via the use of a pre-loaded `definitions` variable. This PR modifies the behaviour of `$ref` resolutions by instead using a `resolver` function, which is generated to scope to a particular schema or subschema.
 - This is still preloaded with `#/$defs` as before
 - `$ref` keys are split to traverse through the scoped schema. When found, they are cached for re-use to avoid re-traversing
 - In `expand_json_schema`, `$id` is an identifier of a scope change, and a new resolver linked to the schema for that scope is used for its children (e.g. for supporting [bundling](https://json-schema.org/understanding-json-schema/structuring#bundling))
 - [Recursion](https://json-schema.org/understanding-json-schema/structuring#recursion) is supported -- note that no effort is made to identify if this recursion is valid or could lead to infinite loops
 
Fixes https://github.com/outlines-dev/outlines/issues/310 for local references (i.e. doesn't attempt to address remote schemas)

### DIFF
--- a/outlines/text/json_schema.py
+++ b/outlines/text/json_schema.py
@@ -1,7 +1,7 @@
 import itertools
 import json
 import re
-from typing import Dict
+from typing import Callable, Dict
 
 STRING_INNER = r'(?:[^"\\\x00-\x1f\x7f-\x9f]|\\.)'
 STRING = f'"{STRING_INNER}*"'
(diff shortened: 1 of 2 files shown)


brandonwillard (approved):
This is great, and much appreciated, @mattkindy!

PR #316, opened 2023-10-10
Support recursive arrays in JSON when an item is an array

AL-377:
I have noticed that the origin outlines doesn't support recursive array type. To be clear, here is a example：

```python
from outlines.text.json_schema import build_regex_from_schema
from typing import List
import json 

class ItemModel(BaseModel):
    name:str

class ArrayModel(BaseModel):
    items:List[List[ItemModel]]

test_json = \
{
    "items": [[{"name":"item1"},{"name":"item2"}]], 
}


json_schema = ArrayModel.model_json_schema()
print(json.dumps(json_schema,indent=2))

regex = build_regex_from_schema(json.dumps(json_schema))
print(regex)

parsed = ArrayModel.model_validate_json(json.dumps(test_json))
print(parsed)

```

Run the code above, then we can get the json schema of  `ArrayModel` , but we will get a error from `build_regex_from_schema` when we meet the `List[List[]]` type property.

And now, I have modified the code in outlines and it works. You can check the `test_json_schema.py`

### DIFF
--- a/outlines/text/json_schema.py
+++ b/outlines/text/json_schema.py
@@ -125,6 +125,18 @@ def build_schedule_from_schema(schema: str):
     return reduced_schedule
 
 
+def expand_item_json_schema(expanded_property: Dict, resolver: Callable[[str], Dict]):
+    """Recursively expand "$ref"s in "item"s."""
+    if "items" not in expanded_property.keys():
+        return
+    elif "$ref" in expanded_property["items"]:
+        expanded_property["items"] = expand_json_schema(
+            resolver(expanded_property["items"]["$ref"]), resolver
+        )
+    else:
+        expand_item_json_schema(expanded_property["items"], resolver)
+
+
 def expand_json_schema(
     raw_schema: Dict,
     resolver: Callable[[str], Dict],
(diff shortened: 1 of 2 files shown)


.DS_Store
brandonwillard, line 1: Looks like this file was accidentally added.

test.py
brandonwillard, line 1: These tests need to be added to the existing test files in `pytest` format.

brandonwillard (changes requested):

.DS_Store
AL-377: OK，I have delete that file

test.py
AL-377: Sorry I have missed the pre-commit step before, and now it's done

AL-377:
It is strange that the pytest here seems to fail in`test_continuation.py`, I have passed the `test_continuation.py` in my local machine.

image

AL-377 (commented):
The changes have done

brandonwillard (approved):

brandonwillard:
@AL-377, this was a great fix/enhancement; much appreciated!  I've added some minor refactoring and it should be ready to merge once everything finishes in CI.

AL-377:
> @AL-377, this was a great fix/enhancement; much appreciated! I've added some minor refactoring and it should be ready to merge once everything finishes in CI.

👍👍👍

AL-377:
> Thank you for the contribution and sorry for the delay in reviewing. Will get to it soon.

Never mind！

PR #318, opened 2023-10-12
Allow configurable sampling steps

brandonwillard:
This PR allows one to specify sampler functions that decide exactly how generated token sequences are sampled/produced from the logit values.  In doing so, this PR also adds a greedy "sampler" function (i.e. one that chooses the max logit values).

### DIFF
--- a/outlines/text/generate/sample.py
+++ b/outlines/text/generate/sample.py
@@ -1,0 +1,111 @@
+from typing import Protocol
+
+import torch
+
+
+class Sampler(Protocol):
+    def __call__(
+        self, logits: torch.DoubleTensor, samples: int, rng: torch.Generator
+    ) -> torch.DoubleTensor:
+        ...
+
+
+def greedy(logits: torch.DoubleTensor, samples: int, *_) -> torch.DoubleTensor:
+    """Greedy Sampling algorithm.
+
+    Greedy sampling consists in choosing the token with the largest
+    likelihood at every step.
+
+    Parameters
+    ----------
+    logits
+        A tensor of shape ``(n_seqs, vocab_size,)`` that represents the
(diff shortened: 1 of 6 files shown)


outlines/text/generate/sample.py
```diff
@@ -0,0 +1,111 @@
...
+    rng
+        A random number generator.
+
+    Returns
+    -------
+    The ids of the sampled tokens having shape ``(samples, n_seqs)``.
+
+    """
+    probs = torch.nn.functional.softmax(logits, dim=-1)
+    # next_token_ids = torch.multinomial(probs, num_samples=samples, generator=rng)
```
rlouf, line 70: `torch.multinomial` accepts matrices for `probs`: https://pytorch.org/docs/stable/generated/torch.multinomial.html

tests/text/generate/test_sequence.py
```diff
@@ -67,38 +67,6 @@ def __hash__(self):
         return id(self)
 
 
-def test_vectorized_random_choice():
```
rlouf, line 70: Shouldn't this function still be tested?

brandonwillard: It was moved.

outlines/text/generate/sample.py
brandonwillard: I have a few follow-up updates to add shortly; I'll put that in place alongside those.

brandonwillard: While `torch.multinomial` does work for matrices, we still need to check to the results against the higher dimensional tensor inputs in the tests (and update the tests that rely on RNG seeds), so the change should come in a follow-up.

namin:
👍 :)

PR #328, opened 2023-10-29
Make sure that indices are available on all relevant devices

brandonwillard:
This PR transfers all is-sequence-finished indices to the relevant devices during token generation.


namin:
👍 
I confirm that this PR fixes my issue with running a BitsAndBytes-configured model on a multi-GPU system.

PR #334, opened 2023-11-06
Fix math error in sampling example

mwesthelle:
If I'm 6 and my sister is half my age, this means she's 3. 64 years later, when I'm 70, my sister is going to be 67, as her age isn't always my age divided by 2.

### DIFF
--- a/examples/sampling.ipynb
+++ b/examples/sampling.ipynb
@@ -135,7 +135,7 @@
    "id": "1a895b6d-d4d4-40f9-9156-24ba7e21cc08",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


rlouf:
Thank you for the PR! Unfortunately this is the model's answer and I don't think we should correct it?

mwesthelle:
> Thank you for the PR! Unfortunately this is the model's answer and I don't think we should correct it?

This is not the model's output, this is markdown text in the notebook :)

brandonwillard (approved):

brandonwillard:
> > Thank you for the PR! Unfortunately this is the model's answer and I don't think we should correct it?
> 
> This is not the model's output, this is markdown text in the notebook :)

Yeah, it looks like a mistake in an expository statement about the correct answer.

Issue #330, opened 2023-10-31
Recursive JSON Schemas


sahewat:
Recursive Pydantic definitions seem unsupported for lists, unions, and optionals. My understanding is these are the basic use cases.

A reproducible example is provided below:
```python
import json

from typing import List, Optional, Union
from pydantic import BaseModel

from outlines.text.json_schema import build_regex_from_schema

class TaskOptional(BaseModel):
    subtask: Optional['TaskOptional']

class TaskWrapperOptional(BaseModel):
    task: TaskOptional

class TaskUnion(BaseModel):
    subtask: Union['TaskUnion', None]

class TaskWrapperUnion(BaseModel):
    task: TaskUnion

class TaskList(BaseModel):
    subtask: List['TaskList']

class TaskWrapperList(BaseModel):
    task: TaskList

TaskWrapperOptional.model_rebuild()
TaskWrapperUnion.model_rebuild()
TaskWrapperList.model_rebuild()

regex = build_regex_from_schema(json.dumps(TaskWrapperOptional.model_json_schema()))
# NotImplementedError: 
# @ outlines/text/json_schema.py:308 

regex = build_regex_from_schema(json.dumps(TaskWrapperUnion.model_json_schema()))
# NotImplementedError: 
# @ outlines/text/json_schema.py:308 

regex = build_regex_from_schema(json.dumps(TaskWrapperList.model_json_schema()))
# RecursionError: maximum recursion depth exceeded while calling a Python object
# @ outlines/text/json_schema.py:174 
```

I'd be interested in adding this functionality but I'm unsure as to what an "unrolled" recursive definition would look like in terms of the generated regex.

rlouf:
Thank you for opening an issue! It looks like something we should be able to support. Do you mind pasting the JSON schema for these 3 models here?

sahewat:
In all cases, the schema is available through `TaskWrapperxxx.model_json_schema()`. I'll post an example of each here as well.



`TaskWrapperOptional`
```json
{
    "$defs": {
        "Task": {
            "properties": {
                "name": {
                    "title": "Name",
                    "type": "string"
                },
                "subtasks": {
                    "default": [],
                    "items": {
                        "$ref": "#/$defs/Task"
                    },
                    "title": "Subtasks",
                    "type": "array"
                }
            },
            "required": [
                "name"
            ],
            "title": "Task",
            "type": "object"
        },
        "TaskOptional": {
            "properties": {
                "subtask": {
[...]

Issue #215, opened 2023-08-12
Implement JSON schema field constraints


rlouf:
We can specify constraints for the different fields in the [JSON schema specification](https://json-schema.org/understanding-json-schema/reference/). Only `maxLength` for strings is currently implemented. Remaining:

## Strings

- [x] `minLength`
- [x] `pattern`

and the default formats that can be specified via the `format` keyword:

- [ ] `date-time`
- [ ] `time`
- [ ] `date`
- [ ] `duration`
- [ ] `email`
- [ ] `idn-email`
- [ ] `hostname`
- [ ] `idn-hostname`
- [ ] `ipv4`
- [ ] `ipv6`
- [ ] `uuid`
- [ ] `uri`
- [ ] `uri-reference`
- [ ] `iri`
- [ ] `iri-reference`
- [ ] `uri-template`
- [ ] `regex`

## Numeric types

- [ ] `multipleOf`
- [ ] `minimum`
- [ ] `exclusiveMinimum`
- [ ] `maximum`
- [ ] `exclusiveMaximum`

## Arrays

- [ ] ~`minItems`~
- [ ] ~`maxItems`~
- [ ] ~`uniqueItems`~ (may only be applicable dynamically)
- [ ] Set length

## Tuples

See https://json-schema.org/understanding-json-schema/reference/array#tupleValidation

## Required fields

We should handle optional fields as well, i.e. those not specified in the `required` field of the schema.

brandonwillard:
Here are some examples of integer range constraints expressed as regular expressions: https://stackoverflow.com/a/34680927/3006474, https://3widgets.com/

CarloNicolini:
What is the status of this? Are length of tuples/lists also implemented?

fire:
To be honest I switched to ggml’s ebnf for grammar constraints.

brandonwillard:
> To be honest I switched to ggml’s ebnf for grammar constraints.

How does an EBNF grammar provide these constraints?

Rassibassi:
While playing around trying to validate meta schemas, I ran into `$recursiveRef` or `$dynamicRef`, any thoughts on implementation of recursive schemas? I'd like to give it a try

`$recursiveRef` found in:
https://json-schema.org/draft/2019-09/schema
https://json-schema.org/draft/2019-09/meta/core

`$dynamicRef` found in:
https://json-schema.org/draft/2020-12/schema
https://json-schema.org/draft/2020-12/meta/core

[...]

PR #342, opened 2023-11-08
Improve the coverage of the JSON Schema specification

rlouf:
In this PR I simplify the codebase by using the `referencing` library to dereference the specification. We also check the validity of the schema before compiling it. We add coverage for the following:

- Situations where `minLength` and `maxLength` are both defined;
- The `pattern` keyword that allows to constrain the string to follow a regular expression;
- `oneOf`, `allOf`, `anyOf` are now all properly covered;
- `type` given an array value
- `array` without a specified type
- `enums` without a specified type

## TODO

- [x] More testing
- [x] ~Handle optional fields (those not specified in "required")~
- [x] ~`minItems` for arrays~
- [x] ~`maxItems` for arrays~

## Note

We cannot handle `minItems` and `maxItems` because of interegular's lack of support for lookaheads:

```
 interegular.patterns.Unsupported: Group can not have lookbacks/lookaheads that go beyond the group bounds.
```

## Related issues

#330 #215


janvandervegt-db:
Regarding the `minItems` and `maxItems` lookaheads issue, would it be possible if both `minItems` and `maxItems` are the same value? As you could generate a regular expression that "hardcodes" the number of items?

rlouf:
> Regarding the `minItems` and `maxItems` lookaheads issue, would it be possible if both `minItems` and `maxItems` are the same value? As you could generate a regular expression that "hardcodes" the number of items?

We could, would that be useful?

janvandervegt-db:
For our use case that would be super helpful. We are using LLMs to synthesize a lot of structured data. Frequently a single input maps to a list of different outputs, the exact number is not super important but it needs to be at least a few. Doing them in the same call leads to higher diversity in the outputs and much simpler pipelines. 

Two alternatives for us would be:
- Manually provide hardcoded properties for each of the entries (item1, item2, ...)
- Manipulate the JSON schema generated by Pydantic to dynamically do this before passing it to `outlines` and then include custom parsing logic

Both of these options are quite ugly compared to restricting the guided generation to a fixed number of items in the array.

rlouf:
Noted, I added the feature. Don't hesitate to open an issue if there is something missing.

brandonwillard (approved):

PR #343, opened 2023-11-09
Add integration for `AutoGPTQ` and `AutoAWQ`

### LINKED ISSUE #245: using quantized models from llama
I have the model derived from llama 7bthat I usually call like:

```python
from transformers import StoppingCriteria, AutoModelForCausalLM, AutoTokenizer, StoppingCriteriaList, TextIteratorStreamer
from auto_gptq import AutoGPTQForCausalLM

tokenizer = AutoTokenizer.from_pretrained('./modelo/', padding_side="right", use_fast=True)
modelo = AutoGPTQForCausalLM.from_quantized('./modelo/',
                        model_basename="model_gptq_model-4bit-128g",
                        max_position_embeddings=512,
                        use_safetensors=True,
                        device="cuda:0",
                        use_triton=False,
                        quantize_config=None
        )
```

It feels like it would have a transformers way to be called into outlines, but I can't figure it out... anyone faced this problem?

I basically want to do regex validation:

```python
unguided = generate.continuation(modelo, max_tokens=200)('\n'.join(template))
guided = generate.regex(modelo, combined_regex, max_tokens=200)(
    '\n'.join(template)
)
```

### LINKED ISSUE #270: Feature request: GGML/GGUF support
GGML/GGUF model versions are significantly faster and lighter compared to their equivalent pytorch-format models.
It would be great to add to the documentation how to interface with these models (not sure how to go about it).
I imagine this could be possible without necessarily writing the cpp version of this library.

rlouf:
Closes #245. Closes #270

### DIFF
--- a/outlines/models/__init__.py
+++ b/outlines/models/__init__.py
@@ -5,5 +5,7 @@ completion, diffusers, etc.) and use routing functions everywhere else in the
 codebase.
 
 """
+from .awq import awq
+from .gptq import gptq
 from .openai import OpenAI, openai
 from .transformers import Transformer, transformers
--- a/outlines/models/__init__.py
+++ b/outlines/models/__init__.py
@@ -6,4 +6,4 @@ codebase.
 
 """
 from .openai import OpenAI, openai
-from .transformers import Transformers, transformers
+from .transformers import Transformer, transformers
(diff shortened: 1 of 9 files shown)


rlouf:
Not integrating `ctransformers` because of issues initializing the model: https://github.com/marella/ctransformers/issues/154

Issue #353, opened 2023-11-12
Minor bugs in `babyagi.py` and `meta_prompting.py` example scripts


HerrIvan:
### Describe the issue as clearly as possible:

There a couple of easy fixable in those two sample scripts.

- `baby_agi`: the scripts does not call the two defined formatting functions. That causes an exception to be triggered.

- `meta_prompting`: some misplaced or missing stop criteria prevent the model to generate complete answers.

### Steps/code to reproduce the bug:

```python
python babyagi.py
python meta_prompting.py
```

### Expected result:

```shell
babyagi.py:
----
*****TASK LIST*****

 • Find a repeatable, low-maintainance, scalable business.

*****NEXT TASK*****

Educate yourself on the basics of investment and finance to enhance your understanding of index funds, ETFs,

*****TASK RESULT*****

To achieve your objective of becoming rich while doing nothing, I recommend exploring the following business idea: 

Investing in index funds or exchange-traded funds (ETFs). This approach requires minimal effort and provides a passive income stream. Index funds are designed to track the performance of a particular market index, such as the S&P 500, while ETFs are similar but can be bought and sold throughout the trading day like stocks. By investing in a diverse range of index funds or ETFs, you can benefit from overall market growth and minimize the risk associated with individual stocks. 

To make this business scalable, you can gradually increase your investment portfolio by allocating a percentage of your income towards index funds or ETFs regularly. Implementing an automatic investing plan, such as setting up automatic monthly contributions, allows you to invest consistently without any manual effort. By reinvesting any dividends received, your investment can compound over time and potentially generate substantial returns.

It is important to note that investing in the stock market carries some level of risk, and past performance is not indicative of future results.

*****TASK LIST*****

 • Develop a financial plan to allocate a percentage of your income towards index funds or ETFs regularly, considering factors such as risk tolerance, financial goals, and time horizon.
...
---

meta_prompting.py
for instance, the second example
---
Directions: In the following question, a related pair of words or phrases is followed by five pairs of words or phrases. Choose the pair that best expresses a relationship similar to that in the original pair. 
BRAGGART :: MODESTY
A) FLEDGLING : EXPERIENCE
B) EMBEZZLER : GREED
C) WALLFLOWER : TIMIDITY
D) INVALID : MALADY
E) CANDIDATE : AMBITION

    
Let's solve this problem by splitting it into steps.Step 1: Identify the relationship between the original pair of words/phrases.
[...]

PR #354, opened 2023-11-12
fixes BUG #353. Example scripts don't run as expected.

HerrIvan:
Fixes example scripts.

- Adds calls to *_fmt functions in `babyagi.py`, s.t. the script can work properly without triggering an Exception.
- Removes one stop criterion and increases max_tokens in a few calls, s.t. the answers by the GPTs are not truncated.


HerrIvan:
Sorry, my bad. Now it passes the pre-commit tests.

rlouf:
Thank you for contributing!

PR #358, opened 2023-11-12
Implement `text.generate.format` constrained generation

### LINKED ISSUE #346: Type-guided generation
We currently support regex-guided generation, so that to return a date in the Python datetime format you would do:

```python
import outlines.generate as generate

model = ...
result = generate.regex(model, "r'\s+(?=\d{2}(?:\d{2})?-\d{1,2}-\d{1,2}\b)'")
```

Instead we would like to be able to pass Python types:

```python
from typing import List
import datetime

import outlines.generate as generate

model = ...
result = generate.type(model, datetime.datetime)
result = generate.type(model, List[str])
result = generate.type(model, int)
```

rlouf:
`text.generate.format` takes a model, a Python type and returns a generator which, given a prompt, will output text in the format described by the Python type:

```python
from datetime import datetime

from outlines import models
from outlines import text


result = text.generate.format(model, datetime.date)("Which day is today?")
# 2023-11-12
```

Closes #346.

## TODO

- [x] `int`
- [x] `float`
- [x] `datetime.datetime`
- [x] `datetime.date`
- [x] `datetime.time`
- [x] `bool`
- [ ] Add some documentation + README

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,27 @@ print(parsed)
 
 The method works with union types, optional types, arrays, nested schemas, etc. Some field constraints are [not supported yet](https://github.com/outlines-dev/outlines/issues/215), but everything else should work.
 
+### Open functions
+
+Outlines can infer the structure of the output from the signature of a function. The result is a dictionary, and can be passed directly to the function using the usual dictionary expansion syntax `**`:
+
+```python
+from outlines import models
+from outlines import text
+
+def concat(a: int, b: int):
+    return a + b
+
+model = models.transformers("mistralai/Mistral-7B")
+generator = text.generate.json(model, add)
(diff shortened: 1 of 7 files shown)


outlines/text/format.py
```diff
@@ -0,0 +1,18 @@
...
+@singledispatch
+def format_to_regex(_):
+    raise ValueError(
+        f"{type} is not currently supported by Outlines. Please open an issue if you'd like to see it supported."
+    )
+
+
+@format_to_regex.register(type(float))
+def float_format_to_regex(_):
+    return r"([+-]?((0|[1-9]+)([.][0-9]*)?)|([.][0-9]+))"
```
rlouf, line 13: Floats in Python can be of the form e.g. `1e9`

Issue #361, opened 2023-11-13
Migrate OpenAI integration to v1.0 of their SDK


rlouf:
v1.0 introduces many breaking changes: https://github.com/openai/openai-python/discussions/742. Most of which should make our life easier in the library.

PR #364, opened 2023-11-15
Modify the openai model to conform to the new openai sdk v1.0.0

RobinPicard:
This PR aims at solving this issue: https://github.com/outlines-dev/outlines/issues/361

The new Python sdk of Openai allows us to instantiate a client that we can then use for API calls. I thought it would make sense to put the instantiation of the client (and the import of the openai library) in the `__init__` of the `OpenAIAPI` class. 

An issue I ran into was that the functions `call_chat_completion_api` and `call_completion_api` would then receive the client as an argument which would cause problems with the `cache` decorator. I'm not sure what's the best way to handle that. I created two intermediary functions `cached_call_completion_api` and `cached_call_chat_completion_api` in the `__init__` of the `OpenAIAPI` class as it looked like the solution that required the fewest modifications to the existing logic.

The new sdk handles the retry logic itself so I removed the `retry` decorator (and added a `max_retries` argument to `OpenAIAPI` instead)

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -1,20 +1,17 @@
 """Integration with OpenAI's API."""
 import functools
 import os
-from typing import Callable, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Union
 
 import numpy as np
-from tenacity import (
-    retry,
-    retry_if_exception_type,
-    stop_after_attempt,
-    wait_random_exponential,
-)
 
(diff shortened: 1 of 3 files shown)


rlouf:
@RobinPicard could you allow me to push to your branch? The tests are failing because `transformers` is using a deprecated Pydantic decorator and our test suite errors on warning.

rlouf:
This works locally, thank you so much for this great contribution!

Issue #335, opened 2023-11-06
Add temperature, top-k and top-p logits processors


SimJeg:
Hello,

In [outlines.text.generate.sample](https://github.com/outlines-dev/outlines/blob/main/outlines/text/generate/sample.py), only `greedy` and `multinomial` methods are available. Would it be possible to implement some classical features in the `multinomial` function such as `top_k`, `top_p` or `temperature` ? 

Thanks for your work,
Simon

brandonwillard:
Yes, definitely!

rlouf:
> Also, we need to add beam search.

Tracked by #258

rlouf:
#533 introduces a small refactor of the samplers that would allow to pass the `top_k` and `top_p` arguments to the `multinomial` sampler. We could pass the temperature using the same mechanism as well

rlouf:
top-k was added in #642, and I opened separate issues for the other requests: #645 and #643.

Issue #163, opened 2023-06-25
Add `vLLM` integration for the generation API


rlouf:
[vLLM](https://github.com/vllm-project/vllm) allows for subtantially faster inference via smart management of the KV-cache. The library proposes seemless integration with some of the most popular HuggingFace models. We suggested in #150 to re-use some of the ideas in this paper/library for Outlines.

In a first iteration we can dispatch the sequence generation functions introduced in #139 to use vLLM's user-facing APIS, although the use would be limited to the simplest generation methods. Longer term, we should look into vLLM's internals and see if we can make Outlines compatible.

louisoutin:
Agree, would be nice to see an integration for vLLM into outlines

amir-in-a-cynch:
Is there any progress on this? I'm using vllm with outlines - I'd be willing to help out / try and put out a PR if there's interest in any community contribution here.

amir-in-a-cynch:
> There's always an interest in community contributions :) However, this requires substantial changes in Outlines' codebase that has far-reaching design implications. We need a bit more time. After that design change the integration should be very straightforward.

Oh absolutely. Not trying to nag. OSS work is voluntary and time consuming. Just showing interest, and willing to help out if it'd accelerate things. Thanks for the update!

simon-mo:
Hi @rlouf, outline is really awesome! 

I'm from the vLLM team and I'm quite excited about guideline's approach. We recently added the `logits_processors` API and I tried to integrate vLLM with outlines. Here's a hacky version that I got working

```python
import outlines.models as models
from outlines.text.generate.regex import Regex

from vllm import LLM, SamplingParams
import torch

prompts = [
    "What is the IP Address of Google",
]

# We are not using this model for actual inference. But it seems to be required for Regex class.
model = models.transformers("gpt2-medium", device="cuda")

regex_processor = Regex(
    model,
    regex_string=r"((25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(25[0-5]|2[0-4]\d|[01]?\d\d?)",
    max_tokens=16,
)

def create_proposal(token_ids, logits):
    token_ids = torch.Tensor([token_ids]).long().to("cuda")
    return  regex_processor.create_proposal(token_ids, logits)

sampling_params = SamplingParams(
    logits_processors=[create_proposal],
    max_tokens=16,
)

# Create an LLM in vLLM.
llm = LLM(model="gpt2-medium")

for _ in range(10):
[...]

Issue #155, opened 2023-06-21
Add a `Constrained` generation method


rlouf:
Using [SMC sampling](https://arxiv.org/abs/2306.03081v1) we can generate sequences that follow arbitrary constraints. All we need is a function that takes previously-generated tokens, a possible completion and returns a boolean. For instance this example from [LlamPPL](https://github.com/probcomp/LLaMPPL) to constrain the sequence generated to not have longer that are more than 5 letters long:

```python
def can_follow(str_so_far, s):
    if isinstance(s, llp.Token):
        s = str(s)
    if len(s.strip()) > 5:
        return False
    if len(s.strip()) == 0:
        return True
    if not s[0].isalpha():
        return True
    if len(str_so_far) == 0:
        return True # First token, can be alphanumeric
    words = str_so_far.split()
    if len(words) >= 1 and len(words[-1]) + len(s) <= 5:
        return True
    else:
        return False
```

I propose to add a `Constrained` subclass to `Sequence` :

```python
import outlines.models as models
import outlines.text as text

model = models.transformers("gpt2")
state = text.constrained(model, can_follow)("Prompt")
```

We will need to add  a `create_proposal` and a `reweigh` method to `Sequence`.

Issue #53, opened 2023-04-16
Return extra information with model calls


rlouf:
We currently return the completed prompt in addition to the model's return value. However models often return more information than this: the total `logprob` value for the completion, API usage stats, etc. Instead of simply returning the completed prompt we should return an object that contains more information, e.g.:

```python
from typing import NamedTuple

class OpenAICompletionInfo(NamedTuple):
    prompt: str
    completion: str
    logprob: str
    api_usage: OpenAICompletionUsageInfo

model = text.completion("openai/davinci")
answer, info = model("Who was Arianna Rosenbluth?")
```

PR #366, opened 2023-11-15
Refactor the sequence generation

### LINKED ISSUE #185: Make sure that the new API works with arbitrary arrays as inputs
The current API allows to pass arrays of prompts with an arbitrary shape to models, get an array in return. This also works with multiple samples. This is useful [to implement Tree of Thoughts](https://github.com/normal-computing/outlines/issues/176#issuecomment-1632604261), and this should be possible anyway. We need to make sure the new API works in the same way.

### LINKED ISSUE #216: Allow for arguments to be passed to the base model
looking at auto-gptq code I spotted that it already contains kv-caching : 
https://github.com/PanQiWei/AutoGPTQ/blob/main/auto_gptq/nn_modules/fused_llama_attn.py

would be really nice to let users utilize this.

### LINKED ISSUE #247: Please add streaming output

### LINKED ISSUE #266: Batching
From my testing of using generate.choice, batching does not work (even though the type hint indicates that it accepts `List[str]` which is a bit confusing). I am wondering whether that hs a theoretical limitation or whether that is something that can still be implemented. If it is something that can be implemented, can you give some pointers where to start? I can work on this at the start of October.

### LINKED ISSUE #317: Update logits in-place during guided generation instead of masking
We should consider updating the logit arrays in-place instead of creating separate mask arrays and adding those to the logits.

### LINKED ISSUE #340: Assertion error when running on batches

When generating JSON on a batch of prompts, it will frequently run into an assertion error in `text/generate/regex.py` on line 156.

`assert last_state > -1`

Followed is an end-to-end example that runs into the problem when it reaches batch size 2.

```python
from timeit import default_timer as timer
from typing import List
import torch

from outlines.models.transformers import transformers
import outlines.text.generate as generate

model = transformers(model_name="mosaicml/mpt-7b-instruct", device="cuda:0", model_kwargs={"trust_remote_code": True})


rng = torch.Generator(device="cuda:0")
rng.manual_seed(42)


class HouseAd(BaseModel):
    city: str
    amount: int
    description: str

class HouseAds(BaseModel):
    house_ads: List[HouseAd]


p = "Generate 5 house ads. Use the `house_ads` property with an array of JSON objects, each containing the `city`, `amount` and `description` properties."


def time_generation(obj, max_tokens, n=1, batch=None):
    print(str(obj.__name__), max_tokens, batch)
    generator = generate.json(model, obj, max_tokens=max_tokens)
    start = timer()
    for _ in range(n):
        if batch is None:
            inp = p
        else:
            inp = [p] * batch
        sequence = generator(inp, rng=rng)
        print(sequence)
    end = timer()
    time = (end - start) / n
    print(str(obj.__name__), max_tokens, batch, time)
    return (end - start) / n


results = {"400_na": time_generation(HouseAds, 500),
           "400_1b": time_generation(HouseAds, 500, batch=1),
           "400_2b": time_generation(HouseAds, 500, batch=2),
           "400_5b": time_generation(HouseAds, 500, batch=5),
           "400_10b": time_generation(HouseAds, 500, batch=10),
           }

print(results)
```

### Expected result:

```shell
A list of generated strings for each generation with `batch` > 1
```

### Error message:

```shell
File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-bff3490b-b252-4701-a559-e6a822d1937b/lib/python3.10/site-packages/outlines/text/generate/regex.py", line 156, in create_proposal
    assert last_state > -1
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-bff3490b-b252-4701-a559-e6a822d1937b/lib/python3.10/site-packages/outlines/text/generate/sequence.py", line 104, in step
    probs = self.create_proposal(token_ids[:, num_prompt_tokens:], probs)
  File "/local_disk0/.ephemeral_nfs/envs/pythonEnv-bff3490b-b252-4701-a559-e6a822d1937b/lib/python3.10/site-packages/outlines/text/generate/sequence.py", line 199, in __call__
    unfinished_next_token_ids, _, past_key_values = self.step(
  File "/databricks/python/lib/python3.10/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/Workspace/Repos/jan.vandervegt@databricks.com/ds-projects/app/llm/test_tunnel.py", line 106, in time_generation
    sequence = generator(inp, rng=rng)
  File "/Workspace/Repos/jan.vandervegt@databricks.com/ds-projects/app/llm/test_tunnel.py", line 116, in <module>
    "400_2b": time_generation(HouseAds, 500, batch=2),
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main (Current frame)
    return _run_code(code, main_globals, None,
```

```python
outlines==0.0.9
python==3.10.x
```

### Context for the issue:

_No response_

rlouf:
In the current design the FSM that drives the generation process and the next token sampling are coupled, which makes adding new FSMs cumbersome. We also lack the possiblity to stream the tokens. In this PR I introduce a few changes to the core of the library:

- The logic contained in the `Sequence` class now happens in a `generate_next_token` Python generator, which returns sampled tokens one at a time. This generator is responsible for calling the model, applying logit biases and sampling the next token.
- The FSM logic, currently contained in `Regex` is factored out in a new `process` generator. This generator is responsible for creating the logit biases, calling the models and concatenating the new token to the running sequence.

After these changes are made there is a clear path to #335, #163, #155, #53.

Closes #317. Closes #340. Closes #266. Closes #247. Closes #216. Closes #185.

## TODO

- [x] Move `prompts.py` to the parent module
- [x] Rename `sample.py` to `sampler(s).py`
- [x] Move the logic contained in `text` one level down so everything can be called via `outlines.generate`
- [x] Handle exceptions when the number of tokens in `token_ids` exceeds the context length;
- [x] Append the new tokens to the running sequence
- [x] Update the masks
- [x] Pass token_ids, masks and KV cache to the model via `dataclass` or `NamedTuple`
- [x] ~Update and return the current sequence, last generated token ids, last logits and total sequence logprob in a single object~
- [x] Ensure that `generator` can work with batches on inputs;
- [x] Make sure `processor` works with batches of sequences
- [x] Implement the basic FSM which stops when the eos token was generated. Implement the `outlines.generate.sequence` and `outlines.stream.sequence` functions that build the generators.
- [x] Create a regex FSM
- [x] Return `pad` tokens once an EOS token has been found
- [x] Create `outlines.generate.text`
- [x] Create `outlines.generate.regex`
- [x] Create `outlines.generate.format`
- [x] Create `outlines.generate.json`
- [x] Remove old logic
- [ ] ~Return logits + sequence logprob~
- [x] Add aliases for the old interface with deprecation waning
- [x] Test streaming in integration
- [x] Convert JSON output to dict or pydantic model
- [ ] ~Add `stop_at` kwarg to `generate.text` for feature parity~
- [x] Interface with `Index` should be `int`s and `List[int]`
- [x] `index.py` -> `fsm.py`, `/index` -> `/fsm` and `fsm.py` -> `regex.py`
- [x] Test batched guided generation
- [x] Update the paths in documentation

## Notes

- We currently assume the initial state of the FSM is labelled "0". We might want the return the initial state during init.


outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+class GenerationState:
+    token_ids: torch.Tensor
+    attention_masks: torch.Tensor
+    kv_cache: Optional[torch.Tensor] = None
+    rng: Optional[torch.Generator] = None
+
+
+class SequenceGenerator:
+    def __init__(self, fsm, model, sampler):
+        self.generate_token = token_generator(model, sampler)
```
rlouf, line 49: fsm, token_generator

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+        self.fsm = fsm
+        self.model = model
```
rlouf, line 51: Why keep this?

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+
+"""
+
+
+@dataclass(frozen=True)
+class GenerationState:
+    token_ids: torch.Tensor
+    attention_masks: torch.Tensor
+    kv_cache: Optional[torch.Tensor] = None
+    rng: Optional[torch.Generator] = None
```
rlouf, line 44: I don't think that's necessary

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+
+class SequenceGenerator:
+    def __init__(self, fsm, model, sampler):
+        self.generate_token = token_generator(model, sampler)
+        self.fsm = fsm
+        self.model = model
+        self.device = self.model.device
+
+    def init_generation_state(
+        self, prompt: Union[str, List[str]], rng: Optional[torch.Generator] = None
```
rlouf, line 55: Should be passed along with `token_generator` as this defines the internal state. Allows to use other things than LLM.

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+            The prompt on which the generation is conditioned.
+        rng
+            The state of the random number generator.
+
+        Returns
+        -------
+        A `GenerationState` object.
+
+        """
+        token_ids, attention_masks = self.model.tokenizer.encode(prompt)
```
rlouf, line 74: Also `self.model` necessary because of this function. It needs to be factored out.

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+        index
+            The index that drives the generation.
+
+        """
+
+
+def process(generator: Callable, index: "Index", state: GenerationState):
+    fsm_states = [0 for _ in range(state.token_ids.shape[0])]
+    while True:
+        logits_masks = get_next_instructions(index, fsm_states)
```
rlouf, line 112: Should return an `FSMInstruction` object.

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+
+        next_token_ids, kv_cache = generator(
+            state.token_ids,
```
rlouf, line 115: Should be able to pass `state, samples, rng`

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+            state.attention_masks,
+            state.kv_cache,
+            logits_masks,
+            1,
+            state.rng,
+        )
+
+        token_ids = update_token_ids(state.token_ids, next_token_ids)
```
rlouf, line 123: Update state function?

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+        distribution over the next tokens.
+
+    Returns
+    -------
+    A tensor with the sampled tokens.
+
+    """
+
+    def generate(
+        token_ids,
```
rlouf, line 159: `generation_state,instructions,rng`

outlines/generate/generator.py
```diff
@@ -0,0 +1,237 @@
...
+    ):
+        try:
+            logits, new_kv_cache = model(token_ids, attention_masks, kv_cache)
+        except IndexError:  # Exceeding the context length
+            return
+
+        biased_logits = bias_logits(logits, logits_masks)
+        next_token_ids = sampler(biased_logits, 1, rng)
+
+        yield next_token_ids, new_kv_cache
```
rlouf, line 173: Return

outlines/generate/generator.py
```diff
@@ -0,0 +1,315 @@
...
+if TYPE_CHECKING:
+    from outlines.generate.samplers import Sampler
+    from outlines.index.index import FSM
+
+
+@dataclasses.dataclass(frozen=True)
+class GenerationState:
+    token_ids: torch.Tensor
+    attention_masks: torch.Tensor
+    kv_cache: Optional[torch.Tensor] = None
```
rlouf, line 18: Sequence logprob?

outlines/index/index.py
```diff
@@ -0,0 +1,101 @@
...
+@dataclass(frozen=True)
+class GenerateInstruction:
+    tokens_to_mask: List[int]
+
+
+class FSM(Protocol):
+    def next_instruction(self, state: FSMState) -> GenerateInstruction:
+        ...
+
+    def next_state(self, state: FSMState, token_id: torch.Tensor) -> FSMState:
```
rlouf, line 24: This should take all the token ids of the sequence as an input. This way we can implement the `stop_at` condition: stop when a specific string has been encountered

outlines/index/index.py
```diff
@@ -0,0 +1,101 @@
...
+        ...
+
+    def is_final_state(self, state: FSMState) -> bool:
+        ...
+
+
+class StopAtTokenFSM:
```
rlouf, line 31: We can implement a map similar to the one used in `Regex`, if there is a simple way to represent "all tokens"

outlines/generate/generator.py
rlouf: Also this might be overkill but we may want a separate `Sequence` class that contains the token ids and the logprob.

rlouf: We should handle the KV cache at the model level (which becomes a `Session` object)

outlines/generate/generator.py
```diff
@@ -0,0 +1,315 @@
...
+from outlines.index.index import FSMState
+
+if TYPE_CHECKING:
+    from outlines.generate.samplers import Sampler
+    from outlines.index.index import FSM
+
+
+@dataclasses.dataclass(frozen=True)
+class GenerationState:
+    token_ids: torch.Tensor
```
rlouf, line 16: Depending on how expensive `concatenate` is it might make sense to carry lists of `Sequence` objects so we can handle beam search & SMC more easily.

outlines/generate/generator.py
```diff
@@ -0,0 +1,359 @@
...
+        except IndexError:  # Exceeding the context length
+            raise IndexError(
+                "The input length exceeds the context length of the model."
+            )
+
+        biased_logits = bias_logits(logits, logits_masks)
+        next_token_ids = sampler(biased_logits, 1, rng)
+        next_tokens = model.tokenizer.decode(next_token_ids)
+
+        return next_tokens, next_token_ids, new_kv_cache, biased_logits
```
rlouf, line 199: I think `decode` should happen at a higher level unless we have a very good reason not to.

outlines/generate/generator.py
```diff
@@ -0,0 +1,359 @@
...
+        )
+
+        def token_generator():
+            while True:
+                try:
+                    sequence = next(tokens)
+                except StopIteration:
+                    return
+
+                next_tokens: str = sequence.next_tokens
```
rlouf, line 100: We do not return the logits and kv-cache because we would like to keep compatibility with the former API.

We may actually want to transition now while we have aliases.

rlouf (commented):
T

outlines/generate/generator.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+class SequenceGenerator:
+    def __init__(self, fsm, model, sampler, device):
+        self.generate_token = token_generator(model, sampler)
+        self.fsm = fsm
+        self.tokenizer = model.tokenizer
+        self.device = device
+
+    def init_generation_state(
```
rlouf, line 28: This should be a function and initiate the FSM states as well so `sequence_generator` can be used independently by advanced users. This class would then belong to `api.py`.

outlines/generate/generator.py
```diff
@@ -0,0 +1,362 @@
...
+        attention_masks = expand_attention_masks(attention_masks)
+
+        fsm_states = get_next_fsm_states(fsm, fsm_states, next_token_ids)
+        is_finished = is_generation_finished(fsm, fsm_states)
+
+        if is_finished:
+            yield GenerationState(token_ids, kv_cache, logits)
+            return
+
+        yield GenerationState(token_ids, kv_cache, logits)
```
rlouf, line 156: The generator should only return the last token ids

rlouf: In a next iteration we can return a trie where each node corresponds to a token and links to the value of the KV values for this token at this position in the sequence, and the logprob corresponding to this token. 

When we concatenate a generated sequence with a new prompt, the next generation steps looks for nodes which have no logprob or kv values associated and computes them as a first step.

outlines/fsm/fsm.py
```diff
@@ -0,0 +1,216 @@
...
+            The id of the token that was just generated.
+
+        Returns
+        -------
+        The new state of the FSM.
+
+        """
+        self.num_tokens_generated += 1
+
+        if self.max_tokens is not None:
```
rlouf, line 196: I am not sure whether the `max_token` should live here, or closer to the generation. We current can only set the value when initialising the generator, but it may be useful to be able to specify it when *calling* the generator.

Same question regarding stopping once a series of tokens (or characters) has been found.

outlines/fsm/fsm.py
```diff
@@ -0,0 +1,216 @@
...
+        max_tokens: Optional[int] = None,
+    ):
+        regex_pattern = interegular.parse_pattern(regex_string)
+        regex_fsm, _ = make_deterministic_fsm(regex_pattern.to_fsm().reduce())
+        (
+            self.states_to_token_maps,
+            self.empty_token_ids,
+        ) = create_fsm_index_tokenizer(regex_fsm, tokenizer)
+
+        if not any(
```
rlouf, line 124: Leave a comment or refactor into a function.

outlines/generate/generator.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+@torch.inference_mode
```
brandonwillard, line 159: Should this instead be on the closure `generate` created by this constructor function?  My understanding is that `inference_mode` creates a context manager that's in effect for the duration of the call to `token_generator`, but we want one that applies to the function constructed by `token_generator` (i.e. `generate`).  Perhaps I'm forgetting something that propagates that context manager and/or its effect?

rlouf:
I am currently wondering whether `max_token` and `stop` should live closer to the generation, i.e. be enforced at the level of the `stream` method of the `SequenceGenerator` class. I find the currently implementation awkward as we have to duplicate the logic between `StopAtFSM` and `RegexFSM`. It is also cumbersome to ask users to re-init a generator when using the same model but different stopping conditions.

brandonwillard (commented):
FYI: I was seeing failures with the following traces under `torch==2.0.0`:
```python
tests/generate/test_integration_transfomers.py::test_transformers_integration_text FAILED
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
tests/generate/test_integration_transfomers.py:51: in test_transformers_integration_text
    sequence = generate.text(model)("Write a short sentence ", rng=rng)
outlines/generate/api.py:35: in __call__
    tokens = [token for token in sequence_generator]
outlines/generate/api.py:35: in <listcomp>
    tokens = [token for token in sequence_generator]
outlines/generate/api.py:65: in token_generator
    sequence = next(states)
outlines/generate/generator.py:83: in sequence_generator
    next_token_ids, kv_cache, logits = token_generator(
../../../../apps/anaconda3/envs/outlines-dev/lib/python3.10/site-packages/torch/utils/_contextlib.py:115: in decorate_context
    return func(*args, **kwargs)
outlines/generate/generator.py:137: in generate
    logits, new_kv_cache = model(token_ids, attention_masks, kv_cache)
outlines/models/transformers.py:116: in __call__
    logits, kv_cache = self.forward(input_ids, attention_mask, past_key_values)
E   TypeError: _DecoratorContextManager.__call__() takes 2 positional arguments but 4 were given
```

We appear to need to specify `torch>=2.1`, because that fixes it.

outlines/fsm/fsm.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns
+        -------
+        The new state of the FSM.
+
+        """
+        self.num_tokens_generated += 1
+
+        if self.max_tokens is not None:
+            if self.num_tokens_generated == self.max_tokens:
+                return FSMState(-1)
```
brandonwillard, line 198: Why use the type constructor here?

outlines/fsm/fsm.py
```diff
@@ -0,0 +1,212 @@
...
+        next_tokens_to_end_states = self.states_to_token_maps.get(state)
+
+        if next_tokens_to_end_states is None:
+            authorized_tokens = [self.end_token_id]
+        else:
+            authorized_tokens = list(next_tokens_to_end_states.keys())
+
+        forbidden_tokens = [
+            token for token in self.vocabulary if token not in authorized_tokens
+        ]
```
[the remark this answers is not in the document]
brandonwillard: Why are we using the negation of the valid tokens given by the pre-computed indices anyway?  Regardless, if we want to reframe generation in that way, we  need to construct FSMs/indices that match that inverse/invalid/forbidden token framing.

outlines/fsm/fsm.py
rlouf: mypy will otherwise complain afair

outlines/fsm/fsm.py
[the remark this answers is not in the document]
rlouf: We don't need to, we can set the mask from the authorized tokens.

brandonwillard (approved):
From a somewhat quick look (i.e. relative to my last review), it looks good!  We can merge whenever you want and follow-up if need be.

PR #373, opened 2023-11-16
Add support for OpenAI functions

rlouf:
We would like to integrate OpenAI functions to make it easy to compare guided generation on OSS model with the feature offered by OpenAI. Here are the design requirement:

- Usage should be very similar to guided generation, and users should be able to change a single line to switch between OpenAI and OSS models. Since OpenAI's function API does not necessarily return valid JSON I am not sure about dispatching `outlines.text.generate.json` to different functions depending on the model type. This however makes the model switch even easier;
- When inferring the output format from the function's docstring we should extract the docstring and add it as the model's description since OpenAI uses this information in their prompt;
- Output format should be the same as with the current `outlines.text.generate.json` interface: Pydantic model instance when the input format is a Pydantic model, dictionary when it is a function or a JSON schema.

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -6,6 +6,7 @@ from itertools import zip_longest
 from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set, Tuple, Union
 
 import numpy as np
+from pydantic import BaseModel
 
 import outlines
 from outlines.caching import cache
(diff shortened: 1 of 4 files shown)


A-F-V:
@rlouf I am happy to pick this one up if no one else is working on it :)

rlouf:
Would be awesome to add an interface to function calling that uses `generate.json` 😁 Go for it!

A-F-V:
I would be up for trying it but I can't setup the repo locally on my mac as per this thread: https://discordapp.com/channels/1182316225284554793/1252592587387310080

A-F-V:
c.r. #989

Issue #368, opened 2023-11-15
Reduce token consumption for "choice" generation with openAI.


HerrIvan:
### What behavior of the library made you think about the improvement?

The current solution calls the openAI api once per token. This may be costly to the point of rendering this library not valid when using openAI (at least the "is_in" completion). Specially, for languages other than English, where even short choices  require multiple tokens (acting as a cost multiplier over the prompt tokens).

Indeed, in English this may not be always a problem:

```python
prompt = """You are a sentiment-labelling assistant.
Is the following review positive or negative?

Review: This restaurant is just awesome! 
"""

answer = model(prompt, is_in=["Positive", "Negative"])
```

results in:
```
Api response: 'Positive'
Current solution: ['Positive']
Answer: Positive
```

But see the same example in Spanish:

```python
prompt = """Eres un asistente etiquetador de sentimientos.
La siguiente reseña, ¿es positiva o negativa?

Reseña: Este restaurante es increíble! 
"""

answer = model(prompt, is_in=[ "Positiva", "Negativa"])
```

Results in:

```
Api response: 'Pos'
Solution: ['Pos']

Api response: 'it'
Solution: ['Pos', 'it']

Api response: 'iva'
Solution: ['Pos', 'it', 'iva']

Answer: Positiva
```

### How would you like it to behave?

The proposed solution is heuristic. It is based in trying "optimistic" requests. Sketched algorithm:

```
We activate all tokens that may construct a 
valid answer and directly ask for max_tokens to the API. 

After that:

- In case the response exactly matches one of the complete 
  choices, return it. 

- If it does not completely match any:
  * If it contains a valid prefix for at least one choice, 
    filter the choices by the given prefix, extend the prompt 
    with the current prefix, and iterate again.
  * If there was no valid prefix, make a _greedy_ call to the 
    API just with the valid next tokens unmasked and max_tokens 
    set to 1, and iterate again.

Additionally, we can monitor at which point the amount of still 
valid choices reduces to 1, and then directly construct the answer, 
without querying the API anymore.
```

#### Pros:
- Much reduced token consumption. Final response will in general coincide with the one of the 1-token-per-call approach.

#### Cons:
- Sometimes it may involve retries, when the API responds with a string/token that is no a prefix of any valid choice. This can be resolved with a subsequent _greedy_ call.
- **(Theoretical) worst-case**: Inconsistent prefixes in string and token domains:
[...]

PR #376, opened 2023-11-17
Remove unused arguments from openai generate functions.

HerrIvan:
Removes unused arguments in `generate_base` and `generate_choice` openai model functions.

- self.client was not necessary, as the function already uses a callable pointing to a method of the self.client instance.
- max_tokens is not used in `generate_choice`.
- reduces samples to 1 when calling the openai API in `generate_choice`, as discussed in #368 . This fix is not too relevant, as the extra token-cost generated per call was just `samples-1`, but it corrects the code. Note that the alternative solution suggested is not feasible: One can call the api in parallel, but only with a same prompt. As the different samples diverge, both prompts and masks do so, hence any parallelization should be upstream from the API call.

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -30,20 +30,20 @@ class OpenAIAPI:
             )
 
         try:
-            self.client = openai.AsyncOpenAI(api_key=api_key, max_retries=max_retries)
+            client = openai.AsyncOpenAI(api_key=api_key, max_retries=max_retries)
         except openai.OpenAIError as e:
             raise e
 
         @error_handler
         @cache
         async def cached_call_completion_api(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


rlouf:
Thank you! Can we go all the way and remove it from `generate_base`? Also we don't need the client to be a property of the class.

HerrIvan:
I corrected the argument from `generate_base`. It somehow escaped me, sorry... 
I also removed `client` as a property.

HerrIvan:
- removed `client` from as an argument from generate_base, it somehow escaped me.
- removed as a property from the class. It is created but just stored in the closure.
- Warning: please check if y changes to these kind of arguments ( signature="(),(),(m),()->(s)" ) are ok.

rlouf:
Thank you!

PR #378, opened 2023-11-17
"Optimistic" is_in generation for openai.

HerrIvan:
Addresses #368.

It is quite verbose: adds two helper functions: `longest_common_prefix` and `get_choices_with_longest_common_prefix`.

Maybe you would order the code differently.

ITo make the solution 100% secure, instead of computing the "prefixes" string-based, it should be done "token-based". That would change just those two functions. Let me know what you think.

The algorithm follows the pseudo-code in the issue.

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -1,7 +1,9 @@
 """Integration with OpenAI's API."""
 import functools
 import os
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Union
+from collections import deque
+from itertools import zip_longest
+from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Set, Tuple, Union
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


outlines/models/openai.py
```diff
@@ -111,36 +147,72 @@ async def generate_choice(
...
-                    mask = {}
-                    for word, tokenized_word in zip(is_in, encoded):
-                        if not word.startswith("".join(decoded)):
-                            continue
-                        try:
-                            mask[tokenized_word[i]] = 100
-                        except IndexError:
-                            pass
+
+                greedy = False  # we try to generate the full response at each iteration
```
rlouf, line 156: We might as well remove the greedy case altogether

rlouf:
It might take me until Monday to be able to properly review the PR, thank you for contributing!

outlines/models/openai.py
HerrIvan: Say you have to tokenized choices `[T1, TX]`, `[T2, TX]`. You would unmask the three tokens and request `2` tokens from the API. If the system is very biased towards `TX` it may answer `[TX, TX]`, and  you have gained nothing. Repeating the same call you would enter a loop. Thus, it make sense then to try a _greedy_ step next just `T1` and `T2` unmasked, such that you know that you are guaranteed to move forward one token forward.

HerrIvan:
I pushed changes such that it now computes the prefix matches as sequences of tokens (list of ints) and not as strings. That makes the algorithm 100% consistent with the output generated by the LLM. The code may require some refactoring for readability/testability, but as for the algorithmic solution, I think it is complete. I won't push anything else.

outlines/models/openai.py
rlouf: Ok, this makes sense.

rlouf:
This is a very good addition, thank you! I just rebased the commits into a single commit. I will refactor the entire OpenAI integration soon, so I will leave the code as is for now and merge.

Issue #345, opened 2023-11-09
Add Outlines LlamaIndex


rlouf:
We should open a PR on the LlamaIndex repo to add Outlined as a model / guided generation provider.

PR #380, opened 2023-11-19
Proposition integration LlamaIndex

RobinPicard:
The aim of this PR is not to be merged but just to propose a potential way of integrating outlines with LlamaIndex (issue: https://github.com/outlines-dev/outlines/issues/345)

How to use LlamaIndex in outlines is not that straightforward as the way LlamaIndex is built is not fully suited for our use case. To avoid having to replicate too much of LlamaIndex's logic or interacting with its objects in a way that feels too hacky, I thought that the easiest would be to create a custom version of their `LLM` object specific to outlines that would be called when LlamaIndex would otherwise call the LLM normally chosen by the user.

The idea is that by calling the `query` method of the LlamaIndex engine as a regular user would, we let LlamaIndex figure out what context elements to add to the original prompt (including how many calls to make in case of a more sophisticated type of response synthesiser). Then, the initial outlines function the user has selected is called downstream of the LlamaIndex process through our custom `LLM` object with the modified prompt. It returns the expected LlamaIndex `CompletionResponse` object and then LlamaIndex decides whether to make some more queries or just to return the response based on the context configuration.

There are definitely many possible configuration arguments that are not covered here and I'm not sure it would already work with more complex cases but the basic example included seems to be working as intended. I'd be curious to know whether you think it looks like a promising direction to take


rlouf:
Thanks for opening a PR! My wording was confusing, what I meant was opening a PR on the LlamaIndex repo to add Outlines as a model provider 😬

RobinPicard:
Ahhh 😅 I can try but I'm not sure I'll manage to convince them to make the necessary change to accommodate outlines on top of just adding a new model provider

RobinPicard:
Ahhh 😅 I can try but I'm not sure I'll manage to convince them to make the necessary change to accommodate outlines on top of just adding a new model provider

Issue #387, opened 2023-11-22
Return the number of prompt and generated tokens for API providers


rlouf:
The number of prompt and generated tokens is important to evaluate the cost of each query. I suggest to return a class that contains the generation as well as these numbers.

HerrIvan:
I suppose the goal is keeping the counts of the tokens consumed for each instance of the OpenAI, right? Maybe keeping last and accumulated tokens, s.t. whoever is interested can inspect these values after each call.

rlouf:
I was thinking about returning it at every step, but keeping it as a class attribute is much better!

Issue #246, opened 2023-08-17
there no 'system' role in the gpt-3.5 when build prompt?


vegetax:

rlouf:
Is this a feature request?

jscheel:
to be fair, gpt-3.5 doesn't really pay attention to system messages much

PR #388, opened 2023-11-24
Track usage, add system prompt and timeout

HerrIvan:
A few improvements to the openai client.

- Added a function to wrap the call to async generate_chat, s.t. it can process the usage data and add an additional parameter (the system prompt).

- Added the system_prompt parameter: not groundbreaking, but just required a small change. It may reduce the need to do prompt composition. There was issue about this was closed as not planned. #246

- Added a timeout parameter: Without it sometimes the script was waiting for minutes and minutes for a 1-token response.

- Added the processing of usage: Accumulated usage is stored within each instance of the client. #387

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -57,7 +57,7 @@ class OpenAIConfig:
 
     """
 
-    model: str
+    model: str = ""
     frequency_penalty: float = 0
     logit_bias: Dict[int, int] = field(default_factory=dict)
     max_tokens: Optional[int] = None
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -79,6 +79,8 @@ class OpenAI:
         model_name: str,
         api_key: Optional[str] = None,
         max_retries: int = 6,
+        timeout: Optional[float] = None,
+        system_prompt: Optional[str] = None,
         config: Optional[OpenAIConfig] = None,
     ):
         """Create an `OpenAI` instance.
@@ -93,6 +95,10 @@ class OpenAI:
             `openai.api_key`.
         max_retries
             The maximum number of retries when calls to the API fail.
+        timeout
+            Duration after which the request times out.
+        system_prompt
+            The content of the system message that precedes the user's prompt.
         config
             An instance of `OpenAIConfig`. Can be useful to specify some
             parameters that cannot be set by calling this class' methods.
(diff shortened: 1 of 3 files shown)


outlines/models/openai.py
```diff
@@ -158,7 +182,7 @@ def __call__(
                 )
             )
         if "gpt-" in self.config.model:
-            return generate_chat(prompt, self.client, config)
+            return self.generate_chat(prompt, config)
```
rlouf, line 185: What I would do instead is move the response + token count extraction here instead of creating a new method.

outlines/models/openai.py
```diff
@@ -243,6 +267,26 @@ def generate_choice(
...
+        ----------
+        prompt
+            A string used to prompt the model as user message
+        config
+            An instance of `OpenAIConfig`.
+
+        """
+        results, usage = generate_chat(prompt, self.system_prompt, self.client, config)
+
+        self.last_usage = OpenAIUsage(**usage)
```
rlouf, line 285: I thought the idea of attaching the counts to the model class was good. Is there any reason why you changed your mind?

HerrIvan:
It may be that one is connected to different models (GPT-3, GPT-4, …) in a
same application. Therefore better keep the counts separate. I don’t know
if that answers the question or is it something different?

HerrIvan:
I’ve created a new method because the “cachable” function was called from
two different places, and that would’ve mean duplicating the code that
keeps track of the token consumption. (I am afc right now)
That being said, I am sure the solution can be improved :)

outlines/models/openai.py
```diff
@@ -17,6 +17,20 @@
     from openai import AsyncOpenAI
 
 
+@dataclass(frozen=True)
+class OpenAIUsage:
+    prompt_tokens: Optional[int] = 0
+    completion_tokens: Optional[int] = 0
+    total_tokens: Optional[int] = 0
```
rlouf, line 24: Since pricings are different it's probably unnecessary

HerrIvan: That's probably right. I think prices used to be different for the "completion" endpoints, not chat completion. If you want to just track total_tokens, then it simplifies quite a bit.  No class, and the call to async should just return the int.

HerrIvan: On a closer look, prompt and response tokens have indeed different costs: https://openai.com/pricing. 
Hence, if at all, I would return to my previous commit -1, where cost were kept separately.

rlouf: In which case we don't need the total number of tokens

rlouf: If only they had an API that returns the price per token for each model...

HerrIvan: Yep, no luck there. What would be your preferred solution to get this or something similar merged? 
Since we have to count different kinds of tokens, I would prefer the method, to avoid having the code for the token count update in two places.
Whether we need a class or two instance attributes, that's up to you.
At some point, I would like publish a repo where I benchmark guided vs. unguided generation in terms of cost, and would be nicer if I can add a dependency to the released version of outlines.

PR #390, opened 2023-11-25
step by step prompting and colons in prompts

MeDott29:
there appears to be, at least in ChatGPT-3.5 a meta-cognitive 'think step-by-step' approach in every response.

I couldn't get def split_into_steps to output the correct answer with anything but the prompt in its current form.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ __pycache__
 *_version.py
 docs/build
 .coverage
-1st-place-single-model-inference.ipynb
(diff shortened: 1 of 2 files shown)


rlouf:
Thank you for contributing! Could you run `pre-commit` to fix the formatting?

MeDott29:
so, I fundamentally changed the 'think step-by-step' example.  it was way harder than I imagined it would be.  it took me hours of trying prompts.  My guess is that the ":"'s are encouraging the model to infer meaning outside the scope of the problem. so, I used ": :" as an identifier and the model immediately got the answer right.

rlouf:
Well done! 👏

PR #392, opened 2023-11-27
Correct `auto-gpt` name in exception

amrrs:
it was a bit confusing to use `auto_gptq` while the library installation from pypi refers to `auto-gptq` 

https://github.com/PanQiWei/AutoGPTQ

The change should be helpful if the `ImportError` is raised


amrrs:
Feel free to ignore this PR as the naming convention of `auto_gptq` vs `auto-gptq` is not uniform across the ecosystem

PR #393, opened 2023-11-27
Open function example

amrrs:
added the autogpt example code with a sample function call

### DIFF
--- a/examples/autogptq_support.py
+++ b/examples/autogptq_support.py
@@ -1,0 +1,32 @@
+'''
+This is a simple example to show how `outlines` can leverage quantized models from Hugging Face Model Hub using `auto-gptq`
+'''
+
+## make sure to install the following libraries
+# pip install -q outlines transformers accelerate auto-gptq optimum
+
+import outlines.models as models
+import outlines.text as text
+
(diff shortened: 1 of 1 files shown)


rlouf:
Would you mind running `pre-commit` locally to fix the formatting?

rlouf:
Also this looks redundant with the [example in the documentation](https://outlines-dev.github.io/outlines/reference/json/). Is there any way we could make this more original in a way that fits in the [examples section](https://outlines-dev.github.io/outlines/examples/) of the documentation?

amrrs:
> Also this looks redundant with the [example in the documentation](https://outlines-dev.github.io/outlines/reference/json/). Is there any way we could make this more original in a way that fits in the [examples section](https://outlines-dev.github.io/outlines/examples/) of the documentation?

I was primarily intending this to be the example for GPTQ models but I get the point. I'll see if I can make a different example than one that's already there.

amrrs:
> Would you mind running `pre-commit` locally to fix the formatting?

Thanks for the feedback. 
Unfortunately, I don't have a personal computer to connect the git locally. I'll see if I can use codespaces to format or perhaps on the colab itself and update the code.

PR #398, opened 2023-12-01
reverting pydantic to v1

atmiguel:


atmiguel:
sorry this was intended for our fork

rlouf:
You will need to change a few things in the library to revert to v1, not just the dependency.

atmiguel:
Thanks for the heads up! Fortunately, we're not planning on using any of the features related to pydantic right now. We just needed to downgrade the dependency because we have other dependencies that require v1 of pydantic.

atmiguel:
We're just using it for choice-guided generation for now. We do plan on using JSON-guided generation in the future, but we'll cross that bridge when we come to it.

Issue #397, opened 2023-11-30
`anyOf` and `oneOf` implementation of the JSON Schema spec are incorrect


rlouf:
`anyOf` is the equivalent of "or" and `oneOf` is an exclusive "or".

PR #399, opened 2023-12-01
`anyOf` and `oneOf` implementation of the JSON Schema spec are incorrect

BruParis:
This PR addresses [issue #397](https://github.com/outlines-dev/outlines/issues/397).

It implements XOR regex using negative lookaheads.

The approach involves testing each `oneOf` pattern against all others using negative lookaheads. This is done in individual regex groups formatted as: "(pattern)(?!.*(other_pattern0|other_pattern1|...))". This means that 'pattern' is matched only if it is not followed by any of other_pattern0, other_pattern1, etc.

These individual groups are then combined using alternation "(|)".

This ensures exclusive matching, adhering to XOR logic.

The resulting regex has quadratic complexity. I did not find any alternatives, but it seems standard regex syntax does not permit a simpler implementation.


rlouf:
Wonderful, thank you!

PR #413, opened 2023-12-07
Define end-to-end applications

rlouf:
In this PR we allow users to define end-to-end functions from a prompt template, a Pydantic model and a model name (that can be downloaded from the Hugging Face hub):

```python
import outlines


@outlines.prompt
def test_template(text: str):
    """{{ text }}"""


class Foo(BaseModel):
    id: int


fn = outlines.Function(test_template, Foo, "gpt2")

```

These functions can then be executed, and imported by another module:

```python
from .foo import function as foo_fn

result = foo_fn("A prompt")
```

I also added a `from_github` classmethod that allows to directly download a function from a github repository:

```python
import outlines

fn = outlines.Function.from_github("outlines-dev/outlines/chain_of_density")
```

which looks for a variable named `fn` in the `chain_of_density.py` file in the `outlines-dev/outlines` repository on GitHub. It is also possible to specify an alternative name for the function:

```python
import outlines

fn = outlines.Function.from_github("outlines-dev/outlines/chain_of_density", function_name="agent")
```

## Possible improvements

- [ ] Loading models only when a `generator` is called would allow to use other model providers;
- [ ] Users might want to change the model that is being used.

### DIFF
--- a/outlines/function.py
+++ b/outlines/function.py
@@ -1,10 +1,14 @@
+import importlib.util
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Callable, Optional, Union
+from typing import TYPE_CHECKING, Callable, Optional, Tuple, Union
+
+import requests
 
 from outlines import generate, models
 
 if TYPE_CHECKING:
     from outlines.generate.api import SequenceGenerator
+    from outlines.prompts import Prompt
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


outlines/function.py
```diff
@@ -0,0 +1,121 @@
...
+            "Please provide a valid path in the form {USERNAME}/{REPO_NAME}/{PATH_TO_FILE}."
+        )
+    elif short_path[-3:] == ".py":
+        raise ValueError("Do not append the `.py` extension to the program name.")
+
+    username = path[0]
+    repo = path[1]
+    path_to_file = path[2:]
+
+    url = "/".join([GITHUB_BASE_URL, username, repo, BRANCH] + path_to_file) + ".py"
```
brandonwillard, line 84: This is a little nitpicky, but do we want to go with the `pathlib` approach throughout this library, or leave it up to the authors?

rlouf: Can you explain? I don't understand

brandonwillard (approved):

outlines/function.py
rlouf: Ah makes sense now

PR #423, opened 2023-12-09
Cache the result of OSS models

### LINKED ISSUE #371: Cache the output of open source models
We currently cache the results of requests to the OpenAI API. We should do the same thing with OSS models. The cache would only be effective if the same rng is passed at each call.

avirlrma:
fixes #371 

Todo:

- [ ] Add model parameters as part of caching key
- [ ] Unit Tests

### DIFF
--- a/outlines/text/generate/sequence.py
+++ b/outlines/text/generate/sequence.py
@@ -1,8 +1,10 @@
 import math
+from perscache.serializers import PickleSerializer
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 import torch
 
+from outlines.caching import memory as cache
 from outlines.models import OpenAI
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


avirlrma:
Hi @rlouf,
Can you please have a look at this? I realize that we have done some refactoring in the generations module, but if the approach looks good to you, I will port the change over to new code.

avirlrma:
Hi @rlouf,
Could you please have a look at this one?

outlines/text/generate/sequence.py
```diff
@@ -163,83 +165,90 @@ def __call__(
...
-
         if rng is None:
             rng = torch.Generator(device=self.device)
             rng.seed()
 
-        orig_batch_shape = token_ids.shape[:-1]
-        num_prompt_tokens = token_ids.shape[-1]
+        @cache(serializer = PickleSerializer())
+        def _cached_call(prompt: str, samples: int = 1, rng_state: [] = None) -> Union[str, List[str]]:
+            nonlocal rng
```
rlouf, line 174: `nonlocal` should be avoided unless there is no way around it

rlouf:
Hi @avirlrma it would be better to work on the new version directly. Ideally caching would happening here:   https://github.com/outlines-dev/outlines/blob/add37430b47fb914e9945c4f8fd99c875576c9f7/outlines/generate/generator.py#L129 if the logits don't take too much disk space (shouldn't). We may need to use `diskcache` instead of `perscache`, unless the `PickleSerializer` does not complain about having to cache PyTorch tensors.

Issue #438, opened 2023-12-15
Using mistrals tokenizer and a start token in the prompt, results in a failed response


venondev:
### Describe the issue as clearly as possible:

When using Mistrals tokenizer and using their prompt template, which starts with the start token "\<s\>", the generation of valid json fails. This is because part of the generated json sequence is cut off.

This is due to the tokenizer not returning the start token when first encoding the prompt to a token and then decoding the tokens back again.

```
tokens, _ = generator.tokenizer.encode(["<s>"])
print(tokens) # tensor([[1, 1]])
decoded = generator.tokenizer.decode(tokens)
print(decoded) # ['']
```

In the SequenceGenerator class, this leads to an error. First it calculates the length of the initial prompt which includes the start token "\<s\>". It then generates the sequence of tokens and decodes them back to a string, which contains the prompt and the output, but the prompt no longer includes "\<s\>". It then removes the prompt using the length of the initial prompt including the start token "\<s\>", which is longer than the decoded prompt. Because of this, part of the output sequence also gets removed. 

I will open a PR which removes the prompt on token level and not on string level, which should fix the issue.

### Steps/code to reproduce the bug:

```python
import outlines

model_name_or_path = "TheBloke/Mistral-7B-Instruct-v0.2-GPTQ"
revision = "gptq-8bit-32g-actorder_True"

model = outlines.models.transformers(model_name_or_path, model_kwargs={
    "device_map": "auto",
    "trust_remote_code": False,
    "revision": revision
}, tokenizer_kwargs={
    "use_fast": True
})

from pydantic import BaseModel
from langchain.output_parsers import PydanticOutputParser

class User(BaseModel):
    name: str
    last_name: str
    id: int

generator = outlines.generate.json(model, User, max_tokens=500)
parser = PydanticOutputParser(pydantic_object=User)

@outlines.prompt
def generate_model(format_instructions):
    """<s>[INST] Create a user. {{ format_instructions }} [/INST]"""

prompt = generate_model(parser.get_format_instructions())
result = generator(prompt)
print(result.model_dump_json(indent=2))
```

### Expected result:

```shell
A valid JSON object e.g.

{
  "name": "John Doe",
  "last_name": "Doe",
  "id": 123
}
```
```


### Error message:

```shell
Problem:


ValidationError: 1 validation error for User
__root__
[...]

PR #439, opened 2023-12-15
Fix prompt removal logic and update test case

venondev:
Solves #438


rlouf:
Could you explain the rationale for this change please? It is quite a big interface change.

rlouf:
Thank your for opening a PR! Is it problematic when streaming as well?

venondev:
Works fine in streaming mode. Probably as we do not have to remove the prompt.

rlouf:
I will need to rebase the merge commit, otherwise this looks good! Thank you for contributing

PR #447, opened 2023-12-19
Update README to point to latest function of LLM text

immortal3:
Updating Readme to point to the correct function `outlines.generate.text` instead of broken function calls.


rlouf:
Thank you for spotting that and taking the time to correct it!

Issue #433, opened 2023-12-14
Handle `max_token` more gracefully


rlouf:
A bug [was found](https://github.com/outlines-dev/outlines/pull/391#issuecomment-1852486384) in the way `max_token` is managed. Since the number of tokens generated is updated every time `FSMState` is called, when we generate tokens in batch the sequence generator returns `max_tokens / batch_size` tokens. The solution implemented in #391 consists in adding an `idx` argument to the `next_state` method of the FSMs, but this does not respect the a priori interface between the generator and FSMs. 

There are two solutions to this problem:

1. Create a FSM for each sequence in the batch.
2. Handle the constraint in `SequenceGenerator`.

Either of these allows us to remove the `idx` argument. #417 suggests that the `stop_at` constraint should be implemented at the `SequenceGenerator` level (since we need to decode the token ids), and the fact that we should let users control the maximum number of tokens generated manually makes me lean towards (2).

rlouf:
Makes sense. Let's merge #391 first.

rlouf:
An alternate solution that I implemented in the vLLM integration example was to initialize every sequence-dependent variable as a default dict and pass the sequence id. The solution of copying the FSM is much more elegant, and there's some work planned with the vLLM team to generate FSMs dynamically for each new sequence indeed.

PR #449, opened 2023-12-19
Refactor FSM idx tracking during batch generation

benlipkin:
As discussed in https://github.com/outlines-dev/outlines/issues/433, here is a proposal for removing the `idx` argument from the `FSM` interface, as well as the `reset` method, at the cost of adding a `copy` method. An FSM is initialized, then copied for each prompt during a `SequenceGenerator.__call__()` on a new batch. This solves the `max_tokens` issue, and cleans up state tracking for `CFGFSM`. The cost is overhead associated with building a new `FSM` object for each sequence in a batch. There are ways to reduce this cost by improving the new `copy` method, but wanted to get thoughts on the overall interface first.


outlines/fsm/fsm.py
```diff
@@ -103,13 +100,13 @@ def next_state(self, state: FSMState, token_id: int, idx: int = 0) -> FSMState:
...
 
-    def is_final_state(self, state: FSMState, idx: int = 0) -> bool:
+    def is_final_state(self, state: FSMState) -> bool:
         """Determine whether the current state of the FSM is a final state."""
         return state in self.final_states
 
-    def reset(self) -> None:
-        """Reset the FSM to its initial state. Here this only resets the token counter."""
-        self.num_tokens_generated = 0
+    def copy(self) -> "StopAtTokenFSM":
```
rlouf, line 107: Is the copy mechanism only necessary because we enforce `max_token` here? Same question for the other FSM. #451 is removing this mechanism from the FSMs.

rlouf:
I rebased on main and added a small commit to format docstrings, don't forget to pull locally :)

outlines/fsm/fsm.py
benlipkin: Yes, correct! If `max_tokens` is handled elsewhere, as suggested in https://github.com/outlines-dev/outlines/pull/451, then the `copy` method for both `StopAtTokenFSM` and `RegexFSM` can just be `return self`, which removes all overhead. This is a great suggestion

rlouf: Great! This comforts me in the idea that `max_tokens` should not be handled at the FSM level :)

rlouf:
I think it makes sense to wait until #451 is merged so we can simplify the `copy` method as well.

tscholak:
Does this mean we don't need to pass the seq idx as the first argument in #481?

rlouf:
We pass the `seq_idx` to the vLLM logits processor to keep track of which FSM state corresponds to which sequence.

benlipkin:
Should be up to date now post https://github.com/outlines-dev/outlines/pull/451 merge. I changed the `copy` method to `return self` as discussed for the (now) stateless FSMs.

outlines/generate/api.py
```diff
@@ -194,7 +192,9 @@ def __call__(
 
         stop_sequences = stop_at or self.stop_sequences
         max_tokens = max_tokens or self.max_tokens
+
         num_sequences = len(prompts)
+        fsms = [self.fsm.copy() for _ in prompts]
```
rlouf, line 197: Do we still need an array of FSMs now that they’re stateless?

rlouf:
Great! I’ll review shortly. Could you rebase the branch on `main`? I see a lot of commits that are not relates to this PR

benlipkin:
> Great! I’ll review shortly. Could you rebase the branch on `main`? I see a lot of commits that are not related to this PR

Sorry about that. Just rebased now, but had made a bit of a mess before when I pulled main into this branch since a lot had happened in between. If this is still too messy, I can open a new PR on a new branch and just apply the diff with main to make the history cleaner. Let me know preferences.

outlines/generate/api.py
benlipkin: Yes, because of `CFGFSM` (which needs to track incremental parser state for each). That still needs to deepcopy new objects, but the others (`RegexFSM` and `StopAtFSM`) are stateless and can just return pointers to original object.

rlouf:
> If this is still too messy, I can open a new PR on a new branch and just apply the diff with main to make the history cleaner. Let me know preferences.

Yes, I am sorry for the extra work but this would be preferable. In the future, use `rebase` instead of `merge` when you are working on a branch!

benlipkin:
Moved and updated to https://github.com/outlines-dev/outlines/pull/510

PR #458, opened 2023-12-20
Simplify logit biasing logic

Infrared1029:
Came across this library recently, was reading bits of the source code, thought this could be simplified a bit


rlouf:
Much appreciated! Looks like there's a naming problem there, I'll open an issue.

PR #469, opened 2023-12-22
mlx support

sacha-ichbiah:
Loading transformers model uses PyTorch backend. In theory, PyTorch has a mps backend support, but for a reason or another, the backend integration is bad and the models are really slow to run. I am using instead the new mlx library to run some models with outlines on my M1 Pro 16Go, and it works like a charm. 

I lack memory to convert and use Mistral7B, but it should work as well. 
I have only implemented TinyLlama-1.1B-Chat-v0.6 and microsoft/phi-2.

Some resources: 

I took the llms architectures on MLX from the following repo: 
https://github.com/ml-explore/mlx-examples/tree/main
https://github.com/ml-explore/mlx

I love outlines, I love mlx, I just want to merge together two libraries I love.

### DIFF
--- a/outlines/fsm/fsm.py
+++ b/outlines/fsm/fsm.py
@@ -4,7 +4,11 @@ import interegular
 from lark import Lark
 
 # from outlines.fsm.parsing import PartialLark
-from outlines.fsm.regex import create_fsm_index_tokenizer, make_deterministic_fsm
+import torch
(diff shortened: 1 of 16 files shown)


deipkumar:
Need any help with this?

rlouf:
Closing as this was implemented in a separate library. We could consider having this repository be part of Outlines.

PR #476, opened 2023-12-22
Add validation for model_name parameter

smwitkowski:
This PR introduces a validation check for the `model_name` parameter within the `openai` method of the `outlines.models` module. The aim is to ensure that only valid model names are used when calling this function. As per the current implementation, any string passed as the `model_name` would be accepted.

The change enforces the function to only accept `gpt-4` or `gpt-3.5-turbo` as valid model names. If any other model name is passed, a ValueError is raised, informing the user about the invalid input and suggesting the correct model names.

This change also adds tests to `tests/models/test_openai.py`. To ensure an invalid model raises a value error. 

Note: I added `gpt-4-1106-preview` in the tests, so when or if it's added to the suite of models supported by outlines, that will need to be updated, along with the mode in `openai.py`

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -104,6 +104,11 @@ class OpenAI:
             parameters that cannot be set by calling this class' methods.
 
         """
+        if model_name not in ["gpt-4", "gpt-3.5-turbo"]:
+            raise ValueError(
+                "Invalid model_name. It must be either 'gpt-4' or 'gpt-3.5-turbo'."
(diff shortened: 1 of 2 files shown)


rlouf:
Thank you for the contribution! Could you run `pre-commit` and then squash your commits? Looks good to me otherwise!

rlouf:
I fixed the code formatting and squashed the commit. Thank you for contirbuting!

Issue #484, opened 2023-12-26
Whitespace in Regex leads to different behaviours 


immortal3:
### What behavior of the library made you think about the improvement?

```python
from enum import Enum
from pydantic import BaseModel, constr

import outlines
from outlines import generate
import torch


model = outlines.models.transformers(
    "TinyLlama/TinyLlama-1.1B-Chat-v0.6",
    device="cuda",
    model_kwargs={"trust_remote_code": True},
)


class Weapon(str, Enum):
    sword = "sword"
    axe = "axe"
    mace = "mace"
    spear = "spear"
    bow = "bow"
    crossbow = "crossbow"


class Armor(str, Enum):
    leather = "leather"
    chainmail = "chainmail"
    plate = "plate"


class Character(BaseModel):
    age: int
    armor: Armor
    weapon: Weapon
    strength: int


# Construct guided sequence generator
generator = generate.json(model, Character, max_tokens=200)

# Draw a sample
rng = torch.Generator(device="cuda")
rng.manual_seed(789001)

sequence = generator("Give me a character description", rng=rng)
print(sequence)
```

Running above script results into Exception because LLM is generation `"age" \n \n \n \n .......`

```
  File "C:\Users\dippa\anaconda3\envs\ml\Lib\site-packages\pydantic\main.py", line 1079, in parse_raw
    raise pydantic_core.ValidationError.from_exception_data(cls.__name__, [error])
pydantic_core._pydantic_core.ValidationError: 1 validation error for Character
__root__
  Expecting ':' delimiter: line 98 column 2 (char 202) [type=value_error.jsondecode, input_value='{"age" \n \n  \n \n \n \...n \n \n \n \n \n \n \n ', input_type=str]
```

Now, When we update regex definition of whitespace to `whitespace = r"[\n ]?"`, It leads to better output generation.

```
age=71 armor=<Armor.plate: 'plate'> weapon=<Weapon.sword: 'sword'> strength=95
```

rlouf:
This behavior occurs very often, and it is interesting that it doesn't occur by allowing at most one white space. We need to get to the bottom of this by looking at the logits distribution (and the generated regex). It might be a problem with the way the FSM handles tokens with  white spaces but we can't know before we have explored this empirically.

rlouf:
Update (from Discord): using `[\n]*[ ]` as a regex for white spaces seems to produce "better" results. It would be nice to have a couple of comparisons here before we open a PR.

immortal3:
[...]

PR #486, opened 2023-12-27
Add LlamaCpp integration

### LINKED ISSUE #422: Add llamacpp integration

dtiarks:
fixes #422 

This is a draft PR for a LlamaCpp integration using the low-level API.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 *_version.py
 docs/build
 .coverage
+.idea/
--- a/docs/cookbook/index.md
+++ b/docs/cookbook/index.md
@@ -1,6 +1,6 @@
 # Examples
 
-- [Classification](classification): Classify customer requests.
+- [Classification](classification.md): Classify customer requests.
 - [Named Entity Extraction](extraction.md): Extract information from pizza orders.
 - [Dating Profile](dating_profiles.md): Build dating profiles from descriptions using prompt templating and JSON-guided generation.
 - [Chain Of Density](chain_of_density.md): Summarize documents using chain of density prompting and JSON-guided generation.
--- a/docs/reference/models/llamacpp.md
+++ b/docs/reference/models/llamacpp.md
@@ -1,0 +1,15 @@
+# Llama.cpp
+
+!!! Installation
+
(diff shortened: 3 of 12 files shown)


rlouf:
Great work! I still need to review the PR, but one remark for now: can we name the model `llamacpp` instead of `llama` to avoid confusion?

dtiarks:
Sure!

dtiarks:
Is there any guidance on how to handle artifacts (the GGUF model used in the test)? Just check it in using LFS?

rlouf:
So far we've been downloading models from the hub each time the tests are run. Since they're tiny models it doesn't affect the runtime much, but it would be better to cache them for sure.

dtiarks:
Ok. The llama test model is around 2meg so it should be fine. 
In order to have the llama tests running in CI/CD I added llama-cpp-python as test dependency. Question is if we should at it as  project dependency as well.

Another open question is what the current intention behind the outer dimension of the tokenizer encode method is. Is it supposed to be a batch dimension? If so, I have to implement a padding meachnism, because the llama cpp tokenizer doesn't support that out of the box.

rlouf:
We won't add it as a dependency by default but will document it in the `installation` section of the documentation.

And yes `token_ids` tensors are of shape `n_batch x n_tokens` and we keep the outer dimension throughout even for batch size of 1. Does llamacpp handle batch inference?

dtiarks:
It does have a batch-based API for inference (which is also recommended) but not for tokenization. This creates a little bit of overhead within the outlines sided implementation of the tokenizer, but this should be ok.

I will incorporate the batch API in the coming days...

rlouf:
@dtiarks is this ready for review?

dtiarks:
Not yet. I'm still trying to figure out why I get so many whitespaces in the output JSON and if it has something to do with my implementation.

Will ping you once I have news.

rlouf:
It might not be your implementation: https://github.com/outlines-dev/outlines/issues/484#issuecomment-1873733264

dtiarks:
Ok. I have a few smaller issues on my list. But those should be done today or tomorrow.

rlouf:
Thank you so much for this addition! I'm sure it will be greatly appreciated by the community :)

Issue #483, opened 2023-12-25
Cache `RegexFSM`


rlouf:
Compilation of FSMs can be quite costly, and at the very least users shouldn't need to re-compile FSMs that have already been compiled. This is why we should cache (on disk), given a regex and a vocabulary, the map from states to tokens created in the `__init__` method of `RegexFSM`.

brandonwillard:
Related to https://github.com/outlines-dev/outlines/issues/339?

rlouf:
Related, but I think the approach proposed here is different by caching the result of the entire FSM/index building process, without needing to cache the result of intermediate functions directly. Is there something I'm missing?

GastonRAraujo:
is this the reason that i got very long times to get the generator when passing maxLenght to a schema?

PR #441, opened 2023-12-17
Implement caching for OSS Models

avirlrma:
Fixes #371 

```
$ time python examples/dating_profile.py
job='Lawyer' interests=['Gaming', 'Ping pong', 'Cocktails', 'Entertaining friends', 'Wearing suits', 'Weddings', 'Latest series']

real	0m50.691s
user	0m37.346s
sys	0m14.698s

-------

$ time python examples/dating_profile.py
job='Lawyer' interests=['Gaming', 'Ping pong', 'Cocktails', 'Entertaining friends', 'Wearing suits', 'Weddings', 'Latest series']

real	0m29.366s
user	0m19.735s
sys	0m11.075s

```

PR #492, opened 2023-12-30
Cache creation of states_token_mapping in the __init__ of RegexFSM

RobinPicard:
PR aimed at talking about the resolution of issue #483 

I tried to do it in a way that requires as few modification to existing functions as possible. It implies modifying the type of several variables for them to be cacheable/json-serializeable. It looks a bit awkward and leads to the initialization of the RegexFSM instance taking a bit more time when not cached because of the type conversions, but I thought that could be the start of further discussions on the topic


rlouf:
Great! I wonder if using `diskcache` everywhere in the codebase instead of `perscache` could help make the code cleaner? See #441

RobinPicard:
I think it makes the code much cleaner in that case (and does not change the use of `@cache` in OpenAI model)

outlines/caching.py
```diff
@@ -1,20 +1,79 @@
...
 
 
-def cache(ignore: Optional[str] = None):
-    def cache_fn(fn: Callable):
-        return memory.cache(ignore=ignore)(fn)
+def hash_data(*data) -> str:
+    """Pickles and hashes all the data passed to it as args.
+    Pickling and then hashing significantly reduces the size of the key especially when dealing with large tensors.
+    """
+    result = hashlib.md5()  # nosec B303
```
lapp0, line 20: Nit: Any reason to not just use built in `hash`?

outlines/fsm/fsm.py
```diff
@@ -114,27 +115,48 @@ def __init__(
...
+        ) -> Tuple[str, list]:
+            """Return the values that will be used to create the cache key of create_states_mapping"""
+            cacheable_vocabulary = sorted(tokenizer.vocabulary.values())
+            return (regex_string, cacheable_vocabulary)
+
+        @cache(func_cache_key_args)
+        def create_states_mapping(
+            regex_string: str, tokenizer: "Tokenizer"
+        ) -> Tuple[dict, set, set]:
+            """Create the variables related the mapping between stzates and tokens
```
lapp0, line 129: Nit: typo

outlines/fsm/fsm.py
```diff
@@ -114,27 +115,48 @@ def __init__(
...
+            if not any(
+                regex_fsm.finals.intersection(v.values())
+                for v in states_to_token_maps.values()
+            ):
+                raise ValueError(
+                    "The vocabulary does not allow us to build a sequence that matches the input regex"
+                )
+
+            final_states = regex_fsm.finals | {
+                -1
```
lapp0, line 151: What does `-1` indicate in the context of a `RegexFSM` state?

outlines/caching.py
```diff
@@ -1,20 +1,79 @@
...
+    for datum in data:
+        if isinstance(datum, torch.Tensor):
+            datum = datum.cpu().numpy()
+        result.update(cloudpickle.dumps(datum))
```
lapp0, line 24: IMHO, since you're already defining a key function for each hashable, the key function should be responsible for ensuring hashability, rather than ensuring arbitrary objects are dumpable via `cloudpickle`, which may have non-deterministic outputs: https://github.com/cloudpipe/cloudpickle/issues/385

outlines/fsm/fsm.py
rlouf: Everything that is a final state not defined by the regex. This was used to stop generation when `max_token` was found but I'm not sure that's relevant anymore.

outlines/caching.py
RobinPicard: The built-in hash function does not return consistent values across sessions: https://docs.python.org/3.4/reference/datamodel.html#object.__hash__

outlines/caching.py
RobinPicard: I removed the line regarding `torch.Tensor` that I had picked up from #441 as it's not needed for the issue this PR tries to solve. On hashability, I wanted to give the option not to provide a key function if we just want to use the arguments of the cached function for the key (the standard situation I guess) so that we don't have to encumber the file of the cached function with an extra function definition.

lapp0:
Good work!

outlines/caching.py
```diff
@@ -1,20 +1,74 @@
...
+    """Caching decorator for memoizing function calls.
+    The cache key is created based on the values returned by the key_function callable
+    if provided or based on the arguments of the decorated function directly otherwise
+    Parameters
+    ----------
+    key_function
+      A callable function used to generate a unique key for each function call. It's
+      called with the arguments of the decorated function as arguments
+    Returns
+    -------
```
rlouf, line 33:
```suggestion
    Parameters
    ----------
    
    key_function
      A callable function used to generate a unique key for each function call. It's
      called with the arguments of the decorated function as arguments
      
    Returns
    -------
```

outlines/fsm/fsm.py
```diff
@@ -114,27 +115,48 @@ def __init__(
...
         tokenizer: "Tokenizer",
     ):
-        regex_pattern = interegular.parse_pattern(regex_string)
-        regex_fsm, _ = make_deterministic_fsm(regex_pattern.to_fsm().reduce())
+        def func_cache_key_args(
+            regex_string: str, tokenizer: "Tokenizer"
+        ) -> Tuple[str, tuple]:
+            """Return the values that will be used to create the cache key of create_states_mapping"""
+            cacheable_vocabulary = tuple(sorted(tokenizer.vocabulary.items()))
+            return (regex_string, cacheable_vocabulary)
```
rlouf, line 123: Why not call the decorator directly with the parameters that are used for the cache key? It would be more explicit.

outlines/caching.py
rlouf: Also, does will this work with the OpenAI integration?

outlines/fsm/fsm.py
```diff
@@ -114,27 +115,48 @@ def __init__(
...
+
+        @cache(func_cache_key_args)
```
rlouf, line 125: Could we call the decorator with `regex_string` and `tokenizer.vocabulary.items()` directly instead of using a function to extract the cache keys? Default behavior could be to use the decorated functions' arguments.

outlines/caching.py
RobinPicard: Yes. I've added `async_wrapper` for it to work with the OpenAI integration

outlines/fsm/fsm.py
RobinPicard: Ah yes that would be more straightforward in that case. I've just changed it

rlouf:
Thank you!

Issue #419, opened 2023-12-08
Handle `Optional` fields in Pydantic models


rlouf:
Outlines does not take the `required` field into account in JSON Schemas.

RobinPicard:
On the more general topic of the translation of a JSON Schema into a regex, do you think it would make sense at some point to create an intermediary representation of the JSON Schema with a custom Python object to facilitate the creation of the regex string afterward?

rlouf:
What would be the benefits?

RobinPicard:
I think the benefit could be a clearer separation between the characters for the actual content of the keys/values, those for the logical operations (caused by required, anyOf...) and those related to the json syntax (the quotation marks, commas...). I imagined it could make it easier to add new features

brandonwillard:
> I think the benefit could be a clearer separation between the characters for the actual content of the keys/values, those for the logical operations (caused by required, anyOf...) and those related to the json syntax (the quotation marks, commas...). I imagined it could make it easier to add new features

This sounds like a broader, speculative design-level implementation consideration.  For this issue, we first need to determine how we might want to implement `Optional` using the current regex approach.  After that, we'll have a better idea of which implementation designs/refactoring/IR will best facilitate this and the existing features.

PR #495, opened 2024-01-01
Support required keyword in JSON Schema

RobinPicard:
The aim of this PR is to solve issue #419

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -103,9 +103,10 @@ def to_regex(resolver: Resolver, instance: dict):
         properties = instance["properties"]
         required_properties = instance.get("required", [])
         is_required = [item in required_properties for item in properties]
-        # If at least one property is required, we include it without any comma.
+        # If at least one property is required, we include the one in the lastest position
+        # without any comma.
         # For each property before it (optional or required), we add with a comma after the property.
-        # For each property after it (optional or required), we add with a comma before the property.
+        # For each property after it (optional), we add with a comma before the property.
         if any(is_required):
             last_required_pos = max([i for i, value in enumerate(is_required) if value])
             for i, (name, value) in enumerate(properties.items()):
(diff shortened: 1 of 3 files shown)


rlouf:
Great! Is this ready for review?

RobinPicard:
Yes. I'm worried I've forgotten some cases but review should help spotting the things I may have missed

rlouf:
I'm on vacation, will take a look as soon as I can!

outlines/fsm/json_schema.py
```diff
@@ -102,13 +100,43 @@ def to_regex(resolver: Resolver, instance: dict):
...
-            regex += f'{whitespace}"{name}"{whitespace}:{whitespace}'
-            regex += to_regex(resolver, value)
-
-            # No comma after the last key-value pair in JSON
-            if i < len(instance["properties"]) - 1:
-                regex += f"{whitespace},"
+        properties = instance["properties"]
+        required_properties = instance.get("required", [])
+        is_required = [item in required_properties for item in properties]
+        # If at least one property is required, we include the one in the lastest position
```
rlouf, line 106: Is this changing the order of the fields as they were specified in the JSON schema?

rlouf:
```python
We need to make sure that the last item generated does not end with a comma and thus end up with invalid JSON.

When at least one property is required, we select the last in the list and include it at the last position without a comma. We then prepend every prior (optional or required) property with a comma.
```

rlouf: Why not do the same as below, actually?

RobinPicard: This is not changing the order of the fields.
An issue with `When at least one property is required, we select the last in the list and include it at the last position without a comma` is that it would change the order of the fields if the last property is an optional one.

RobinPicard: I can't see a way of easily combining both as the case in which there are no required properties creates as many subregexes as there are properties and connect them with `|` while the case with at least one required property does not need that.

rlouf:
We need to fix that merge conflict before merging

RobinPicard:
> We need to fix that merge conflict before merging

I can't see the conflict. I'm told my branch is up to date when trying to rebase in git and I have "Able to merge" on Github on my forked repository

rlouf:
Thank you!

PR #496, opened 2024-01-01
wip: support empty and recursive schemas

tscholak:
The main contribution of this PR is an enhanced `to_regex` function, which addresses both empty and recursive JSON schemas. I utilize the capabilities of the `regex` library, which notably supports recursion. Here's how this is achieved:

#### Handling Empty JSON Schemas

We identify empty schemas (i.e., `{}` or `True`) as those that should accept any valid JSON. To achieve this, a specific regex pattern for JSON values is implemented. This pattern is designed to be universally accepting.

#### Addressing Recursive JSON Schemas

The core of handling recursion lies in the use of named capture groups within the regex patterns. These groups are critical for creating references within the regex that can call back to themselves or to other groups.

The `regex` library's support for recursion is fundamental to this approach. It allows for patterns that can reference themselves, enabling the definition and validation of recursive and self-referential schemas. This is particularly important for complex schemas where elements may be nested within themselves or have mutual references.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,5 +28,5 @@ repos:
   rev: v1.2.0
   hooks:
     - id: mypy
-      args: [--allow-redefinition]
+      args: [--allow-redefinition, --install-types]
       exclude: ^examples/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,5 +28,5 @@ repos:
   rev: v1.2.0
   hooks:
     - id: mypy
-      args: [--allow-redefinition, --install-types]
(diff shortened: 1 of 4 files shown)


brandonwillard (commented):
Could you provide a high-level explanation of the approach and how it will address empty and recursion JSON schemas?

tscholak:
@brandonwillard I added a description to the PR.

I realize now that my approach is not going to work, because `interegular` does not support calls to named groups (like `(?&name)`).

brandonwillard:
I'll close this for now, but feel free to reopen if you have questions specific to these changes that you would like us to answer.

PR #500, opened 2024-01-03
Check OpenAI model name using their client

### LINKED ISSUE #499: Remove the check for invalid model names in the OpenAI API integration
This means we have to manually add models every time they come out. We might as well let the API call error if the model specified by the user does not exist.

remic33:
Check call usable openai models using openai API retrieval
close #499

### DIFF
--- a/outlines/models/openai.py
+++ b/outlines/models/openai.py
@@ -104,10 +104,6 @@ class OpenAI:
             parameters that cannot be set by calling this class' methods.
 
         """
-        if model_name not in ["gpt-4", "gpt-3.5-turbo"]:
-            raise ValueError(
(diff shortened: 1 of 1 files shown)


rlouf:
What API endpoint exactly?

remic33:
> What API endpoint exactly?

GET https://api.openai.com/v1/models/{model}

remic33:
@rlouf tests are broken because of Openai installation not recquired by config. Should I had it by default?

rlouf:
Removing the [now deprecated test](https://github.com/outlines-dev/outlines/blob/0355ab4272a5d7e4d94c4a53a52593f885b81a61/tests/models/test_openai.py#L54) should be enough.

PR #517, opened 2024-01-10
Add CFG to vllm serving

mory91:
Hi,
This pull request adds support for CFG in vllm serving.

### DIFF
--- a/outlines/serve/serve.py
+++ b/outlines/serve/serve.py
@@ -25,6 +25,7 @@ from vllm.sampling_params import SamplingParams
 from vllm.utils import random_uuid
 
 from .vllm import (
+    CFGLogitsProcessor,
     JSONLogitsProcessor,
     RegexLogitsProcessor,
     _patched_apply_logits_processors,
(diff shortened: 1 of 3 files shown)


outlines/serve/serve.py
```diff
@@ -122,6 +127,8 @@ async def stream_results() -> AsyncGenerator[bytes, None]:
     # Sets default for the model (`facebook/opt-125m`)
     engine = AsyncLLMEngine.from_engine_args(engine_args)
 
+    _adapt_tokenizer(engine.engine.tokenizer)
```
rlouf, line 130: Why are you calling this function here? The result is not used.

mory91: The tokenizer is changed inside the function anyways. I now assigned it to the tokenizer though.

rlouf: Ah makes sense. It's not needed however as vLLM handles tokenisation on its end during encoding/decoding.

rlouf:
Thank you for your contributions! I added some documentation before merging.

outlines/serve/serve.py
mory91: It's needed because in here https://github.com/outlines-dev/outlines/blob/fde61a80a58de0401fdecdee7408db53e17ca4f4/outlines/fsm/fsm.py#L345 outlines expects tokenizer to return a list but vllm tokenizers return string. 
Also I just realized that this change makes a breaking change to the library. If it's fine by the project directors its fine, if not we might need to revert and re-think about this.

PR #481, opened 2023-12-24
update vllm patch to v0.2.6 and add regex support for `/generate`

tscholak:
closes #478 

- [x] update `outlines.serve` to vllm-v0.2.6 and ray-2.9.0
- [x] simplify monkey patching
- [x] don't use `nest_asyncio` with `uvloop.Loop` (see https://github.com/erdewit/nest_asyncio/issues/49)
- [x] add regex support for `/generate`
- [x] simplify/deduplicae vllm integration example by using `outlines.serve.vllm` code
- [x] update documentation
- [ ] integration tests

Here's a `Dockerfile` that I used for testing:

```Dockerfile
# syntax=docker/dockerfile:1.2

# Use the vllm/vllm-openai image with the specified tag
ARG BASE_TAG=latest
FROM vllm/vllm-openai:${BASE_TAG}

# Install git and openssh-client, then clean up to keep the image small
RUN apt-get update && \
    apt-get install -y git openssh-client python3.10-venv && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

# Clone the repository, checkout the specific commit, build the source distribution, install it, and clean up
ARG COMMIT
RUN --mount=type=secret,id=ssh_key \
    mkdir -p -m 0700 ~/.ssh && \
    ssh-keyscan github.com >> ~/.ssh/known_hosts && \
    cp /run/secrets/ssh_key ~/.ssh/id_rsa && \
    chmod 600 ~/.ssh/id_rsa && \
    git clone git@github.com:tscholak/outlines.git && \
    cd outlines && \
    git checkout ${COMMIT} && \
    python3 -m pip install build && \
    python3 -m build && \
    pip install dist/outlines-*.whl && \
    cd .. && \
    rm -rf outlines ~/.ssh /root/.cache/pip

# Install ray with pydantic-2.5 support and clean up
RUN pip install ray==2.9.0 && \
    rm -rf /root/.cache/pip
```

and the accompanying `Makefile`:

```Makefile
TAG=v0.2.6
COMMIT=c4e0b5ae2e97ec0377c726aa65a2207c939377de

# Build the Docker image with a configurable tag
build:
	DOCKER_BUILDKIT=1 \
	docker build \
	--secret id=ssh_key,src=$(SSH_KEY) \
	--build-arg BASE_TAG=$(TAG) \
	--build-arg COMMIT=$(COMMIT) \
	-t $(FULL_IMAGE_NAME) \
	.
```

PR #520, opened 2024-01-10
Patch nest_asyncio in notebooks only

dzhulgakov:
The logic to call `nest_asyncio` was added in https://github.com/outlines-dev/outlines/commit/587f9ac139651ff2e4af860fbd6e155f38eca675 presumably to make notebook interactions easier.

However, it's a bad practice to unconditionally call global python patches from library code.

It caused some issues with vLLM integration and was patched in https://github.com/outlines-dev/outlines/pull/481

We ran into different issues when using Outlines in our codebase. The unittests started hanging on shutdown because of stray asyncio threads.

I propose to make the patch happen only when running inside IPython. According to https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook it's an ok detention method for notebooks. Other proposed methods are too narrow and e.g. don't capture Colab.

However, even better fix might be to remove this patch all together and instead include it explicitly in the notebooks of choice. @rlouf - do you have more context on what was the original issue?

### DIFF
--- a/outlines/base.py
+++ b/outlines/base.py
@@ -1,9 +1,9 @@
 import asyncio
+import builtins
 import functools
 import inspect
 from typing import Callable, Optional
 
-import nest_asyncio
 import numpy as np
 from numpy.lib.function_base import (
     _calculate_shapes,
@@ -12,11 +12,17 @@ from numpy.lib.function_base import (
     _update_dim_sizes,
 )
 
-# Allow nested loops, useful to run in notebooks
-try:
-    nest_asyncio.apply()
-except ValueError as e:
-    print("Could not apply nest_asyncio:", e)
+# Allow nested loops for running in notebook. We don't enable it globally as it
+# may interfere with other libraries that use asyncio.
+if hasattr(builtins, "__IPYTHON__"):
+    try:
+        import nest_asyncio
+
+        nest_asyncio.apply()
+    except ImportError:
+        print(
+            "Couldn't patch nest_asyncio because it's not installed. Running in the notebook might be have issues"
+        )
 
 
 class vectorize:


rlouf:
Thank you for opening a PR! It was indeed to make interaction with notebooks easier.

outlines/base.py
```diff
@@ -12,11 +12,19 @@
...
+    try:
+        import nest_asyncio
+
+        nest_asyncio.apply()
+    except ImportError as e:
+        print(
+            "Couldn't patch nest_asyncio because it's not installed. Running in the notebook might be have issues"
+        )
+    except ValueError as e:
+        print("Could not apply nest_asyncio:", e)
```
rlouf, line 27: When is `ValueError` raised?

dzhulgakov: It was there from https://github.com/outlines-dev/outlines/pull/481 . Presumably some vllm integration but I didn't run into it myself

dzhulgakov:
> What’s your use case?

We're trying to use Outlines in some proprietary codebase (thank you for building it!). In one of the tests just importing `outlines` caused the test to hang at shutdown. The symptom is that `threading.py` `_shutdown` tries to join some threads. And there's a bunch of asyncio threads stuck in the process. I didn't debug it further, but the test uses asyncio and multiprocessing so it's kind of involved. Commenting out nest_asyncio or applying this PR solves the issue.

Btw, I verified that `builtins.__IPYTHON__` trick works correctly when importing in colab.

outlines/base.py
rlouf: We can remove it

rlouf:
Thank you for contributing!

PR #521, opened 2024-01-10
Updates chess cookbook example with Phi-2 and adds omitted code. 

kgourgou:
Hello, 

This was a fun example in the cookbook, but there were some elements missing, e.g., `model` was referenced in the code, but not defined anywhere, `board_state` was set to be the model, etc. 

I updated the example to use the smaller Phi-2 (which actually plays terrible, but otherwise longer chess) and added the code needed so [that the example can run.](https://colab.research.google.com/drive/1-f9CC64HGgaZhxZTvJkZPfenySkb0Yt7?usp=sharing)


kgourgou:
Done.

rlouf:
Thank you for updating the example, much appreciated!

PR #522, opened 2024-01-10
Support Azure OpenAI

### LINKED ISSUE #271: Add support for Azure OpenAI API and Example in the documentation
Would be very nice to have the functionality to support Azure OpenAI model API. 

And to have more examples showing parsing using JSON schemas and pedantic schemas on OpenAI models

joeddav:
Adds a wrapper around the `OpenAI` class, creating an `AsyncAzureOpenAI` client instead of an `AsyncOpenAI` client.

Also required one change to `openai.py`: I put the tiktoken logic inside a computed `tokenizer()` property so that I could override this in the subclass. This is necessary because Azure OpenAI has a distinction between a `deployment` and a `model`, the former of which is used when making calls to the API, but the latter is what must be passed to tiktoken.

Usage:

```python
model = outlines.models.azure_openai(
    model_name="gpt-4",
    deployment_name="gpt4-1106",   # my custom deployment name
    azure_endpoint="https://<myendpoint>.openai.azure.com/",
    api_key="<mykey>"
)
model.generate_choice(
    "What is the capital of the country where DaVinci is from?",
    choices=["Milan", "Rome", "Paris", "Italy"]
)
# 'Rome'
```

vs.

```python
model = outlines.models.openai("gpt-4")
model.generate_choice(
    "What is the capital of the country where DaVinci is from?",
    choices=["Milan", "Rome", "Paris", "Italy"]
)
# 'Rome'
```

I coded this up for my own local use and but thought I'd share here in case you'd like to incorporate it, or in case someone else wanted to use it.

Resolves #271

### DIFF
--- a/outlines/models/azure.py
+++ b/outlines/models/azure.py
@@ -1,0 +1,119 @@
+"""Integration with Azure OpenAI's API."""
+import functools
+import os
+from dataclasses import replace
+from typing import Optional
+
+from outlines.models.openai import OpenAI, OpenAIConfig
+
+__all__ = ["AzureOpenAI", "azure_openai"]
+
+
+AZURE_API_VERSION = "2023-05-15"
+
+
+class AzureOpenAI(OpenAI):
+    def __init__(
+        self,
+        model_name: str,
+        deployment_name: str,
+        azure_endpoint: Optional[str] = None,
+        api_key: Optional[str] = None,
+        max_retries: int = 6,
+        timeout: Optional[float] = None,
+        system_prompt: Optional[str] = None,
+        config: Optional[OpenAIConfig] = None,
+    ):
+        """Create an `AzureOpenAI` instance.
+
+        Parameters
+        ----------
+        model_name
+            The name of the OpenAI model being used
+        deployment_name
+            The name of your Azure OpenAI deployment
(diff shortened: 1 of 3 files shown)


rlouf:
This is a great addition, thank you so much for your contribution!

outlines/models/azure.py
```diff
@@ -0,0 +1,119 @@
...
+
+        """
+        try:
+            import openai
+        except ImportError:
+            raise ImportError(
+                "The `openai` library needs to be installed in order to use Outlines' Azure OpenAI integration."
+            )
+        try:
+            client = openai.OpenAI()
```
pamelafox, line 57: Thanks for adding Azure support! When I try to use azure_openai, it fails at this line of code, since the OpenAI client attempts to authenticate. Is there a different way we can validate the model name, besides instantiating the client?

rlouf: I am now leaning towards "let it crash" i.e. let users enter anything for the model name and then handle a potential error gracefully, letting the user know that the model name is incorrect.

rlouf: I don’t have the time to fix it right now, but happy to review a PR.

PR #531, opened 2024-01-11
Proposition implementation token alignment

RobinPicard:
Opening this PR to discuss the implementation of token alignment (#161)

This is not intended to be merged, I was just wondering whether you think this is a promising direction to look into

The idea is to modify the `self.states_to_token_maps` of the fsm associated to each prompt to create a new state to accommodate the fact that a token has been removed compared to when this map was created during the initialization of the fsm


rlouf:
> This is not intended to be merged, I was just wondering whether you think this is a promising direction to look into

I think this is the right general direction.


> - The case in which the text after the "boundary" of a token matching the end of the prompt does not exist in the vocabulary by itself is not covered

Could you illustrate this?

RobinPicard:
> Could you illustrate this? I had a PR opened (can't find it right now) where I iterated once over the vocabulary to find the overlapping tokens.

Making up a fake example. My prompt is "Good mor". Let's say there's a token for "mor" and it's the last one of the prompt. We would want token alignment to replace "mor" with "morning". However, if the token "ning" by itself does not exist, then there's nothing in the `states_to_token_maps` that correspond to it as, at this point, the character-based FSM that would allow to generate "ning" has already been turned into a token-based mapping.

I was looking at creating `states_to_token_maps` only after the call is made (and the FSM is updated) but that would add too much overhead.

I was then thinking that a solution could be to create at initialization a mapping that contains both information about characters and about tokens (so we would have some states with no tokens leading to them that would be used for the token alignement)

rlouf:
How about looping over the entire vocabulary and store the tokens that accept `mor` as a prefix. Then, in the unconstrained case the first state of the FSM would have transitions to the overlapping tokens only?

Haven't taken the time to think about the constrained case yet.

RobinPicard:
What I thought was the issue with this approach concerns at what state we end up in the FSM after the part "the first state of the FSM would have transitions to the overlapping tokens only". I had initially imagined that we take the postfix, find the associated token and the state to which it would transition from the original initial state of the FSM. But then the transitions in the original FSM from the initial state can only include postfixes that exist as standalone tokens in the vocabulary.

I have a pretty hard time explaining what I have in mind (I may just be confused actually), sorry. I'm going to try to find a real example

RobinPicard:
Here's the issue I was blocked by @rlouf:

```
# our prompt's last token is "ra"
# we want to be able to replace with the token "rans" that is part of the vocabulary

import outlines
import interegular
from outlines.fsm.regex import create_fsm_index_tokenizer, make_deterministic_fsm

regex_string = r'[n-s]{2}[0-1]'

model = outlines.models.transformers("hf-internal-testing/tiny-random-gpt2")

regex_pattern = interegular.parse_pattern(regex_string)
regex_fsm, _ = make_deterministic_fsm(regex_pattern.to_fsm().reduce())
print(regex_fsm)

#  name final? 0 1 n o p q r s anything_else
#--------------------------------------------
#* 0    False      1 1 1 1 1 1
#  1    False      2 2 2 2 2 2
#  2    False  3 3
#  3    True

states_to_token_maps, empty_token_ids = create_fsm_index_tokenizer(
    regex_fsm, model.tokenizer
)
characters_map = {}
for key, value in states_to_token_maps.items():
    characters_map[key] = {}
    for a, b in value.items():
        characters_map[key][model.tokenizer.decode([a])[0]] = b
print(characters_map)

#{
#    0: {'so': 2, 'o': 1, 'p': 1, 'q': 1, 'pp': 2, 'op': 2, 'pr': 2, 'ro': 2, 'r': 1, 'or': 2, 'os': 2, 's': 1, 'n': 1, 'on': 2},
#    1: {'r': 2, 's': 2, 'n': 2, 'o': 2, 'p': 2, 'q': 2},
#    2: {'0': 3, '1': 3}
#}

# problem: there's no token at state 0 whose characters are equal to "ns"
# if we were to replace the "ra" at the end of the prompt by "rans", we would not know
# where we are in the fsm
```

Now that I'm writing it down like that, I realize that maybe we're always sure that there are tokens for each individual characters that compose the postfix in our states_to_token_maps (like here) such we could always walk the states_to_token_maps character by character until we end up at the end of the postfix (so we would know what state we end up at)

RobinPicard:
I had not realized that I could walk the `states_to_token_maps` character by character for the postfix part of the crossing tokens in the constrained case. I think it works with almost no additional overhead like that. Let me know if you think it's fine and I'll update the tests afterward

outlines/fsm/fsm.py
```diff
@@ -37,8 +41,23 @@ class StopAtTokenFSM(FSM):
 
     def __init__(self, tokenizer: "Tokenizer", stop_token_id: int):
         self.stop_token_id = stop_token_id
-        self.vocabulary = tokenizer.vocabulary.values()
-        self.final_states = {1}
+        self.tokenizer = tokenizer
+        self.vocabulary = tokenizer.vocabulary
+        self.final_states = {2}
+        self.valid_alignment_tokens: List[int] = []
```
rlouf, line 47: `valid_boundary_tokens`?

outlines/fsm/fsm.py
```diff
@@ -37,8 +41,23 @@ class StopAtTokenFSM(FSM):
...
+
+    def align_prompt_tokens(self, prompt: str) -> str:
```
rlouf, line 49: `remove_boundary_token`

Can we be sure that we cannot go back more than one token? Say the prompt contains ["mo", "rn"] and the token "morning" is present in the vocabulary.

outlines/fsm/fsm.py
```diff
@@ -37,8 +41,23 @@ class StopAtTokenFSM(FSM):
...
+        """Remove the last token from the prompt and set the value of self.valid_alignment_tokens"""
+        token_ids, _ = self.tokenizer.encode(prompt)
+        last_token_id = int(token_ids[0][-1])
+        last_token_text = self.tokenizer.decode([last_token_id])[0]
```
rlouf, line 53: I am not a huge fan of encoding/decoding here. I think we should return `token_ids` and use those to initialize the generation later on.

outlines/fsm/fsm.py
```diff
@@ -119,9 +140,63 @@ def __init__(self, regex_string: str, tokenizer: "Tokenizer"):
...
+        last_token_length = len(last_token_text)
+        # select the tokens that start with the text removed from the prompt
+        crossing_tokens = {
+            token: text
+            for text, token in self.vocabulary.items()
+            if text.startswith(last_token_text)
+        }
```
rlouf, line 158: The section above would be re-used for the 3 FSMs. This could be a sign that the code is not located where it should be.

outlines/fsm/fsm.py
```diff
@@ -119,9 +140,63 @@ def __init__(self, regex_string: str, tokenizer: "Tokenizer"):
...
+            is_valid = True
+            crossing_text = text[last_token_length:]
+            state = 0
+            for char in crossing_text:
+                char_token = self.vocabulary.get(char)
+                try:
+                    state = self.states_to_token_maps[state][char_token]  # type: ignore
+                except KeyError:
+                    is_valid = False
+                    break
```
rlouf, line 195: Yes, this works because of the way tokenizers are trained, if a given token is present at any given transition then the individual characters that make up this token are also necessarily present.

We should add a word of warning here.

rlouf:
Yes I think that's the right approach. There's some stuff to figure out in terms of design, but otherwise looks good.

outlines/fsm/fsm.py
RobinPicard: Ah yes, right. I changed it such that it could go back as many tokens as needed

RobinPicard:
I'll write unit tests next if you think having those separate functions is the right design

outlines/fsm/fsm.py
```diff
@@ -333,3 +387,162 @@ def next_state(self, state: FSMState, token_id: int) -> FSMState:
...
+    """
+    reversed_vocabulary = {value: key for key, value in vocabulary.items()}
+    len_token_ids = len(token_ids)
+    max_length_token_text = max(len(item) for item in vocabulary.keys())
+    characters_considered = ""
+    crossing_tokens_map = {}
+
+    for index, token_id in enumerate(reversed(token_ids)):
+        characters_considered = reversed_vocabulary[token_id] + characters_considered
+        if len(characters_considered) >= max_length_token_text:
```
rlouf, line 436: Can we break on `crossing_token_ids` being empty?

outlines/generate/api.py
```diff
@@ -176,10 +189,50 @@ def __call__(
         num_samples = self.num_samples
         batch_size = len(prompts)
 
-        prompt_token_ids = torch.repeat_interleave(prompt_token_ids, num_samples, dim=0)
-        attention_masks = torch.repeat_interleave(attention_masks, num_samples, dim=0)
         fsm_states = [FSMState(0) for _ in range(batch_size * num_samples)]
         fsms = [self.fsm.copy() for _ in range(batch_size * num_samples)]
+
+        prompt_token_ids = torch.repeat_interleave(prompt_token_ids, num_samples, dim=0)
```
rlouf, line 195: Why did you move these?

outlines/generate/api.py
```diff
@@ -176,10 +189,50 @@ def __call__(
...
+        attention_masks = torch.repeat_interleave(attention_masks, num_samples, dim=0)
+
+        # Token alignment may shorten some of the prompts by removing tokens at their end.
```
rlouf, line 198: Create a method for this?

outlines/fsm/fsm.py
```diff
@@ -121,9 +156,22 @@ def create_states_mapping(
         self.states_to_token_maps, self.empty_token_ids = create_states_mapping(
             regex_string, tuple(sorted(tokenizer.vocabulary.items()))
         )
-        self.vocabulary = tokenizer.vocabulary.values()
+        self.vocabulary = tokenizer.vocabulary
         self.eos_token_id = tokenizer.eos_token_id
 
+    def align_prompt_tokens(
```
rlouf, line 162: Is there any way we could avoid modifying the `outlines.fsm.fsm` file and implement all the logic in `outlines.generate.api`? These feel like separate concerns.

rlouf:
I have small design comments, but nothing that would dramatically affect your solution. You can start implementing tests.

rlouf (changes requested):
I have made several comments on the overall design, but nothing that would dramatically affect your implementation. You can start implementing tests.

outlines/fsm/fsm.py
RobinPicard: I'm not sure I get it. If we were to go through all tokens in `token_ids` we would end up doing more loops over the vocabulary because of that line I think
```
crossing_token_ids = [
            token_id
            for text, token_id in vocabulary.items()
            if text.startswith(characters_considered)
            and len(text) > len(characters_considered)
        ]
```

outlines/generate/api.py
RobinPicard: Mindlessness, I'll put it back

outlines/fsm/fsm.py
RobinPicard: My issue was that this token alignment does not apply to CFGFSM as it does not have a `states_to_token_maps`

outlines/generate/api.py
```diff
@@ -176,10 +189,50 @@ def __call__(
...
+            ]
+        )
+        max_length_aligned_prompt = max(prompt.shape[0] for prompt in aligned_prompts)
+        padded_aligned_prompts = [
+            torch.cat(
+                [
+                    torch.full(
+                        (max_length_aligned_prompt - prompt.shape[0],),
+                        0,
+                        dtype=prompt.dtype,
```
shawnz, line 213: should also set device=self.device

outlines/generate/api.py
```diff
@@ -176,10 +189,50 @@ def __call__(
...
+            )
+            for prompt in aligned_prompts
+        ]
+        padded_aligned_masks = [
+            torch.cat(
+                [
+                    torch.full(
+                        (max_length_aligned_prompt - mask.shape[0],),
+                        0,
+                        dtype=mask.dtype,
```
shawnz, line 226: same as above, should also set device=self.device

shawnz (commented):
Hi there, I am just a user here who is looking forward to this change. However, I noticed that there is an error if the model is running on a GPU. I think it could be fixed by passing the device in these two statements here (at least, this fixes it for me)

rlouf:
We're getting really close. There are a few design changes remaining, and mostly we should have comprehensive tests before merging.

rlouf:
I rebased your branch on `main` after a big refactor of the `FSM` interface. I will take a closer look this week.

RobinPicard:
Is this still something we want to work on?

rlouf:
Yes! I'm currently thinking about how we could integrate that to the logits processors since most integration are going to use this :)

shawnz:
Sorry to prod but please don't lose sight of this! I think this is a very important change to make Outlines the most competitive structured generation system

rlouf:
I think it is time to revisit this as #966 is about to be merged and the custom sampling loop will be removed. We can still implement this via passing logit processors to downstream libraries. Incidentally that also means that we will be adding this feature to every upstream library :) 

@RobinPicard are you still interested in implementing this?

RobinPicard:
> @RobinPicard are you still interested in implementing this?

I can look at adapting it to the change made this week end

lapp0:
> > @RobinPicard are you still interested in implementing this?
> 
> I can look at adapting it to the change made this week end

That's great news!

Please let me know if you run into any issues or have any questions about `OutlinesLogitsProcessor`.

You probably want to branch from https://github.com/outlines-dev/outlines/pull/966 since it has fixes to the logits processors and a more detailed docstring.

RobinPicard:
To make sure I understand the wider context, the plan is to eventually remove `SequenceGenerator` and only use `SequenceGeneratorAdapter` @lapp0, right?

rlouf:
Indeed

RobinPicard:
I rebased on your branch and modified my initial commit @lapp0

rlouf:
Could you rebase on `main` now that #966 was merged?

RobinPicard:
This [issue](https://github.com/outlines-dev/outlines/issues/1048) is causing problems for the PR. If we don't have an explanation/solution for it, we would have to modify the logic related to `FSMLogitsProcessor._fsm_states`. I don't know yet what we could replace it with if the order of sequences is not maintained + the values of previous tokens can change though.

lapp0:
@RobinPicard per my comment in the linked issue, it appears that transformers beam search submits an unused sequence group to logits processors during the final generation step. Is this still an issue if it only occurs on the last step and it's not actually part of generation?

Please let me know how I can help.

RobinPicard:
It's fine if it only happens at the final generation step, I simply added a `try:... except KeyError:...` block

lapp0:
Glad that it's not blocking, please let me know if you run into any other issues or have any questions!

RobinPicard:
I don't have more questions, I would be interested by a review of the PR though!

outlines/integrations/llamacpp.py
```diff
@@ -33,7 +33,7 @@
 from numpy.typing import NDArray
 from pydantic import BaseModel
 
-from outlines.fsm.guide import CFGGuide, Guide, RegexGuide
+from outlines.fsm.guide import CFGGuide, Guide, RegexGuide, StopAtEOSGuide
```
lapp0, line 36: The individual integrations are being removed, they're no longer used. https://github.com/outlines-dev/outlines/issues/1054

lapp0 (changes requested):
I think there might be a better way to implement this

Could you give me your thoughts on the following design:

If alignment is enabled, in
- 1) `SequenceGeneratorAdapter`: Update the prompt token IDs in accordance with alignment, track `removedchars`
- 2) `guide.py`: Create `TokenHealingGuide` which enforces generation starting with `removedchars`, and passes the "unhealed" tokens to the original Guide.
- 3) `SequenceGeneratorAdapter`: left-truncate `len(removedchars)` once generation is complete

This would be forward-compatible with CFG, and minimize the the context of the token healing to just `SequenceGeneratorAdapter` and a `TokenHealingGuide`.

RobinPicard:
I don't get how step 2 would work. Are the unhealed tokens passed down to the original Guide just the user prompt?

Another design I'm considering for better separation of token healing and Guides is to create a dedicated class and have Guides that implement token healing inherit from it on top of GuideProtocol. Do you think that would be better?

outlines/fsm/guide.py
```diff
@@ -96,24 +102,69 @@ def __init__(self, tokenizer: "Tokenizer"):
...
+                else self.final_state
+                for token_id in self.vocabulary.values()
+            }
+        }
+
+    def align_prompt_tokens(self, prompt: str, tokenizer: "Tokenizer") -> str:
+        """Update the states_to_token_maps and return the aligned prompt"""
+        token_ids, _ = tokenizer.encode(prompt)
+        # possible return types of tokenizers include list, 1d Tensor and 2d Tensor
+        if not isinstance(token_ids, list):
```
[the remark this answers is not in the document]
RobinPicard: I've made a commit for the encode/decode methods of the tokenizer

outlines/fsm/guide.py
```diff
@@ -96,24 +102,69 @@ def __init__(self, tokenizer: "Tokenizer"):
...
 
-        return self.start_state
+        return self.states_to_token_maps[state][token_id]
 
     def is_final_state(self, state: int):
         return state == self.final_state
 
     def copy(self):
-        return self
+        return deepcopy(self)
```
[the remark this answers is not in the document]
RobinPicard: Right, I don't know why I became convinced I needed that.

outlines/generate/api.py
```diff
@@ -461,6 +461,47 @@ def prepare_generation_parameters(
 
         return generation_params
 
+    def strip_completions(
+        self,
+        completions,
+        prompts: Union[str, List[str]],
```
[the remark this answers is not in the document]
RobinPicard: This is already a method of `SequenceGeneratorAdapter`

outlines/fsm/guide.py
```diff
@@ -334,6 +418,10 @@ def __init__(self, cfg_string: str, tokenizer):
         self.start_state = 0
         self.final_state = -1
 
+    def align_prompt_tokens(self, prompt: str, tokenizer: "Tokenizer") -> str:
+        """Not applicable to this type of Guide"""
```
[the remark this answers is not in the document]
RobinPicard: There's no `states_to_token_map` available before the start of the generation and there's the problem that the crossing tokens are not valid in the context free grammar. It's probably possible to implement it for `CFGGuide`, I just don't know how to do it at the moment.

outlines/generate/api.py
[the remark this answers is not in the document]
lapp0: What I mean is that none of the logic of prompt/generation alignment should exist in `SequenceGenerator`

outlines/fsm/guide.py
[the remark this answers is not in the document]
lapp0: Great!

Could you please add a test which verifies correct functionality exists within `tests/generate/test_generate.py` against all `outlines.models`?

```
@pytest.mark.parametrize("pattern", REGEX_PATTERNS)
@pytest.mark.parametrize("model_fixture", ALL_MODEL_FIXTURES)
def test_generate_token_healing_regex(request, model_fixture, pattern):
    model = request.getfixturevalue(model_fixture)
    # similar to test_generate_regex, but verifies a match of the pattern is generated with token healing as well
```

lapp0:
>I don't get how step 2 would work. Are the unhealed tokens passed down to the original Guide just the user prompt?

>Another design I'm considering for better separation of token healing and Guides is to create a dedicated class and have Guides that implement token healing inherit from it on top of GuideProtocol. Do you think that would be better?

**Here's roughly what I'm thinking** 

In `SequenceGeneratorAdapter`
- set the prompt as `alignment_guide.prompt_prefix`
- then call `model.generate`, 
- then left-truncate generations

In `AlignmentGuide`
- Constructor: `AlignmentGuide(prompt, tokenizer, child_guide=None)`
  - Let `self.start_generation_tokens` be the set of legal starting tokens for `child_guide` (or all tokens if no guide)
  - Tokenize (prompt, token) for token in `start_generation_tokens` and determine the "longest common `self.prompt_prefix`" and the set of legal `self.prompt_suffix_ids` 
  - `self.initial_state = AlignmentGuideState(legal_paths=legal_start_generation_tokens, child_guide_state=child_guide.initial_state)`
- For `get_next_instruction(state)`
  - If `legal_paths` None, alignment is complete. Defer to the child_guide.
  - If `legal_paths` is not None, tokens = `[path[0] for path in state.legal_paths]`
- For `get_next_state(state, token_id)`
  - If legal paths: filter `state.legal_paths` such that it only includes those starting with `token_id`
    - If generating the final token in a path, pass the **unaligned** token to the child_guide, return `AlignmentGuideState(legal_paths=none, child_guide_state=new_child_guide_state)
 - If no `legal_paths is None`: Use the child guide to update `child_guide_state`

This design requires constructing a Guide for each prompt, however this is necessary because the behavior of the Guide varies prompt-to-prompt. 

Details surrounding how to pre-process and manage state for `AlignmentGuide` may vary, this is just a rough outline.

Please let me know if this makes sense, or if I'm missing something.

RobinPicard:
I think I understand the general idea. The main issue I have concerns

> * If generating the final token in a path, pass the **unaligned** token to the child_guide

I don't get what the unaligned token means in this context.

lapp0:
> I don't get what the unaligned token means in this context.

Sorry I wasn't clear here.

For example, if the prompt is "hello wo" and we truncate it to "hello" for alignment, and the model generates the token " world" as the next output, "rld" is what needs to be passed to the child guide to continue processing.

This allows for alignment to be compatible with any guide with minimal changes.

RobinPicard:
Ah I see, but I thought the problem is that "rld" may not exist as a token in the vocabulary so it would not be found in the `states_to_tokens_map`. Or should we walk through the `states_to_tokens_map` of the child guide character by character in the `get_next_state` method of `AlignmentGuide` when the generation reached the crossing token?

lapp0:
Sorry for the delayed response.

>but I thought the problem is that "rld" may not exist as a token in the vocabulary so it would not be found in the states_to_tokens_map.

If I'm understanding the problem correctly, to mitigate this we need to determine the "longest common prompt prefix". This will allow any legal token to be generated as a "pseudo-token".

>Or should we walk through the states_to_tokens_map of the child guide character by character in the get_next_state method of AlignmentGuide when the generation reached the crossing token?

Can we precompute this when the guide is constructed?

PR #533, opened 2024-01-13
Allow to generate several samples for each prompt

### LINKED ISSUE #416: Restore the ability to draw multiple samples with Open Source models
This was removed in #366 to simplify the PR, and should be added again. This will require to be careful with the shape and an added dimension will need to be added to take sample shape into account. The mechanism implemented there can be re-used when implementing Beam Search #258

rlouf:
Closes #416 

### TODO

- [x] Use `torch.repeat_interleave` so samples of the same sequence in a batch are contiguous (easier for beam search)
- [x] Make sure the FSMs are corrrectly duplicated

### Questions

- Do we make the sequence generator return arrays of shape `(num_samples * batch_size, num_tokens)`, and reshape in `SequenceGenerator`? Reshaping in the generator makes the latter much more complex. We could also not reshape at all, and let the user do it manually, which will simplify chained calls in the future.

### DIFF
--- a/outlines/generate/api.py
+++ b/outlines/generate/api.py
@@ -20,7 +20,6 @@ class SequenceGenerator:
         sampler,
         device,
         *,
-        num_samples: int = 1,
         max_tokens=None,
         stop_at=None,
     ):
@@ -29,7 +28,7 @@ class SequenceGenerator:
         self.tokenizer = model.tokenizer
         self.device = device
         self.max_tokens = max_tokens
-        self.num_samples = num_samples
+        self.num_particles = sampler.particles
 
         if isinstance(stop_at, str):
             stop_at = [stop_at]
@@ -167,7 +166,6 @@ class SequenceGenerator:
         prompts: Union[str, List[str]],
         max_tokens: Optional[int] = None,
         stop_at: Optional[Union[str, List[str]]] = None,
-        num_samples: int = 1,
         rng: Optional[torch.Generator] = None,
         kv_cache: Optional[torch.tensor] = None,
     ) -> Union[str, List[str], List[List[str]]]:
@@ -210,6 +208,11 @@ class SequenceGenerator:
 
         stop_sequences = stop_at or self.stop_sequences
         max_tokens = max_tokens or self.max_tokens
+<<<<<<< HEAD
+=======
+        num_samples = self.num_particles
+        num_sequences = len(prompts)
+>>>>>>> c8bf540 (Pass number of samples via sampler)
         fsms = [self.fsm.copy() for _ in prompts]
 
         if rng is None:
@@ -276,7 +279,6 @@ class SequenceGenerator:
         prompts: Union[str, List[str]],
         max_tokens: Optional[int] = None,
         stop_at: Optional[Union[str, List[str]]] = None,
-        num_samples: int = 1,
         rng: Optional[torch.Generator] = None,
         kv_cache: Optional[torch.tensor] = None,
     ) -> Iterator[Union[List[str], List[List[str]], str]]:
@@ -318,6 +320,11 @@ class SequenceGenerator:
 
         stop_sequences = stop_at or self.stop_sequences
         max_tokens = max_tokens or self.max_tokens
+<<<<<<< HEAD
+=======
+        num_samples = self.num_particles
+        num_sequences = len(prompts)
+>>>>>>> c8bf540 (Pass number of samples via sampler)
         fsms = [self.fsm.copy() for _ in prompts]
 
         if rng is None:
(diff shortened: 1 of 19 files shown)


outlines/generate/generator.py
```diff
@@ -78,6 +79,17 @@ def sequence_generator(
...
     token_ids, attention_masks, kv_cache = init_state
+    batch_shape = token_ids.shape[:-1]
+
+    # To take several samples we duplicate `token_ids`, `attention_masks`
+    # and `fsm_states` as many times as the number of samples requested.
+    # The resulting tensors are of shape (num_samples * num_batches, num_tokens)
+    token_ids = torch.repeat_interleave(token_ids, num_samples, dim=0)
+    attention_masks = torch.repeat_interleave(attention_masks, num_samples, dim=0)
+
+    fsm_states = fsm_states * num_samples
```
rlouf, line 90: Repeat in the same way we repeat `fsms`

outlines/generate/api.py
```diff
@@ -225,8 +242,9 @@ def __call__(
                 break
 
         generated_token_ids = self.get_generated_token_ids(
```
rlouf, line 244: Include `tokenizer.decode` in this function to make the code more understandable.

outlines/generate/api.py
```diff
@@ -225,8 +242,9 @@ def __call__(
...
-            init_state, prompts, last_state
+            init_state, prompts, last_state, num_samples
         )
+
         generated = self.tokenizer.decode(generated_token_ids)
         stripped = [
```
rlouf, line 249: Iterate over sequences inside the function for readability

outlines/generate/api.py
```diff
@@ -225,8 +242,9 @@ def __call__(
...
             self.strip_stop_sequences(sequence, stop_sequences)
```
rlouf, line 250: Same, also for `format_sequence` (and catch the exception inside)

lapp0 (commented):
I'm still learning the `outlines/generate/` side of the code-base. If there are any components you would like test cases for, that would help me learn :)

IMHO we shouldn't reshape at all. Users should expect an array of results with the same length as their input prompts. A shape-agnostic decoder should be a final step so we aren't required to convert to decodable array and back again in multiple places.

outlines/generate/api.py
```diff
@@ -14,16 +14,26 @@
...
+        *,
+        max_tokens=None,
+        stop_at=None,
+    ):
         self.generate_token = token_generator(model, sampler)
         self.fsm = fsm
         self.tokenizer = model.tokenizer
         self.device = device
         self.max_tokens = max_tokens
+        self.num_particles = sampler.particles
```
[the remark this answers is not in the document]
rlouf: I'm trying to find a word that works for multinomial, beam search and SMC :)

outlines/generate/api.py
```diff
@@ -60,18 +71,28 @@ def get_generated_token_ids(
...
-        prompt_lengths = [len(prompt_token_ids[i]) for i in range(len(prompts))]
+        prompt_lengths = [
+            len(prompt_token_ids[i])
+            for _ in range(num_samples)
+            for i in range(len(prompts))
+        ]
+
+        # We flatten the obtained token_ids since the tokenizer's decoder
+        # only accepts tensor with two dimensions
+        token_ids = last_state.token_ids.reshape((-1, last_state.token_ids.shape[-1]))
```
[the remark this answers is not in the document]
rlouf: How would that work? Currently the `sequence_generator` yields arrays of shape `(n_samples, n_batches, n_tokens)` but `batch_decode` will only accept 2-dimensional arrays.

outlines/generate/generator.py
```diff
@@ -55,7 +58,8 @@ def sequence_generator(
     fsms: List["FSM"],
     init_state: Tuple,
     fsm_states: List[FSMState],
-    rng: torch.Generator,
+    num_samples: int = 1,
```
rlouf, line 61: While I think `sequence_generator` should be able to take inputs of any shape, I am not sure it should be aware of `num_samples`.

outlines/generate/api.py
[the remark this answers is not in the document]
lapp0: I'm suggesting  a wrapper for `batch_decode` which reshapes any N-dim input into a 2-dim format for decoding, then reverts the decoded output back to its original dimensions to be returned.

This means we can maintain the structure of our data throughout the process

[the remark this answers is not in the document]
rlouf: The issue with this is that we need to modify every post-processing function. I am currently leaning towards handling the multiple samples outside of `generators`, which would simplify the code a lot.

Issue #524, opened 2024-01-11
VLLM tensor-parallel and RegexLogitsProcessor


BenoitHardier:
### Describe the issue as clearly as possible:

Hi, 
I recently tried to use the RegexLogitsProcessor with VLLM introduced by #481.

When using it with a "small" model like 7B one on a unique GPU it works fine but when I try with a big one, namely Mixtral, on multiple GPUs with the vllm engine argument tensor-parallel, I ran into several problems (monkey patching not working and fsm_state in the Processor not initialize). I suspect the multiple workers of Ray to be the cause (monkey patching may not be propagated to all the workers same for the fsm_states).

I could have missed some relevant information but It seems that #481 only checks without tensor-parallel. 

### Steps/code to reproduce the bug:

```python
import vllm
import vllm.model_executor.layers.sampler as sampler
from pydantic import BaseModel

from outlines.serve.vllm import JSONLogitsProcessor, _patched_apply_logits_processors

# Patch the _apply_logits_processors so it is compatible with `JSONLogitsProcessor`
sampler._apply_logits_processors = _patched_apply_logits_processors


class User(BaseModel):
    id: int
    name: str

model = "mistralai/Mixtral-8X7B-Instruct-v0.1"
#model = "mistralai/Mistral-7B-Instruct-v0.2"
llm = vllm.LLM(model=model, dtype='float16', max_model_len=1024,  tensor_parallel_size=4, max_num_seqs=512, enforce_eager=True)
logits_processor = JSONLogitsProcessor(User, llm)
result = llm.generate(
    ["A prompt", "Another prompt"],
    sampling_params=vllm.SamplingParams(
        max_tokens=100, logits_processors=[logits_processor]
    ),
)
print(result)
```

### Expected result:

```shell
result from vllm
```

### Error message:

```shell
Traceback (most recent call last):
  File "/workspace/./python_scripts/vllm_integration.py", line 19, in <module>
    result = llm.generate(
  File "/usr/local/lib/python3.10/dist-packages/vllm/entrypoints/llm.py", line 165, in generate
    return self._run_engine(use_tqdm)
  File "/usr/local/lib/python3.10/dist-packages/vllm/entrypoints/llm.py", line 185, in _run_engine
    step_outputs = self.llm_engine.step()
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/llm_engine.py", line 581, in step
    output = self._run_workers(
[...]

PR #539, opened 2024-01-15
Draft: fix beam search and multiple concurrent sequences using token_id tuple cache key

lapp0:
- Fixes beam search (and other parallel sequence generation) in vLLM
- removes incompatibility with vLLM by removing `seq_id` argument

TODO: 
- remove `_patched_apply_logits_processors` entirely, it doesn't do anything anymore.


rlouf:
The title contains "draft" but this is not a draft PR. Does that mean it is ready for review?

viktor-ferenczi:
I cannot fork your fork, therefor pasting the working fix to `RegexLogitsProcessor` here. It fixes the `FSMState` cache logic and tested to work for JSON schema and regex.

```py
class RegexLogitsProcessor:
    def __init__(self, regex_string, llm):
        """Compile the FSM that drives the regex-guided generation.

        Parameters
        ----------
        regex_string
            A string that represents a regular expression
        llm
            An instance of `vllm.LLM`

        """
        tokenizer = self.adapt_tokenizer(llm.tokenizer)

        fsm = RegexFSM(regex_string, tokenizer)
        self.fsm = fsm
        self.fsm_state_cache: Dict[int, FSMState] = {}

    def __call__(self, input_ids: List[int], scores: torch.Tensor) -> torch.Tensor:
        """Use the FSM to bias the logits before sampling the next token."""
        state = self.get_fsm_state(tuple(input_ids))
        allowed_tokens = self.fsm.allowed_token_ids(state)

        mask = torch.full((scores.shape[-1],), -math.inf, device=scores.device)
        mask[allowed_tokens] = 0
        biased_scores = scores + mask

        return biased_scores

    def get_fsm_state(self, input_ids: Tuple[int]) -> FSMState:
        if not input_ids:
            return FSMState(0)

        state_key = hash(input_ids)
        state = self.fsm_state_cache.get(state_key)
        if state is not None:
            return state

        prev_input_ids = input_ids[:-1]
        prev_state = self.get_fsm_state(prev_input_ids)

        last_token = input_ids[-1]
        state = self.fsm.next_state(prev_state, last_token)

        self.fsm_state_cache[state_key] = state
        return state

    def adapt_tokenizer(self, tokenizer):
        """Adapt vLLM's tokenizer to use to compile the FSM.

        The API of Outlines tokenizers is slightly different to that of
        `transformers`. In addition we need to handle the missing spaces to
        Llama's tokenizer to be able to compile FSMs for this model.

        """
        tokenizer.vocabulary = tokenizer.get_vocab()
        tokenizer.special_tokens = set(tokenizer.all_special_tokens)

        def convert_token_to_string(token: str) -> str:
            from transformers.file_utils import SPIECE_UNDERLINE

            string = tokenizer.convert_tokens_to_string([token])

            # A hack to handle missing spaces to HF's Llama tokenizers
            if token.startswith(SPIECE_UNDERLINE) or token == "<0x20>":
                return " " + string

            return string

        tokenizer.convert_token_to_string = convert_token_to_string

        return tokenizer
```

viktor-ferenczi:
I keep the working code in my `dev` branch here, so you can cherry-pick the fix from there as well: https://github.com/viktor-ferenczi/outlines/tree/dev

viktor-ferenczi:
@lapp0 Could you please consider the above fix? It works for me. I don't want to hijack your PR, but we need this out of draft and get reviewed. Thanks!

lapp0:
> @lapp0 Could you please consider the above fix? It works for me. I don't want to hijack your PR, but we need this out of draft and get reviewed. Thanks!

Today I will test, and given success incorporate your changes. Additionally, I will start work on tensor parallel.

viktor-ferenczi:
We also need to get some upper limit on the number of recursions here:
```py
prev_state = self.get_fsm_state(prev_input_ids)
```

The default recursion limit on Python 3.10 is 1000 and I don't see we change it.

If the maximum number of possible recursions in the above code is in the hundreds, then it may be better to rewrite the code to avoid using recursions. It is easy to do so while preserving the logic. 

I've been running this with huge regex constraints and haven't see any problems yet. But I haven't use a JSON schema constraint with it yet aside of a single test case I have to cover this constraint mode in my project.

tests/generate/test_vllm_regex_logits_processor.py
```diff
@@ -0,0 +1,88 @@
+import re
```
rlouf, line 1: Can you name this file `test_vllm_logits_processor.py` and move it to `tests/serve`?

tests/generate/test_vllm_regex_logits_processor.py
```diff
@@ -0,0 +1,88 @@
...
+    return torch.multinomial(probs, 1).item()
+
+
+def test_time_regexp():
+    pattern = r"(0?[1-9]|1[0-2]):[0-5]\d\s?(am|pm)?"
+    llm = MockModel()
+    logits_processor = RegexLogitsProcessor(pattern, llm)
+
+    token_ids = []
+    while True:
```
rlouf, line 24: Coulnd't you just generate a few tokens here?

tests/generate/test_vllm_regex_logits_processor.py
```diff
@@ -0,0 +1,88 @@
+import re
+
+import torch
+from transformers import AutoTokenizer
+
+from outlines.serve.vllm import RegexLogitsProcessor, _patched_apply_logits_processors
+
+
+class MockModel:
```
rlouf, line 9: These look more like integration tests than unit tests. We should write unit tests instead.

lapp0:
> > Previous implementation assumed the logits processor always saw `prev_input_ids`.
> 
> That's exactly why it crashed when I used your code here. It does not seem to be the case all the time for some reason, but I don't know why.
> 
> > I think we should `KeyError` if `prev_input_ids` isn't cached. We should assume we have seen the predecessor, because otherwise we are re-parsing the entire generation for each token generated.
> 
> I will test your latest version here and see whether it works. I guess it will `KeyError`, but we'll see...

If it fails, could you give an example of an api call which results in failure so I can debug?

viktor-ferenczi:
Yeah, exactly. Got a `KeyError` as expected:
```sh
...
  File "/home/viktor/env/outlines/lib/python3.10/site-packages/vllm/model_executor/layers/sampler.py", line 172, in _apply_logits_processors
    logits_row = logits_processor(token_ids, logits_row)
  File "/home/viktor/dep/outlines-contrib/outlines/serve/vllm.py", line 61, in __call__
    state = self.get_fsm_state(input_ids)
  File "/home/viktor/dep/outlines-contrib/outlines/serve/vllm.py", line 78, in get_fsm_state
    prev_state = self.fsm_state_cache[prev_state_key]
KeyError: 5740354900026072187
...
```

This is exactly why I added that recursion to fill up any previous token missing in the cache.

Request was:
```json
{
  "prompt": "You are an AI programming assistant, utilizing the Deepseek Coder model, developed by Deepseek Company, and you only answer questions related to computer science. For politically sensitive questions, security and privacy issues, and other non-computer science questions, you will refuse to answer.\n\nYou are a helpful AI assistant. You give concise answers. If you do not know something, then say so.\n### Instruction:\nWrite down the first 10 prime numbers as a comma separated list, starting with 2.\n\n### Response:\n",
  "n": 1,
  "best_of": 1,
  "presence_penalty": 0.0,
  "frequency_penalty": 0.0,
  "repetition_penalty": 1.0,
  "temperature": 0.0,
  "top_p": 1.0,
  "top_k": -1,
  "min_p": 0.0,
  "use_beam_search": false,
  "length_penalty": 1.0,
  "early_stopping": false,
  "stop": [],
  "stop_token_ids": [],
  "include_stop_str_in_output": false,
  "ignore_eos": false,
  "max_tokens": 50,
  "logprobs": null,
  "prompt_logprobs": null,
  "skip_special_tokens": true,
  "spaces_between_special_tokens": true,
  "regex": "\\d+(\\s*,\\s*\\d+)*\\s*"
}
```

Model: `TheBloke/deepseek-coder-33B-instruct-AWQ`

GPUs: 2x4090 (2x24GB)

vLLM command:
```sh
python -O -u -m outlines.serve.serve \
  --model=TheBloke/deepseek-coder-33B-instruct-AWQ \
  --quantization=awq \
  --dtype=float16 \
  --host=0.0.0.0 \
  --port=8000 \
  --max-model-len=16384 \
  --max-num-seqs=16 \
  --tensor-parallel-size=2 \
  --swap-space=8 \
  --gpu-memory-utilization=0.95 \
  --enforce-eager \
  --disable-log-requests
```

lapp0:
@viktor-ferenczi thank you, will investigate.

viktor-ferenczi:
@lapp0 Caught the `KeyError` and printed the relevant variable values:
```
(RayWorkerVllm pid=205777) state_key = 4477961998282403984
(RayWorkerVllm pid=205777) input_ids = [17]
(RayWorkerVllm pid=205777) self.fsm_state_cache = {}
```

So it fails on the very first token, when there has been no previous tokens.

The `KeyError` is raised at this code line:
```py
prev_state = self.fsm_state_cache[prev_state_key]
```

The `prev_state_key` here is `hash( () )`, e.g. the hash of an empty tuple, which is constant: `5740354900026072187`

Certainly this item is not present in in `self.fsm_state_cache`, because the cache is empty.

It means that the body of this if condition is never executed:
```py
        if not input_ids:
            self.fsm_state_cache[state_key] = FSMState(0)
```

lapp0:
> @lapp0 Caught the `KeyError` and printed the relevant variable values:
> 
> ```
> (RayWorkerVllm pid=205777) state_key = 4477961998282403984
> (RayWorkerVllm pid=205777) input_ids = [17]
> (RayWorkerVllm pid=205777) self.fsm_state_cache = {}
> ```
> 
> So it fails on the very first token, when there has been no previous tokens.
> 
> The `KeyError` is raised at this code line:
> 
> ```python
> prev_state = self.fsm_state_cache[prev_state_key]
> ```
> 
> The `prev_state_key` here is `hash( () )`, e.g. the hash of an empty tuple, which is constant: `5740354900026072187`
> 
> Certainly this item is not present in in `self.fsm_state_cache`, because the cache is empty.
> 
> It means that the body of this if condition is never executed:
> 
> ```python
>         if not input_ids:
>             self.fsm_state_cache[state_key] = FSMState(0)
> ```
> 
> I think my fix was correct, unless it can be explained why the above happens.

Makes sense, the only variation from the previous implementation was that a defaultdict is no longer used. Your recursive solution alleviated this, but my removal of the recursion resulted in the error. 

For the mentioned reason I still don't think we should recurse.

Before I push a new test case and specific handling for empty previous token IDs, could you confirm that you never received the log line

```
 input_ids = []
```

on a separate ray worker?

I want to eliminate the possibility that ray workers duplicated the logits processors resulting in two separate state caches. I'm not even sure how you have a vLLM world size of 2, as I ran into issues with this until I made the logits processor a separate `ray.actor`.

viktor-ferenczi:
Fortunately I kept all the logs from my tests today and can confirm that they don't contain any lines with `input_ids = []` in it.

I had crashes before with `defaultdict` as well, just different ones. Anyway, I don't like the recursion either, so any solution which works without that would be perfect. 

What I see is that for some reason the cache is not initialized here with the empty token case, which it was initialized in your tests. Why is that difference?

Are we still sticking to vLLM `0.2.6`? 

I'm using that version, because it was mentioned on the doc page that `outlines.serve.serve` requires that version.

rlouf:
You should be able to use vLLM `0.2.7`, I opened a PR to change the doc #547

lapp0:
@viktor-ferenczi can confirm, the recursive solution reconstructs the grammar from start every time

https://gist.github.com/lapp0/8370bde4d977088487c34bc7501b78af

rlouf:
What is the status on this? I have a hard time following the discussion 😬

viktor-ferenczi:
@rlouf @lapp0 The latest progress seems to be mentioned in #524

The latest code discussed there is working for me on 2x4090 with `--tensor-parallel=2`, both with multiple parallel sequences and beam search. Performance looks good. So I guess this is only about integrating the changes into this PR and finishing up the tests.

lapp0:
Need to merge Viktors changeset. A few people have pinged me about this so it's my priority in terms of Outlines, but I have a few things to take care of for my job. Will try to get it done today or tommorw though.

lapp0:
Two test errors are in Lark `tests/test_grammars.py`

```
E       AttributeError: 'ParserConf' object has no attribute 'deterministic'
```

`lark==1.1.9` was released 6 weeks ago and works on my machine. Investigating...

lapp0:
@viktor-ferenczi it seems the latest version of your `dev` branch no longer uses the `diskcache`, which resulted in

```
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 28, in _raise_exception_on_finish                                                                
    task.result()                                                                                                                                                                       
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 359, in run_engine_loop                                                                          
    has_requests_in_progress = await self.engine_step()                                                                                                                                 
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 338, in engine_step                                                                              
    request_outputs = await self.engine.step_async()                                                                                                                                    
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 191, in step_async                                                                               
    output = await self._run_workers_async(                                                                                                                                             
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/async_llm_engine.py", line 219, in _run_workers_async                                                                       
    all_outputs = await asyncio.gather(*coros)                                                                                                                                          
  File "/usr/lib/python3.10/asyncio/tasks.py", line 650, in _wrap_awaitable                                                                                                             
    return (yield from awaitable.__await__())                                                                                                                                           
ray.exceptions.RayTaskError(KeyError): ray::RayWorkerVllm.execute_method() (pid=10479, ip=172.17.0.2, actor_id=55888572aea4240caa3da5af01000000, repr=<vllm.engine.ray_utils.RayWorkerVl
lm object at 0x7f789717a4d0>)                                                                                                                                                           
  File "/usr/local/lib/python3.10/dist-packages/vllm/engine/ray_utils.py", line 31, in execute_method
    return executor(*args, **kwargs)                                                       
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)                                                           
  File "/usr/local/lib/python3.10/dist-packages/vllm/worker/worker.py", line 159, in execute_model
    output = self.model_runner.execute_model(seq_group_metadata_list,                      
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)                                                           
  File "/usr/local/lib/python3.10/dist-packages/vllm/worker/model_runner.py", line 354, in execute_model
    output = self.model.sample(                                                            
  File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/models/opt.py", line 313, in sample
    next_tokens = self.sampler(self.lm_head_weight, hidden_states,                         
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)                                                
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)                                                   
  File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/layers/sampler.py", line 52, in forward
    logits = _apply_logits_processors(logits, sampling_metadata)                           
  File "/usr/local/lib/python3.10/dist-packages/vllm/model_executor/layers/sampler.py", line 172, in _apply_logits_processors
    logits_row = logits_processor(token_ids, logits_row)                                   
  File "/root/outlines/outlines/serve/vllm.py", line 119, in __call__                      
    state = self.fsm.get_state_by_token_ids(tuple(input_ids))                              
  File "/root/outlines/outlines/serve/vllm.py", line 81, in get_state_by_token_ids         
    prev_state = self.state_cache[prev_state_key]                                          
KeyError: 5740354900026072187                                                              
```

Have you reproduced this error? Which revision of your `dev` branch would you recommend?

lapp0:
Actually it works on `vllm==0.2.7`.

Remaining work
- [ ] determine why it works on `vllm==0.2.7`
- [ ] pin `vllm`
- [ ] Fix ParserConf tests

viktor-ferenczi:
@lapp0 Updated my dev branch to match your branch, plus made a minor change for vLLM 0.3.0 compatibility:
https://github.com/viktor-ferenczi/outlines/tree/dev

Please consider taking my changes from there.

What is missing to finish this PR? Only the tests?

lapp0:
@viktor-ferenczi tests were fixed in https://github.com/outlines-dev/outlines/pull/596

For vLLM 0.3.0, could you let me know in https://github.com/outlines-dev/outlines/pull/579

Are there any other changes you would like on this branch before it's merged? If not, let me know. If so, could you please provide the diff since your branch has a variety of other branches merged in it?

viktor-ferenczi:
@lapp0 Are you sure you pulled my `dev` branch again? (I had to force-push it.)

This was the minor change for vLLM 0.3.0 compatibility:
https://github.com/viktor-ferenczi/outlines/commit/c8b68c4100a6b64d2903b673e8f2d466d95f3194

It belongs here, because it is meaningful only in the context of changes made by this PR.

viktor-ferenczi:
@lapp0 

Got this while trying to use beam search:
```
  File "/home/viktor/env/outlines/lib/python3.10/site-packages/vllm/model_executor/layers/sampler.py", line 74, in forward
    logits = _apply_logits_processors(logits, sampling_metadata)
  File "/home/viktor/dep/outlines-contrib/outlines/serve/vllm.py", line 33, in _patched_apply_logits_processors
    logits_row = logits_processor(token_ids, logits_row)
  File "/home/viktor/dep/outlines-contrib/outlines/serve/vllm.py", line 123, in __call__
    state = self.fsm.get_state_by_token_ids(tuple(input_ids))
  File "/home/viktor/dep/outlines-contrib/outlines/serve/vllm.py", line 88, in get_state_by_token_ids
    new_state = self.next_state(prev_state, last_token)
  File "/home/viktor/dep/outlines-contrib/outlines/fsm/fsm.py", line 178, in next_state
    last_token_to_end_state = self.states_to_token_maps[state]
KeyError: 9
```

It sounds familiar. So something is up with vLLM 0.3.0 compatibility.

viktor-ferenczi:
How this could happen? No key `9` exists in the FSM:
```
self.states_to_token_maps = {0: {63: 1, 4686: 2, 10252: 3},
 1: {63: 2, 4686: 3},
 2: {63: 3},
 3: {185: 4},
 4: {47: 5, 63: 6, 1426: 11, 4686: 7, 5959: 10, 10252: 8, 16097: 15},
 5: {81: 10, 295: 11, 12483: 12},
 6: {63: 7, 4686: 8},
 7: {63: 8},
 8: {185: 9},
 10: {78: 11, 493: 12, 18596: 15},
 11: {70: 12, 877: 13, 1644: 15, 16795: 14},
 12: {81: 13, 401: 14, 3477: 15},
 13: {64: 14, 302: 15},
 14: {76: 15},
 15: {13: 16},
 16: {66: 17, 5494: 18},
 17: {82: 18},
 18: {185: 19},
 19: {63: 6, 4686: 7, 10252: 8}}
 ```

**Comment by viktor-ferenczi** [comment]
Fixed in #606

File: tests/serve/test_vllm.py [hunk 4]
```diff
@@ -0,0 +1,122 @@
...
+    tokenizer = MockTokenizer()
+
+
+def sample_from_logits(logits):
+    probs = torch.exp(logits) / torch.sum(torch.exp(logits))
+    return torch.multinomial(probs, 1).item()
+
+
+@pytest.mark.parametrize("forget_logits_processor", [True, False])
+def test_time_regexp(forget_logits_processor):
```
rlouf, line 46: What is this testing exactly?

tests/serve/test_vllm.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        new_token_id = sample_from_logits(logits)
+        if new_token_id == llm.tokenizer.eos_token_id:
+            break
+        token_ids.append(new_token_id)
+
+    assert re.fullmatch(pattern, llm.tokenizer.decode(token_ids)) is not None
+
+
+def test_time_regexp_multiple_samples():
```
rlouf, line 72: What is this testing?

pyproject.toml
```diff
@@ -60,7 +60,7 @@ test = [
     "huggingface_hub"
 ]
 serve = [
```
rlouf, line 62: Does it work with `vllm==0.3.0`

tests/serve/test_vllm.py
lapp0: In vLLM 0.2.6 logits processors would be applied on "random" ray worker. This would result in a regression because a new logits processor had to be created to be created for each worker, causing the loss of previous state.

This test simulates that scenario by creating a new logits processor for every token processed.

However I will have to experiment to determine whether this test and the changes to make it pass are still necessary, as vLLM now applies logits processors on the same worker each time as of vLLM 0.2.7 per https://github.com/outlines-dev/outlines/pull/539#issuecomment-1913752978

tests/serve/test_vllm.py
lapp0: I observed a lack of stability in sequence order when using beam search with Outlines. This resulted in a new token for one sequence being applied to a different sequence.

This test reproduces that behavior.

I will leave an explanatory doc string.

rlouf:
I don't see any issue linked to this PR, but maybe it closes a few?

lapp0:
@rlouf 

https://github.com/outlines-dev/outlines/pull/579 resolves https://github.com/outlines-dev/outlines/issues/524

PR #544, opened 2024-01-17
[WIP] Misc FSM Enhancements

benlipkin:
I'm opening a PR to include a few small misc FSM enhancements. These should each be minor, but all focus on improving the completeness and/or performance of the FSM classes.

___

In this first commit, I modify how `final_state` is determined. Previously, for `RegexFSM`, the implementation of `is_final_state` was `return state in self.final_states`. This logic is incorrect, since generation should not terminate on any accepting state; it should only terminate when an EOS has been generated (else `*` is reduced to `?`). Here are the counts of unique outputs on an example generation of 1000 samples from the regex `a*`

```
['', 'a']
[557, 443]
```

Instead, now for the FSM interface, we set `final_state = FSMState(-1)`, and in `next_state`, `if token_id == self.eos_token_id: return self.final_state` and `is_final_state` can now be implemented as `return state == self.final_state`. Here are 1000 new samples, which are now correctly geometrically distributed.

```
['', 'a', 'aa', 'aaa', 'aaaa', 'aaaaa', 'aaaaaa', 'aaaaaaa', 'aaaaaaaa', 'aaaaaaaaa']
[524, 192, 101,  85,  45,  26,  11,   4,   6,   6]
```

While making this modification, I also cleaned up the code in a few more places.

Now that `stop_at` is handled by `SequenceGenerator` (thanks to https://github.com/outlines-dev/outlines/pull/451), `StopAtTokenFSM` is only ever used to stop at `EOS`. I've made this explicit. Next, with both of the other FSMs using this new implementation, I brought `CFGFSM` on board as well (which also reduces some messiness and brings it closer to the rest of the interface) and moved this implementation to the abstract protocol level. This also allowed us to get rid of some hard-coded `-1` states floating around in the code by just defining it once at the interface.


lapp0:
Nice!state_to_token_maps: {0: {46071: 1, 64: 1, 24794: 1, 50256: 0, 7252: 1}, 1: {46071: 1, 64: 1, 50256: 1, 24794: 1, 7252: 1}}
finals: frozenset({0, 1, -1})
allowed(0): [46071, 64, 24794, 50256, 7252]
allowed(1): [46071, 64, 50256, 24794, 7252]
eos_token_id: 50256


One thing to be mindful of as you develop this PR, make sure the eos_token_id is in `allowed()`

```
tokenizer = TransformerTokenizer("gpt2")
rfsm = RegexFSM("a*", tokenizer)
print("state_to_token_maps:", rfsm.states_to_token_maps)
print("finals:", rfsm.final_states)
print("allowed(0):", rfsm.allowed_token_ids(0))
print("allowed(1):", rfsm.allowed_token_ids(1))
print("eos_token_id:", tokenizer.eos_token_id)
```

e.g. on main this prints

```
state_to_token_maps: {0: {46071: 1, 64: 1, 24794: 1, 50256: 0, 7252: 1}, 1: {46071: 1, 64: 1, 50256: 1, 24794: 1, 7252: 1}}
finals: frozenset({0, 1, -1})
allowed(0): [46071, 64, 24794, 50256, 7252]
allowed(1): [46071, 64, 50256, 24794, 7252]
eos_token_id: 50256
```

benlipkin:
Okay, here's the second enhancement. I had raised in https://github.com/outlines-dev/outlines/pull/391 that the current CFG implementation was also incomplete because, in a setting where it could either extend the current terminal or start the next terminal, it would always start the next terminal. This would preclude important strings, e.g., multi-token variable names, etc. This has now been fixed by proposing both tokens that would extend the current terminal as well as tokens that would terminate the current terminal and start the next terminal. The LM can now sample to select which path to take. Both FSMs (tracking progress separately for each terminal) are stored and can then only the correct one is transitioned (and any stale ones removed) once we see what token the LM generates.

For example, for the following grammar:
```
grammar = r"""
    start: s
    s: /a+/
    | "(" s ")"
"""
```

previously, only the following strings were possible:

```
a
(a)
((a))
(((a)))
...
```

It would never be possible for the innermost string to extend as it would always be matched after a single character and only next terminals allowed.

Now, strings such as those in this next example set are supported:
```
(aaaaa)
(((aaa)))
((aa))
aaaa
...
```

I've updated documentation and added a new test case, now covering the example that used to `xfail`, and have also added new examples to `examples/cfg.py` including a dyck-like grammar similar to the one above, as well as JSON.

benlipkin:
Okay, this is it for now; ready for review.

I was also thinking of memoizing the `RegexFSM` objects constructed by `CFGFSM` in the class since these are often shared across a batch and over multiple rounds of generation. But, with the new caching of the state to token maps during `RegexFSM` construction, the actual time saved here is only ~10% in my initial profiling and it increases memory usage and code complexity. So, leaving it off for now. But, maybe worth thinking about more in the future.

Let me know any comments on the other two enhancements here.

Thanks!

lapp0 (commented):
Great work, LGTM.

lapp0:
IMHO, the other enhancements should be a separate PR.

benlipkin:
Thanks @lapp0 ! 

Sounds good re keeping this PR to completeness-focused enhancements and tackling performance-focused optimizations in a different PR. 

Yes, I would be interested in collaborating on this (thanks for sharing your vllm commit). Awesome to see some more benchmarking/profiling happening in https://github.com/outlines-dev/outlines/issues/549

lapp0:
I'm writing a few `CFGFSM` benchmark tests and none of my valid example json and python files could be generated. After merging this branch into my branch they **could** be generated.

@rlouf could you review when you get a chance?

rlouf:
Sorry for the time it took me to review the PR, I needed to make sure I understood the changes. This looks good to me, thank you for opening a PR!

PR #546, opened 2024-01-17
vLLM request parameters `constraint_mode` and `constraint`

viktor-ferenczi:
Introduced generic vLLM request parameters `constraint_mode` and `constraint` to replace `json_schema` and `regex`.

The main reason for the change is to allow for adding an arbitrary number of grammars types without having to add new parameters to the vLLM API every time. It also reduces the probability of a future collision with an original vLLM request parameter name. 

It also reduces confusion due to naming, like `cfg` meaning both "configuration" and "Context-Free Grammar", which was confusing at least to me. (This is for the upcoming PR to introduce full Lark grammar support to the vLLM API.)

Kept the old `json_schema` and `regex` parameters for backwards compatibility, but we should remove them soon.

### DIFF
--- a/docs/reference/vllm.md
+++ b/docs/reference/vllm.md
@@ -33,7 +33,8 @@ For example, to generate a string that matches the schema `{"type": "string"}` (
 curl http://127.0.0.1:8000/generate \
     -d '{
         "prompt": "What is the capital of France?",
-        "schema": {"type": "string"}
+        "grammar_mode": "json_schema",
+        "grammar": {"type": "string"}
         }'
 ```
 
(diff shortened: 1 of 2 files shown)


viktor-ferenczi:
We could also name these `constraint_mode` and `constraint`, then use `"grammar"` as a `constraint_mode` value for the full Lark grammar. What do you think about these names?

viktor-ferenczi:
@rlouf Could you please review this? Thanks!

rlouf:
I find this naming less explicit than the current one. It may be confusing.

viktor-ferenczi:
Fair enough. We'll see whether this will still be the case after adding the 5th logits processor.

rlouf:
We'll cross that bridge when we come to it 😁

PR #548, opened 2024-01-17
Modify llamacpp to process multiple invocations

msjgriffiths:
**Problem**: When running multiple calls to the `llamacpp` model, it segfaults.

```python
from outlines import models, generate
file = "zephyr-quiklang-3b-4k.Q3_K_L.gguf"
model = models.llamacpp(file)
generate.text(model, 20)(prompt)
> 'The sp Account trail restored employees hyster opioid zlikeurdy contracted Get portions says protpad youth Mongo +'
```

But if I call it again -- 
```python
generate.text(model, 20)(prompt)
```

Then I get an error.

```log
llama_decode_internal: n_tokens == 0llama_decode: failed to decode, ret = -1
could not find a KV slot for the batch (try reducing the size of the batch or increase the context)
zsh: segmentation fault  ipython
```

**Resolution**

There are two fixes:

  * Set `self.past=0` at the end of the `forward` method through the model.
  * Clear the `kv_cache` after freeing the batch.

Separately, I also modified the `decode` method to handle some edge cases when the function might receive a `torch.Tensor` or a `List[torch.Tensor]`.

TODO:
  * Run the `pre-commit` style checks.
  * Run `pytest`.

### DIFF
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -29,6 +29,7 @@ class LlamaCpp:
 
         self.ctx = llama_instance.llama_new_context_with_model(model, context_params)
 
+
     def forward(self, input_ids: torch.LongTensor, *_):
         """Compute a forward pass through the llama_cpp model."""
         if input_ids.ndim == 2:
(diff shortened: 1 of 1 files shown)


rlouf:
We're changing the way we integrate with llama.cpp in #556 and it should solve the issue.

rlouf:
I think most issues related to llama.cpp should be fixed in `outlines==0.0.32`. I will close for now but feel free to re-open if the underlying issue is still present.

PR #552, opened 2024-01-17
Fix bug in the treatment of the anyOf keywork in JSON schema

### LINKED ISSUE #497: Bug in the treatment of the `anyOf` keyword in JSON Schema

I ran into this issue while working on #495. It seems to be caused by the fact that the subregex generated for the `anyOf` keyword includes all permutations of the patterns of the types in question concatenated. I don't understand why we need to do such permutations. I have the impression that a simple "or" between the patterns associated to each type would work  

```python
import json
import jsonschema
import re
from outlines.fsm.json_schema import build_regex_from_object

schema = {
    "properties": {
        "foo": {"anyOf": [{"type": "null"}, {"type": "boolean"}]},
    },
    "required": ["foo"],
    "type": "object",
}

regex = build_regex_from_object(json.dumps(schema))
print(regex)

wrong_response = '{ "foo" : nulltrue }'

if re.fullmatch(regex, wrong_response):
    print("match")

try:
    jsonschema.validate(instance=wrong_response, schema=schema)
except jsonschema.exceptions.ValidationError as err:
    print("no match")
    
# \{[\n ]*"foo"[\n ]*:[\n ]*((null)|((true|false))|(null(true|false))|((true|false)null))[\n ]*\}
# match
# no match
```

### Expected result:

```shell
The wrong_response should not match with the regex
```

### Error message:

_No response_

```python
Outlines: current main branch
Python: 3.9.6
```

### Context for the issue:

_No response_

RobinPicard:
Fixes #497

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -1,5 +1,4 @@
 import inspect
-import itertools as it
 import json
 import re
 from typing import Callable, Union
@@ -80,7 +79,6 @@ def to_regex(resolver: Resolver, instance: dict):
     Note
     ----
     Many features of JSON schema are missing:
-    - Support the fact that fields in an object are optional by default
     - Handle `additionalProperties` keyword
     - Handle types defined as a list
     - Handle constraints on numbers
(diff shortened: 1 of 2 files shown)


rlouf:
Just adding a [link to the json schema specification](https://json-schema.org/understanding-json-schema/reference/combining#anyof) for reference. Thank you for opening a PR!

PR #556, opened 2024-01-18
Add llama logits processor

dtiarks:
Note that this is super early.
We should probably provide a core implementation that all the other processors could inherit. Also the implementation of the tokenizer seems to be redundant.

Looking forward to some input.

### DIFF
--- a/examples/llamacpp_processor.py
+++ b/examples/llamacpp_processor.py
@@ -1,9 +1,10 @@
 from enum import Enum
 
-import torch
+from llama_cpp import Llama, LogitsProcessorList
 from pydantic import BaseModel, constr
 
-import outlines
+from outlines.generate.processors import JSONLogitsProcessor
+from outlines.models.llamacpp import LlamaCppTokenizer
 
 
 class Weapon(str, Enum):
@@ -30,17 +31,20 @@ class Character(BaseModel):
 
 
 if __name__ == "__main__":
-    # curl -L -o mistral-7b-instruct-v0.2.Q5_K_M.gguf https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF/resolve/main/mistral-7b-instruct-v0.2.Q5_K_M.gguf
-    model = outlines.models.llamacpp("./mistral-7b-instruct-v0.2.Q5_K_M.gguf")
-
-    # Construct structured sequence generator
-    generator = outlines.generate.json(model, Character)
-
-    # Draw a sample
-    rng = torch.Generator(device="cpu")
-    rng.manual_seed(789005)
+    llama = Llama("./phi-2.Q4_K_M.gguf")
+    tokenizer = LlamaCppTokenizer(llama)
 
     prompt = "Instruct: You are a leading role play gamer. You have seen thousands of different characters and their attributes.\nPlease return a JSON object with common attributes of an RPG character. Give me a character description\nOutput:"
 
-    sequence = generator(prompt, rng=rng, max_tokens=512)
-    print(sequence)
+    logits_processor = JSONLogitsProcessor(Character, tokenizer)
+
+    json_str = llama.create_completion(
+        prompt,
+        top_k=40,
+        top_p=0.95,
+        temperature=0.7,
+        max_tokens=100,
+        logits_processor=LogitsProcessorList([logits_processor]),
+    )["choices"][0]["text"]
+
+    print(json_str)
--- a/outlines/generate/cfg.py
+++ b/outlines/generate/cfg.py
@@ -3,6 +3,7 @@ from functools import singledispatch
 from outlines.fsm.fsm import CFGFSM
 from outlines.generate.api import SequenceGenerator
 from outlines.models import OpenAI
+from outlines.models.llamacpp import CFGLogitsProcessor, LlamaCpp
 from outlines.samplers import Sampler, multinomial
 
 
@@ -29,6 +30,24 @@ def cfg(model, cfg_str: str, sampler: Sampler = multinomial()) -> SequenceGenera
     generator = SequenceGenerator(fsm, model, sampler, device)
 
     return generator
+
+
+@cfg.register(LlamaCpp)
+def cfg_llamacpp(
+    model: LlamaCpp,
+    cfg_str: str,
+    sampler: Sampler = multinomial(),
+):
+    if not isinstance(sampler, multinomial):
+        raise NotImplementedError(
+            r"The llama.cpp integration does not currently support any other sampling algorithm "
+            + "than the multinomial sampler."
+        )
+
+    logits_processor = CFGLogitsProcessor(cfg_str, model.tokenizer)
+    model.logits_processor = logits_processor
+
+    return model
 
 
 @cfg.register(OpenAI)
--- a/outlines/generate/choice.py
+++ b/outlines/generate/choice.py
@@ -13,7 +13,11 @@ def choice(
     model, choices: List[str], sampler: Sampler = multinomial()
 ) -> SequenceGenerator:
     regex_str = r"(" + r"|".join(choices) + r")"
-    return regex(model, regex_str, sampler)
+
+    generator = regex(model, regex_str, sampler)
+    generator.format_sequence = lambda x: x
+
+    return generator
 
 
 @choice.register(OpenAI)
--- a/outlines/generate/json.py
+++ b/outlines/generate/json.py
@@ -4,7 +4,7 @@ from typing import Callable, Optional, Union
 
 from pydantic import BaseModel
 
-from outlines.fsm.json_schema import build_regex_from_object, get_schema_from_signature
+from outlines.fsm.json_schema import build_regex_from_schema, get_schema_from_signature
 from outlines.generate.api import SequenceGenerator
 from outlines.models import OpenAI
 from outlines.samplers import Sampler, multinomial
@@ -45,17 +45,17 @@ def json(
     """
     if isinstance(schema_object, type(BaseModel)):
         schema = pyjson.dumps(schema_object.model_json_schema())
-        regex_str = build_regex_from_object(schema, whitespace_pattern)
+        regex_str = build_regex_from_schema(schema, whitespace_pattern)
         generator = regex(model, regex_str, sampler)
         generator.format_sequence = lambda x: schema_object.parse_raw(x)
     elif callable(schema_object):
         schema = pyjson.dumps(get_schema_from_signature(schema_object))
-        regex_str = build_regex_from_object(schema, whitespace_pattern)
+        regex_str = build_regex_from_schema(schema, whitespace_pattern)
         generator = regex(model, regex_str, sampler)
         generator.format_sequence = lambda x: pyjson.loads(x)
     elif isinstance(schema_object, str):
         schema = schema_object
-        regex_str = build_regex_from_object(schema, whitespace_pattern)
+        regex_str = build_regex_from_schema(schema, whitespace_pattern)
         generator = regex(model, regex_str, sampler)
         generator.format_sequence = lambda x: pyjson.loads(x)
     else:
--- a/outlines/generate/regex.py
+++ b/outlines/generate/regex.py
@@ -3,6 +3,7 @@ from functools import singledispatch
 from outlines.fsm.fsm import RegexFSM
 from outlines.generate.api import SequenceGenerator
 from outlines.models import OpenAI
+from outlines.models.llamacpp import LlamaCpp, RegexLogitsProcessor
 from outlines.samplers import Sampler, multinomial
 
 
@@ -35,8 +36,30 @@ def regex(model, regex_str: str, sampler: Sampler = multinomial()):
     return generator
 
 
+@regex.register(LlamaCpp)
+def regex_llamacpp(
+    model: LlamaCpp,
+    regex_str: str,
+    sampler: Sampler = multinomial(),
+):
+    if not isinstance(sampler, multinomial):
+        raise NotImplementedError(
+            r"The llama.cpp integration does not currently support any other sampling algorithm "
+            + "than the multinomial sampler."
+        )
+
+    logits_processor = RegexLogitsProcessor(regex_str, model.tokenizer)
+    model.logits_processor = logits_processor
+
+    return model
+
+
 @regex.register(OpenAI)
-def regex_openai(model, regex_str: str, sampler: Sampler = multinomial()):
+def regex_openai(
+    model: OpenAI,
+    regex_str: str,
+    sampler: Sampler = multinomial(),
+):
     raise NotImplementedError(
         "Cannot use regex-structured generation with an OpenAI model"
         + "due to the limitations of the OpenAI API."
--- a/outlines/generate/text.py
+++ b/outlines/generate/text.py
@@ -2,7 +2,7 @@ from functools import singledispatch
 
 from outlines.fsm.fsm import StopAtEosFSM
 from outlines.generate import SequenceGenerator
-from outlines.models import OpenAI
+from outlines.models import LlamaCpp, OpenAI
 from outlines.samplers import Sampler, multinomial
 
 
@@ -36,12 +36,23 @@ def text(model, sampler: Sampler = multinomial()) -> SequenceGenerator:
     return generator
 
 
+@text.register(LlamaCpp)
+def text_llamacpp(model: LlamaCpp, sampler: Sampler = multinomial()):
+    if not isinstance(sampler, multinomial):
+        raise NotImplementedError(
+            r"The llama.cpp API does not support any other sampling algorithm "
+            + "than the multinomial sampler."
+        )
+
+    return model
+
+
 @text.register(OpenAI)
 def text_openai(model: OpenAI, sampler: Sampler = multinomial()) -> OpenAI:
     if not isinstance(sampler, multinomial):
         raise NotImplementedError(
             r"The OpenAI API does not support any other sampling algorithm "
-            + "that the multinomial sampler."
+            + "than the multinomial sampler."
         )
 
     return model
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -1,304 +1,184 @@
-import ctypes
-from typing import List, Optional, Tuple, Union
+import math
+from typing import List, Optional, Union
 
 import numpy as np
 import torch
 from numpy.typing import NDArray
 
-from outlines.models.tokenizer import Tokenizer
+from outlines.fsm.fsm import CFGFSM, FSM, FSMState, RegexFSM
 
 
 class LlamaCpp:
     """Represents a `llama_cpp` model."""
 
     def __init__(
-        self, llama_instance, model, tokenizer, device, context_params, **kwargs
+        self, model_path, logits_processor: Optional["LogitsProcessor"] = None, **kwargs
     ):
-        self.device = device
-        self.llama_instance = llama_instance
-        self.tokenizer = tokenizer
+        from llama_cpp import Llama
 
-        # Note: the concept of padding does not exist in llama.cpp as a batched sequence is just
-        # a flat array of tokens that can be assigned to one or more sequences.
-        # To make it compatible with the transformers inspired tokenizer interface
-        # we need a padding token to homogenize to token_ids tensor.
-        self.pad_token_id = -1
-
-        self.n_past = 0
-        self.n_vocab = kwargs.pop("n_vocab")
-
-        self.ctx = llama_instance.llama_new_context_with_model(model, context_params)
-
-    def forward(self, input_ids: torch.LongTensor, *_):
-        """Compute a forward pass through the llama_cpp model."""
-        if input_ids.ndim == 2:
-            seq_tensor = input_ids[:, self.n_past :]
-        elif input_ids.ndim == 1:
-            seq_tensor = input_ids.view(1, -1)[:, self.n_past :]
-        else:
-            raise Exception("Only one and two dimensional inputs allowed.")
-
-        tokens_total = torch.numel(seq_tensor[seq_tensor != self.pad_token_id])
-        batch = self.llama_instance.llama_batch_init(tokens_total, 0, 1)
-
-        seq_token_ids = []
-        for seq_idx, seq in enumerate(seq_tensor):
-            for token_pos, token_id in enumerate(seq):
-                if token_id == self.pad_token_id:
-                    break
-                batch.token[batch.n_tokens] = token_id.item()
-                batch.pos[batch.n_tokens] = token_pos
-                batch.seq_id[batch.n_tokens][0] = seq_idx
-                batch.n_seq_id[batch.n_tokens] = 1
-                batch.logits[batch.n_tokens] = False
-
-                batch.n_tokens += 1
-                self.n_past += 1
-
-            batch.logits[batch.n_tokens - 1] = True
-            seq_token_ids.append(batch.n_tokens - 1)
-
-        if self.llama_instance.llama_decode(self.ctx, batch) != 0:
-            print("Error decoding")
-
-        all_logits = []
-        for seq_token in seq_token_ids:
-            logits = self.llama_instance.llama_get_logits_ith(self.ctx, seq_token)
-            logits_list = (ctypes.c_float * self.n_vocab)(
-                *[logits[token_id] for token_id in range(self.n_vocab)]
-            )
-            logits_tensor = torch.tensor(logits_list)
-            all_logits.append(logits_tensor)
-
-        self.llama_instance.llama_batch_free(batch)
-
-        stacked_logits = torch.stack(all_logits)
-        return stacked_logits, None
+        self.logits_processor = logits_processor
+        self.model = Llama(model_path, **kwargs)
+        self.tokenizer = LlamaCppTokenizer(self)
 
     def __call__(
         self,
-        input_ids: torch.LongTensor,
-        attention_mask: torch.LongTensor,
-        past_key_values: Optional[Tuple] = None,
-    ) -> torch.FloatTensor:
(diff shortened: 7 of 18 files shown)


examples/llamacpp_processor.py
```diff
@@ -0,0 +1,68 @@
...
+    llama = Llama("./phi-2.Q4_K_M.gguf")
+
+    prompt = b"Instruct: You are a leading role play gamer. You have seen thousands of different characters and their attributes.\nPlease return a JSON object with common attributes of an RPG character. Give me a character description\nOutput:"
+
+    logits_processor = JSONLogitsProcessor(Character, llama)
+    stopping_criteria_list = StoppingCriteriaList([EosCriteria(llama.token_eos())])
+
+    json_str = ""
+    tokens = llama.tokenize(prompt)
+    for token in llama.generate(
```
rlouf, line 54: Does `llama.generate` return an iterator or are you just iterating over the results?

dtiarks: It's a generator and we should probably use `llama.create_completion` here...

rlouf:
I think we should use a hybrid approach where we subclass `LLM` to be able to get the logits (useful for debugging), choose the sampling method and give this the same interface as the loop that is used for `transformers` models. What do you think?

dtiarks:
Sounds good. I'm going to move the logit processor into the integration file.

outlines/models/llamacpp.py
```diff
@@ -1,131 +1,170 @@
...
+class JSONLogitsProcessor(RegexLogitsProcessor):
+    def __init__(self, schema, llm):
+        """Compile the FSM that drives the JSON-guided generation.
+
+        Parameters
+        ----------
+        schema
+            A JSON schema that encodes the structure we want the model to generate
+        llm
+            An instance of `vllm.LLM`
```
lapp0, line 68: `An instance of `LlamaCpp`

lapp0 (commented):
IMHO, we should have an abstract implementation of `RegexLogitsProcessor`, `CFGLogitsProcessor`, and `JSONLogitsProcessor` which handles the state management logic, then subclass for specific interfacing with inference library.

dtiarks:
Jip that's a good idea...

outlines/models/llamacpp.py
```diff
@@ -1,131 +1,170 @@
...
+            super().eval(input_ids[i, self.n_tokens :])
+
+            logits = super().eval_logits
+            logits_tensor = torch.FloatTensor(list(logits))
+            all_logits.append(logits_tensor[0, :])
 
-        return next_token_logits, kv_cache
+        return torch.stack(all_logits), None
+
+    def sample(
```
rlouf, line 104: We don't need to subclass `sample` until we make `Llama` support other sampling algorithms)

outlines/models/llamacpp.py
```diff
@@ -1,131 +1,170 @@
...
         self,
         input_ids: torch.LongTensor,
         attention_mask: torch.LongTensor,
         past_key_values: Optional[Tuple] = None,
     ) -> torch.FloatTensor:
-        logits, kv_cache = self.forward(input_ids, attention_mask, past_key_values)
-        next_token_logits = logits
+        all_logits = []
+
+        for i in range(input_ids.size(0)):
```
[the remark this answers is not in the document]
dtiarks: So __call__ is not used elsewhere in outlines?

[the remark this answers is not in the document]
rlouf: For this PR we will do something similar to the new OpenAI integration in #574. The Llama model will be connected to the API by dispatching `generate.text`, `generate.regex`, etc based on the model type.

This should give us a simple, hopefully bug-free, integration. Later on we can build on this and integrate more deeply with the generation procedures in Outlines.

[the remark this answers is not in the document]
dtiarks: sounds good

rlouf:
I updated the code in PR to give you a better idea of the interface I'm thinking about. There are still several things to think about / do:

- We need to add a `stream` method to the `LlamaCpp` class. Signature should be the same as `SequenceGenerator.stream`
- We can use `Llama.__call__` method, and make sure that the signature is identical to `SequenceGenerator.__call__`
- There is potential for simplification: `JSONLogitsProcessor` is redundant with `RegexLogitsProcessor`
- For JSON-structured generation we need to format the output so we have the same interface as the one we have with `transformers` models. We will need to add a `format_sequence` method to `LlamaCpp` and apply it to the output of `__call__`. This way we could also remove the need for the `json_llamacpp` function.
- If llama.cpp provides tiny models just like [`transformers` does](https://huggingface.co/hf-internal-testing) we should add integration tests.

dtiarks:
- [ ] Finalize integration tests
- [ ] Clarify if we need multi prompt streaming in the first iteration (@rlouf )

rlouf:
We can forget about multi-prompt streaming in the first iteration, raise an informative `NotImplementedError` when a user asks for it,  and add an issue for this.

outlines/generate/text.py
```diff
@@ -58,6 +53,23 @@ def text(
...
+def text_llamacpp(
+    model: LlamaCpp,
+    max_tokens: Optional[int] = None,
+    stop_at: Optional[Union[List[str], str]] = None,
+    *,
+    sampler: Sampler = multinomial,
+):
+    if not sampler == multinomial:
+        raise NotImplementedError(
+            r"The OpenAI API does not support any other sampling algorithm "
```
rlouf, line 66: llama.cpp

dtiarks:
This is essentially ready for reviewing. I talked with @rlouf in private about it assuming he would review it.

rlouf:
I am going to as soon as I have time!

examples/llamacpp_example.py
```diff
@@ -31,7 +31,7 @@ class Character(BaseModel):
 
 if __name__ == "__main__":
     # Download model from https://huggingface.co/TheBloke/phi-2-GGUF
-    model = outlines.models.llamacpp("./phi-2.Q3_K_M.gguf", device="cpu")
+    model = outlines.models.llamacpp("./phi-2.Q4_K_M.gguf", device="cpu")
```
rlouf, line 34: Is the `device` argument here doing anything?

outlines/generate/json.py
```diff
@@ -68,6 +69,35 @@ def json(
...
+            f"Cannot parse schema {schema_object}. The schema must be either "
+            + "a Pydantic object, a function or a string that contains the JSON "
+            + "Schema specification"
+        )
+
+    # TODO: format the output
+    # We should be able to use the same interface as transformers and make this
+    # function redundant by adding a `format_sequence` method to `LlamaCpp`
+
+    return regex(model, regex_str, sampler)
```
rlouf, line 98: Yes, we should add a `format_sequence` method to `LlamaCpp`.

outlines/generate/regex.py
```diff
@@ -35,8 +37,33 @@ def regex(model, regex_str: str, sampler: Sampler = multinomial()):
...
+        raise NotImplementedError(
+            r"The llama.cpp integration does not currently support any other sampling algorithm "
+            + "that the multinomial sampler."
+        )
+
+    logits_processor = RegexLogitsProcessor(regex_str, model.tokenizer)
+    model.logits_processor = logits_processor
+
+    fsm = RegexFSM(regex_str, model.tokenizer)
+    model.fsm = fsm
```
rlouf, line 56: Why having to pass a `RegexFSM` and a `LogitsProcessor`?

pyproject.toml
```diff
@@ -56,7 +56,7 @@ test = [
     "beartype<0.16.0",
     "datasets",
     "responses",
-    "llama-cpp-python",
+    "llama-cpp-python==0.2.37",
```
rlouf, line 59: Does this really need to be pinned?

rlouf:
I rebased your branch on `main` and fixed the merge conflicts and made a few fixes. However the integration tests are no longer passing, I'm not sure if that's because of something I did. Could you take a look?

examples/llamacpp_example.py
dtiarks: Nope. This is a left-over from the older implementation. Will remove it.

pyproject.toml
dtiarks: 0.2.38 seemed to have a bug. Older versions worked well. There was also a bug ticket. I will check if it's solved.

dtiarks:
Jip I will check the tests.

outlines/generate/json.py
dtiarks: It is implemented but probably got mixed up when fixing conflicts.

examples/llamacpp_example.py
rlouf: Should we still let the user specify?

dtiarks: Ok for me

pyproject.toml
dtiarks: So this could be something more fundamental and has nothing to do with the implementation. Maybe something model specific. llama-cpp-python >= 0.2.38 is breaking the generation for e-mail adresses...not sure what this is. Maybe we could just pin it for this PR and investigate further in a new PR.

rlouf: Sounds good to me, we need to open an issue before merging the PR

tests/fsm/test_json_schema.py
```diff
@@ -552,10 +552,10 @@ class MockModel(BaseModel):
 
     # assert any ws pattern can be used
     if whitespace_pattern == "abc":
-        build_regex_from_object(MockModel, whitespace_pattern)
+        build_regex_from_schema(MockModel, whitespace_pattern)
```
dtiarks, line 555: This seems to be broken independently of the current PR. `build_regex_from_schema` expects the schema and not the object. I fixed it.

pyproject.toml
dtiarks: https://github.com/outlines-dev/outlines/issues/638

outlines/generate/regex.py
dtiarks: Was this comment on an older version?

outlines/generate/cfg.py
```diff
@@ -31,6 +33,24 @@ def cfg(model, cfg_str: str, sampler: Sampler = multinomial()) -> SequenceGenera
...
+@cfg.register(LlamaCpp)
+def cfg_llamacpp(
+    model: LlamaCpp,
+    cfg_str: str,
+    sampler: Sampler = multinomial(),
+):
+    if not isinstance(sampler, multinomial):
+        raise NotImplementedError(
+            r"The llama.cpp integration does not currently support any other sampling algorithm "
+            + "that the multinomial sampler."
```
lapp0, line 45: nit: "than"

outlines/generate/choice.py
```diff
@@ -13,7 +13,11 @@ def choice(
     model, choices: List[str], sampler: Sampler = multinomial()
 ) -> SequenceGenerator:
     regex_str = r"(" + r"|".join(choices) + r")"
-    return regex(model, regex_str, sampler)
+
+    generator = regex(model, regex_str, sampler)
+    generator.format_sequence = lambda x: x
```
lapp0, line 18: This line is unnecessary, it's what the default `format_sequence` already does.

outlines/generate/processors.py
```diff
@@ -0,0 +1,78 @@
+import math
```
lapp0, line 1: Great job creating generic logits processors! This will help a lot with applying logits processors to other inference engines!

lapp0 (commented):
Smoke tested with the code below adapted from https://github.com/outlines-dev/outlines/issues/611

**Edit: It works on `llama-cpp-python==0.2.37` as you enforce in `pyproject.toml**

```
import outlines
import json
import os
import sys

def raw_outlines_test():
    """
    Just a test

    """
    schema = """{"properties": {"name": {"title": "Name", "type": "string",
    "maxLength": 100}},
    "required": ["name"], "title": "person", "type": "object"}
    """
    import sys
    json.loads(schema)
    sys.stderr.write("XXXXXXXXXXX  JSON schema successfully created  XXXXXXXXXXXXXX\n")
    llm = outlines.models.llamacpp("./mistral-7b-instruct-v0.2.Q5_K_M.gguf",
        model_kwargs = {"n_gpu_layers": 15000, "n_ctx": 5 * 1024})
    sys.stderr.write("XXXXXXXXXX Model loaded  XXXXXXXXXXXXXXXXXXXXXX\n")
    generator = outlines.generate.json(llm, schema)
    sys.stderr.write("XXXXXXXXXXXX Generator created XXXXXXXXXXXXXXX\n")
    text = generator("What is the name of Michael Jackson?")
    sys.stderr.write("XXXXXXXXXXXXX Result: XXXXXXXXXXXXXXXX\n")
    sys.stderr.write(str(text))

raw_outlines_test()
```

It resolves their

`Failed to convert token (b'\x80'): 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte`

error

however it doesn't enforce JSON generation.

```
  File "/root/outlines/run.py", line 27, in <module>                                                                                                                                    
    raw_outlines_test()                                                                                                                                                                 
  File "/root/outlines/run.py", line 23, in raw_outlines_test                                                                                                                           
    text = generator("What is the name of Michael Jackson?")                                                                                                                            
  File "/root/outlines/outlines/models/llamacpp.py", line 64, in __call__                                                                                                               
    formatted = [self.format_sequence(sequence) for sequence in results]                                                                                                                
  File "/root/outlines/outlines/models/llamacpp.py", line 64, in <listcomp>                                                                                                             
    formatted = [self.format_sequence(sequence) for sequence in results]                                                                                                                
  File "/root/outlines/outlines/generate/json.py", line 60, in <lambda>                                                                                                                 
    generator.format_sequence = lambda x: pyjson.loads(x)                                                                                                                               
  File "/usr/lib/python3.10/json/__init__.py", line 346, in loads                                                                                                                       
    return _default_decoder.decode(s)                                                                                                                                                   
  File "/usr/lib/python3.10/json/decoder.py", line 337, in decode                                                                                                                       
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.10/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
```

outlines/generate/choice.py
dtiarks: That's what I thought as well. However monkey patching has the downside that when working in some context the patched method persists unless you reinitialize. So without the trivial lambda the integration tests will fail because it uses the format_sequence method from a prior test.

dtiarks:
> Smoke tested with the code below adapted from #611
> 
> ```
> import outlines
> import json
> import os
> import sys
> 
> def raw_outlines_test():
>     """
>     Just a test
> 
>     """
>     schema = """{"properties": {"name": {"title": "Name", "type": "string",
>     "maxLength": 100}},
>     "required": ["name"], "title": "person", "type": "object"}
>     """
>     import sys
>     json.loads(schema)
>     sys.stderr.write("XXXXXXXXXXX  JSON schema successfully created  XXXXXXXXXXXXXX\n")
>     llm = outlines.models.llamacpp("./mistral-7b-instruct-v0.2.Q5_K_M.gguf",
>         model_kwargs = {"n_gpu_layers": 15000, "n_ctx": 5 * 1024})
>     sys.stderr.write("XXXXXXXXXX Model loaded  XXXXXXXXXXXXXXXXXXXXXX\n")
>     generator = outlines.generate.json(llm, schema)
>     sys.stderr.write("XXXXXXXXXXXX Generator created XXXXXXXXXXXXXXX\n")
>     text = generator("What is the name of Michael Jackson?")
>     sys.stderr.write("XXXXXXXXXXXXX Result: XXXXXXXXXXXXXXXX\n")
>     sys.stderr.write(str(text))
> 
> raw_outlines_test()
> ```
> 
> It resolves their
> 
> `Failed to convert token (b'\x80'): 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte`
> 
> error
> 
> however it doesn't enforce JSON generation.
> 
> ```
> Traceback (most recent call last):                                                                                                                                                      
>   File "/root/outlines/run.py", line 27, in <module>                                                                                                                                    
>     raw_outlines_test()                                                                                                                                                                 
>   File "/root/outlines/run.py", line 23, in raw_outlines_test                                                                                                                           
>     text = generator("What is the name of Michael Jackson?")                                                                                                                            
>   File "/root/outlines/outlines/models/llamacpp.py", line 64, in __call__                                                                                                               
>     formatted = [self.format_sequence(sequence) for sequence in results]                                                                                                                
>   File "/root/outlines/outlines/models/llamacpp.py", line 64, in <listcomp>                                                                                                             
>     formatted = [self.format_sequence(sequence) for sequence in results]                                                                                                                
>   File "/root/outlines/outlines/generate/json.py", line 60, in <lambda>                                                                                                                 
>     generator.format_sequence = lambda x: pyjson.loads(x)                                                                                                                               
>   File "/usr/lib/python3.10/json/__init__.py", line 346, in loads                                                                                                                       
>     return _default_decoder.decode(s)                                                                                                                                                   
>   File "/usr/lib/python3.10/json/decoder.py", line 337, in decode                                                                                                                       
>     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
>   File "/usr/lib/python3.10/json/decoder.py", line 353, in raw_decode
>     obj, end = self.scan_once(s, idx)
> json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
> ```

I will look into this. One thing to note is that outlines currently does not exactly guarantee JSON conformity (last time I checked). While testing I saw some edge cases where the resulting JSON string could not be parsed due to wierd key names.

rlouf:
> I will look into this. One thing to note is that outlines currently does not exactly guarantee JSON conformity (last time I checked). While testing I saw some edge cases where the resulting JSON string could not be parsed due to wierd key names.

Weird key names means there's an issue with the way tokens are handled.

lapp0:
@dtiarks no that was my mistake, I didn't apply your version pin. It works perfectly with `llama-cpp-python==0.2.37`

outlines/generate/processors.py
rlouf: I realized llama.cpp and vLLM do not both pass the prompt token ids in the logit biasing function. I am thus moving these classes from `outlines.generate`  to `outlines.models.llamacpp`

examples/llamacpp_example.py
```diff
@@ -30,8 +30,7 @@ class Character(BaseModel):
 
 
 if __name__ == "__main__":
-    # Download model from https://huggingface.co/TheBloke/phi-2-GGUF
-    model = outlines.models.llamacpp("./phi-2.Q3_K_M.gguf", device="cpu")
+    model = outlines.models.llamacpp("./my_model.gguf")
```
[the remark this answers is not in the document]
rlouf: Even better in the documentation

rlouf (changes requested):
I am currently simplifying a few things in the code. Please wait until I'm done to make changes.

rlouf:
I simplified the code in this PR, among other things removing all the tokenizer code that is not needed anymore. Since the code in `outlines.generator.processors` does not generalize I chose to leave it out of the model.

outlines/models/llamacpp.py
```diff
@@ -1,304 +1,235 @@
...
+            An instance of `Tokenizer`
+        fsm
+            An instance of `FSM`
 
-                seq.append(token_piece)
+        """
+        self.tokenizer = tokenizer
+        self.fsm_state = FSMState(0)
+        self.fsm: FSM = fsm
+        self.is_first_token = True
```
lapp0, line 184: `self.is_first_token` appears to be unused

outlines/generate/cfg.py
dtiarks: We should probably remove this alltogether because llama.cpp does not use the provided sampler at all.

dtiarks:
> I simplified the code in this PR, among other things removing all the tokenizer code that is not needed anymore. Since the code in `outlines.generator.processors` does not generalize I chose to leave it out of the model.

All the processor I looked at so far (transformers, vLLM, TensorRT-LLM etc.) have a similar interface but not exactly the same. So it makes sense to have a model specific implementation for each.

outlines/generate/cfg.py
rlouf: We can pass kwargs related to sampling to `LlamaCpp` based on the sampler passed by the user.  We’re trying to build a unified interface where models are interchangeable 🙂

dtiarks: Do you want this to be handled in this PR? The conversion logic does not seem to be a small feat IMHO.

examples/llamacpp_example.py
dtiarks: I just realized that I misunderstood the comment (don't read comments late at nigh :smile:). The device handling can be controlled via kwargs. E.g. `n_gpu_layers` and `main_gpu`. If this has an effect is based on if you built llama-cpp-python with cublas (or metal etc.) support.

outlines/generate/regex.py
```diff
@@ -3,6 +3,7 @@
 from outlines.fsm.fsm import RegexFSM
 from outlines.generate.api import SequenceGenerator
 from outlines.models import OpenAI
+from outlines.models.llamacpp import LlamaCpp, RegexLogitsProcessor
```
lapp0, line 6: We should import within `def regex_llamacpp`, otherwise using `outlines.generate.regex` and `outlines.generate.json` will require installing `llama_cpp_python`

rlouf: We register on the `LlamaCpp` type which is a subclass of `Llama`, so there's clearly a design issue here

rlouf:
I rebased the branch on `main` and fixed the merge conflicts commit-by-commit so there may be some errors popping up. Please **rebase** on `main` instead of adding a merge commit in the future to preserve the history.

We still need to move a few things around, as noted @lapp0 the current code makes it necessary to install `llama-cpp-python` regardless of whether we're using llama.cpp or not.

rlouf:
I found a workaround, can someone review before we merge?

outlines/models/llamacpp.py
```diff
@@ -1,304 +1,185 @@
...
+
+        if rng is None:
+            rng = torch.Generator(device="cpu")
+            rng.seed()
+
+        results = []
+        for prompt in prompts:
+            processors = []
+            if self.logits_processor is not None:
+                processors = [copy.copy(self.logits_processor)]
```
lapp0, line 45: Won't the logits processors share a `CFGFSM` if we shallow copy?

outlines/models/llamacpp.py
```diff
@@ -1,304 +1,185 @@
...
+                logits_processor=LogitsProcessorList(processors),
+                **model_kwargs,
+            )["choices"][0]["text"]
+            results.append(result)
+
+            self.model.reset()
+
+        formatted = [self.format_sequence(sequence) for sequence in results]
+
+        return formatted if len(formatted) > 1 else formatted[0]
```
lapp0, line 61: nit:

IMO, `List[prompt]` should always correspond to `List[generation]`. I think we should condition on the passed `prompts` type rather than the length of `formatted`.

outlines/models/llamacpp.py
```diff
@@ -1,304 +1,185 @@
...
-            self.model, self.llama_instance.llama_token(self.eos_token_id), buffer, size
-        )
+    def __call__(
+        self, input_ids: NDArray[np.int64], scores: NDArray[np.float32]
+    ) -> NDArray[np.float32]:
+        """Use the FSM to bias the logits before sampling the next token."""
 
-        token_piece = buffer[:n].decode("utf-8")
+        if self.is_first_token:
+            self.is_first_token = False
```
lapp0, line 142: Note: This works so long as we don't expose llama logits processors to the user. If the user uses the same logits processor twice, `self.is_first_token` not being reset will cause problems.

I don't see any docs or other ways in which the logits processor would be exposed, so I think it's fine for now.

pyproject.toml
lapp0: My fix for `llama-cpp-python` is in `v0.2.42`, we should pin accordingly.

outlines/models/llamacpp.py
rlouf: True, logits processors should have a `copy` method.

outlines/models/llamacpp.py
rlouf: Would using a token_ids -> state map solve this?

lapp0: I believe so but I haven't tested. This is working for now, so perhaps I should refactor so there is a base logits processor in a separate PR?

outlines/generate/regex.py
```diff
@@ -35,8 +36,30 @@ def regex(model, regex_str: str, sampler: Sampler = multinomial()):
...
+    sampler: Sampler = multinomial(),
+):
+    if not isinstance(sampler, multinomial):
+        raise NotImplementedError(
+            r"The llama.cpp integration does not currently support any other sampling algorithm "
+            + "than the multinomial sampler."
+        )
+
+    logits_processor = RegexLogitsProcessor(regex_str, model.tokenizer)
+    model.logits_processor = logits_processor
```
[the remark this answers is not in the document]
rlouf: But a new generation would bind a different logit processor?

rlouf:
@dtiarks You can take a last look

outlines/generate/regex.py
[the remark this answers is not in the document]
rlouf: I think it's fine for now given how `__call__` works, we may revisit this in the future though.

outlines/models/llamacpp.py
dtiarks: We should probably open an issue for this.

rlouf:
I added a `copy` method to `LogitsProcessor` that uses the FSMs' `copy` method. Fixes #672.

rlouf:
Merging, great job everyone!

PR #560, opened 2024-01-19
Added subclass of OpenAI class to support custom endpoints

shannonsands:
Adds a new class to support for providing a custom base_url to the OpenAI client, and a couple of small updates to remove the checks on the model via the "retrieve" api (since it's not always available), sets tiktoken to use gpt4 by default (with the option to override it), and enables passing temperature as a named parameter to the __call__method to override the config.

Usage is basically the same as the regular OpenAI model wrapper, with the addition of the base_url parameter. Eg:

```
hermes = CustomOpenAI(
        model_name="NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO",
        system_prompt=system_prompt,
        api_key="xxx",
        base_url='https://api.together.xyz',
 )

print(hermes(test_input))

```
The environment variables checked to set the API key was also updated, so as not to conflict with any OpenAI configurations in place, and sound a little more generic, along with a new env var `INFERENCE_BASE_URL` to act as an alternative to the constructor parameter.

### DIFF
--- a/outlines/models/custom_openai.py
+++ b/outlines/models/custom_openai.py
@@ -1,0 +1,162 @@
+"""Integration with custom OpenAI compatible APIs."""
+import functools
+import os
+from dataclasses import replace
+from typing import List, Optional, Union
+
+import numpy as np
+
+from outlines.models.openai import generate_chat, OpenAIConfig, OpenAI
+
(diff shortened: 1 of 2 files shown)


rlouf:
I think it would be nice eventually to have users pass an `OpenAI` Client themselves, which I think should simplify this code?

PR #567, opened 2024-01-22
Implement JSON schema field constraints

aschwa:
Add JSON schema field constraints addressing  #215 

So far this PR includes:
* date-time
* time
* date
* UUID

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -24,6 +24,18 @@ type_to_regex = {
     "null": NULL,
 }
 
+DATE_TIME = r"(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]{3})?(Z)?"
+DATE = r"(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1])"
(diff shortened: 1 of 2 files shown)


outlines/fsm/json_schema.py
```diff
@@ -24,6 +24,18 @@
     "null": NULL,
 }
 
+DATE_TIME = r"(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]{3})?(Z)?"
+DATE = r"(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1])"
+TIME = r"(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?"
+UUID = r"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
```
ash211, line 30: Should these constants have literal `"` wrapping them, like how on L13 `STRING` is wrapping `STRING_INNER` in a double-quote characters?

I think this may be causing the bug observed at https://github.com/predibase/lorax/issues/392 where it's producing invalid json like this
```
{"key": 2022-02-02}
```
instead of
```
{"key": "2022-02-02"}
```

rlouf: Yes! Happy to quickly review and merge a PR that fixes this :)

Issue #568, opened 2024-01-22
Transformers 4.37.0 Breaks Outlines


lapp0:
### Describe the issue as clearly as possible:

`transformers==4.37.0` is incompatible with outlines and results in test failure.

Summary:

```
tests/test_base.py ..............                                        [  8%]
tests/test_cache.py ....                                                 [ 10%]
tests/test_function.py F.....                                            [ 14%]
tests/test_prompts.py .........                                          [ 19%]
tests/fsm/test_fsm.py ......x                                            [ 23%]
tests/fsm/test_json_schema.py .......................................... [ 48%]
                                                                         [ 48%]
tests/fsm/test_parsing.py ..                                             [ 49%]
tests/fsm/test_regex.py .....ss                                          [ 53%]
tests/fsm/test_types.py ......                                           [ 57%]
tests/generate/test_generator.py ..........................              [ 72%]
tests/generate/test_integration_transfomers.py FFFxxFFFFFFFFFFFFFFFFFxFF [ 87%]
.F                                                                       [ 88%]
tests/generate/test_samplers.py ..                                       [ 89%]
tests/models/test_llama_cpp.py ..                                        [ 91%]
tests/models/test_openai.py ..........                                   [ 97%]
tests/models/test_tokenizer.py .                                         [ 97%]
tests/models/test_transformers.py ..F.                                   [100%]
```

Example:

```
tests/generate/test_integration_transfomers.py:20: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
outlines/models/transformers.py:220: in transformers
    model = AutoModelForCausalLM.from_pretrained(model_name, **model_kwargs)
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/transformers/models/auto/auto_factory.py:566: in from_pretrained
    return model_class.from_pretrained(
/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/transformers/modeling_utils.py:3527: in from_pretrained
    state_dict = load_state_dict(resolved_archive_file)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #569, opened 2024-01-22
Pin transformers version to 4.36.2

lapp0:
Temporarily change to prevent failures described in https://github.com/outlines-dev/outlines/issues/568

Won't apply "fix" label because we don't know the root source of the issue (i.e. whether it's a result of incompatability or if it's a bad version of `transformers`)


rlouf:
Thank you for opening a PR for this!

Issue #583, opened 2024-01-25
All examples fail using ExllamaV2


dnhkng:
### Describe the issue as clearly as possible:

Generating using the examples given on the front page of the repo all generate the same error:

`RuntimeError: Index put requires the source and destination dtypes match, got Float for the destination and Half for the source.`

### Steps/code to reproduce the bug:

```python
import outlines

model = outlines.models.exl2(model_name = "models/TinyLlama-1.1B-Chat-v1.0-5.0bpw-h6-exl2", model_kwargs={'num_experts_per_token':1}, device="cpu")

prompt = """You are a sentiment-labelling assistant.
Is the following review positive or negative?

Review: This restaurant is just awesome!
"""

generator = outlines.generate.choice(model, ["Positive", "Negative"])
answer = generator(prompt)
```

### Expected result:

```shell
Either "Positive", "Negative"
```

### Error message:

```shell
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[34], line 10
      3 prompt = """You are a sentiment-labelling assistant.
      4 Is the following review positive or negative?
      5 
      6 Review: This restaurant is just awesome!
      7 """
      9 generator = outlines.generate.choice(model, ["Positive", "Negative"])
---> 10 answer = generator(prompt)

File ~/miniforge3/envs/exllama/lib/python3.10/site-packages/outlines/generate/api.py:213, in SequenceGenerator.__call__(self, prompts, max_tokens, stop_at, rng, kv_cache)
    211 while True:
    212     try:
--> 213         last_state = next(states)
    214         if max_tokens or stop_sequences:
    215             generated_token_ids = self.get_generated_token_ids(
    216                 init_state, prompts, last_state
    217             )

File ~/miniforge3/envs/exllama/lib/python3.10/site-packages/outlines/generate/generator.py:84, in sequence_generator(token_generator, fsms, init_state, fsm_states, rng)
     81 while True:
     82     allowed_tokens = get_allowed_tokens(fsms, fsm_states)
---> 84     next_token_ids, kv_cache, logits, _ = token_generator(
     85         token_ids,
     86         attention_masks,
     87         kv_cache,
     88         rng=rng,
     89         allowed_tokens=allowed_tokens,
     90     )
     92     token_ids = update_token_ids(token_ids, next_token_ids)
     93     attention_masks = expand_attention_masks(attention_masks)

[...]

PR #584, opened 2024-01-25
Bugfix:  Fixes dtype issue described in #583

dnhkng:
Exllama generates logits in torch Half-dtype, but Outlines requires the Float-dtype. 

This small change converts the logits to the required dtype (whatever that might be), solving issue #583.

Tested with Exllama on the example code on the github front page, and #583 is resolved.

### DIFF
--- a/outlines/generate/generator.py
+++ b/outlines/generate/generator.py
@@ -282,7 +282,7 @@ def bias_logits(logits: torch.Tensor, allowed_token_ids: List) -> torch.Tensor:
     A view of the original logits tensor where some values are masked.
 
     """
-    biased_logits = torch.full(logits.shape, -math.inf, device=logits.device)
+    biased_logits = torch.full_like(logits, -math.inf, device=logits.device)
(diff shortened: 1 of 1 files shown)


outlines/generate/generator.py
```diff
@@ -284,5 +284,5 @@ def bias_logits(logits: torch.Tensor, allowed_token_ids: List) -> torch.Tensor:
     """
     biased_logits = torch.full(logits.shape, -math.inf, device=logits.device)
```
rlouf, line 285: Thank you for opening a PR! Could we create `biased_logits` with `dtype=logits.dtype` instead? Or better, use [torch.full_like](https://pytorch.org/docs/stable/generated/torch.full_like.html)

dnhkng: Ahh ok, if bias_logits can stay in half format, than this is a nicer solution. I wasn't sure what happens upstream.

rlouf:
Thank you so much for fixing the bug!

Issue #563, opened 2024-01-20
Make users initialize and pass the OpenAI client to `OpenAI`


rlouf:
We currently have a lot of duplicated code between `OpenAI`, `AureOpenAI` and `OpenAICompatibleAPI`. A cleaner solution would be to let users initialize the OpenAI client themselves and initialize `OpenAI` with it.

rlouf:
This would also help with testing. `OpenAI` currently makes a call to retrieve the name of models in the `__init__` method which prevents us from testing the class and the API.

RobinPicard:
I'm working on this issue and I find the line `openai = OpenAI` quite inconvenient at 2 levels:
- when writing parts of the model as it needs to import `openai` in some methods
- when using the model as I also need to import `openai` to instantiate the OpenAI client that will be given as an argument to the outlines model

So, I was wondering whether we can remove it and just provide an `OpenAI` object and, if not, whether we can rename it to avoid conflict.

PR #593, opened 2024-01-27
Modify openai model to receive openai client as argument during initialization

RobinPicard:
PR to solve issue #563 

I have an issue with testing though. The fact that there's an object named `openai` in the `openai` module creates an import conflict when trying to use `with patch('outlines.models.openai.generate_chat') as mocked_generate_chat`. I see 2 solutions:
- Rename/remove the openai object (see comment in the linked issue)
- Use something else than `unittest.mock.patch` to mock functions in that module. Maybe there's another way with the `unittest` package I'm unaware of. Otherwise I've used the package [expectise](https://github.com/tcassou/expectise) in the past and find it quite convenient, but that implies adding another dependency to outlines


rlouf:
Can't we remove `AzureOpenAI` and `OpenAICompatibleAPI` if we pass the client?

RobinPicard:
> Can't we remove `AzureOpenAI` and `OpenAICompatibleAPI` if we pass the client?

There are small differences:
- `AzureOpenAI` currently requires to have both a `deployment_name` and a `model_name`. The `deployment_name` is actually the equivalent of the `model_name` for the OpenAI client (it's what ends up in the OpenAIConfig) while the `model_name` is used for the `tokenizer` property
- OpenAICompatibleAPI currently requires an `encoding` argument used for the `tokenizer` property and we remove the check on the `model_name` for it

I think we could fit all of them in a single class with some optional parameters and conditions based on them in the initialization. I'm not sure it would be simpler to use though

RobinPicard:
I've merged the 3 into a single model/class, let me know if you think it's better like that

lapp0:
Please add `openai` to `[project.optional-dependencies]`

lapp0:
Please rebase, `main` has a fix for your test failure.

RobinPicard:
> Please rebase, `main` has a fix for your test failure.

Done, but I still have the issue of the conflict between the name of the module and the name of the `openai` object for the mocking of the `generate_chat` function in the tests

lapp0:
Ya the import structure is tricky, we might consider doing `import openai as _openai_module` in `outlines/models/__init__.py` to accommodate patching.

rlouf:
Looks good! I can take a look at the import issue. We will also need to update the documentation, this is a big interface change.

outlines/models/openai.py
```diff
@@ -76,88 +75,104 @@ class OpenAI:
...
+            import openai
+
+            client = openai.OpenAI(api_key=api_key)
+            client.models.retrieve(model_name)
+        except openai.AuthenticationError:
+            raise ValueError("Incorrect API key provided")
+        except openai.NotFoundError:
+            raise ValueError(
+                "Invalid model_name. Check openai models list at https://platform.openai.com/docs/models"
+            )
```
rlouf, line 167: Isn't that going to fail with 3rd party providers?

outlines/models/openai.py
```diff
@@ -218,7 +242,7 @@ def tokenizer(self):
                 "The `tiktoken` library needs to be installed in order to choose `outlines.models.openai` with `is_in`"
             )
 
-        return tiktoken.encoding_for_model(self.config.model)
+        return tiktoken.encoding_for_model(self.model_name)
```
rlouf, line 245: This is also going to fail with model providers other than OpenAI, right?

outlines/models/openai.py
```diff
@@ -76,88 +75,104 @@ class OpenAI:
...
+        if deployment_name:
+            if config is not None:
+                self.config = replace(config, model=deployment_name)  # type: ignore
+            else:
+                self.config = OpenAIConfig(model=deployment_name)
+        else:
+            if config is not None:
+                self.config = replace(config, model=model_name)  # type: ignore
+            else:
+                self.config = OpenAIConfig(model=model_name)
```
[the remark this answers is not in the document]
RobinPicard: For Azure, the idea was that `deployment_name` is the [custom name](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints) of the model chosen by the user (what's used when calling the client) while `model_name` is the original openai model name that we need for the `tokenizer` method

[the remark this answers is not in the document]
RobinPicard: But it's quite confusing as I tried to retain too much of the previous interface. Maybe what we could do instead is that in all cases a `model_name` is provided (what's actually used when calling the client) and on top of that an `encoding` (or another name) must be given when using Azure or a 3rd-party with the name of a standard openai model for the `tokenizer`

RobinPicard:
I've made a few modifications based on your comments to simplify the model and have fewer elements specific to OpenAI vs. Azure vs. OpenAICompatible. Sorry for the back and forth

rlouf:
I thought a little more about this, and given the changes I made in #727 I suggest the following interface:

1. The function `openai` initializes an `AsyncOpenAI` client. It takes as arguments everything that pertains to the client: `model_name`, `api_key`, `max_retries`, `timeout`.
2. We can have another `azure_openai` alias that also returns an `AsyncOpenAI` client.
3. The `OpenAI` class is initialized with an `AsyncOpenAI` (or compatible) client and a tokenizer. 

There are a few details that need to be fleshed out, but I think this solves many issues, like the impossibility to use choice with 3-rd party APIs.

rlouf:
@RobinPicard I implemented what I suggested above and fixed the issue with mocking. Could you please review?

outlines/models/openai.py
```diff
@@ -468,3 +412,57 @@ def call(*args, **kwargs):
...
+    try:
+        import tiktoken
+        from openai import AzureAsyncOpenAI
+    except ImportError:
+        raise ImportError(
+            "The `openai` and `tiktoken` libraries needs to be installed in order to use Outlines' Azure OpenAI integration."
+        )
+
+    if config is not None:
+        config = replace(config, model=model_name)  # type: ignore
```
RobinPicard, line 459: In the `azure.py` file, it used to be `self.config = replace(config, model=deployment_name)`, so I think the model should be the deployment_name

outlines/models/openai.py
```diff
@@ -468,3 +412,57 @@ def call(*args, **kwargs):
             raise e
 
     return call
+
+
+def openai(
+    model_name: str,
+    api_key: Optional[str] = None,
+    system_prompt: Optional[str] = None,
```
RobinPicard, line 420: This parameter is not used in the function

RobinPicard (commented):
The interface makes sense to me!

Issue #512, opened 2024-01-08
vLLM OpenAI compatible server integration


jamestwhedbee:
Your vLLM integration works beautifully, I was wondering if you were looking to support their OpenAI-compatible server as well

brucethemoose:
+1, interested in this as well.

Issue #594, opened 2024-01-28
Expose OpenAI API with CFG


7flash:
I understand currently outlines already serves OpenAI API with vllm, but what I would like to be actually able utilize all the features of outlines while using external chat UI, perhaps it can be another message type "outline" which would translate to generate.cfg(model, message) ?

rlouf:
What's the external chat UI you would like to be able to use?

7flash:
> What's the external chat UI you would like to be able to use?

https://github.com/ztjhz/BetterChatGPT

It works with vllm if you choose "Use custom API endpoint"

Also it shows role dropdown for every message in chat, perhaps i thought it can be another type "outline" next to "assistant" "system" and "user"

davidsyoung:
+1 on this, having the OAI API exposed from VLLM would be very valuable. Even if it was just pulling through the same API that VLLM can expose in the serve example.

louis030195:
+1

```bash
curl http://127.0.0.1:8000/chat/completions \
    -d '{
        "model": "microsoft/phi-2",
        "messages": [
          {
            "role": "system",
            "content": "You are a helpful assistant."
          },
          {
            "role": "user",
            "content": "What is the capital of France?"
          }
        ],
        "schema": {"type": "string", "maxLength": 5}
        }'
        
curl http://127.0.0.1:8000/chat/completions \
    -d '{
        "model": "microsoft/phi-2",
        "messages": [
          {
            "role": "system",
            "content": "You are a helpful assistant."
          },
          {
            "role": "user",
            "content": "What is Pi? Give me the first 15 digits: "
          }
        ],
        "regex": "(-)?(0|[1-9][0-9]*)(\\.[0-9]+)?([eE][+-][0-9]+)?"
        }'
```

7flash:
Currently I have implemented this script based on #598

```
import argparse
import json
from http import HTTPStatus
from fastapi import FastAPI, Request, Response
from fastapi.exceptions import RequestValidationError
from fastapi.middleware.cors import CORSMiddleware
from fastapi.responses import JSONResponse, StreamingResponse
from pydantic import BaseModel, Field
from typing import Optional, Union, List, Dict
from outlines.models.openai import OpenAI, OpenAIConfig, openai
import outlines.models as models
import os
import uvicorn
import time
from datetime import datetime
import re

HOST = os.getenv("HOST") or "0.0.0.0"
PORT = os.getenv("PORT") or "8000"
[...]

PR #598, opened 2024-01-29
Add outlines vLLM OpenAI server

Soufiane-Ra:
To use with Langchain / OpenAI add extra_body={"schema": yourSchema} or extra_body={"regex": yourRegex} example:

```
llm = ChatOpenAI ( openai_api_key="EMPTY",
                   openai_api_base="http://localhost:8000/v1",
                   model_name=model_path,
                   extra_body={"schema": json_schema },
                 )
```

### DIFF
--- a/outlines/serve/openai_server.py
+++ b/outlines/serve/openai_server.py
@@ -1,0 +1,785 @@
+import argparse
+import asyncio
+import codecs
+import json
+import time
+from http import HTTPStatus
+from typing import AsyncGenerator, Dict, List, Optional, Tuple, Union
+
+from aioprometheus import MetricsMiddleware
+from aioprometheus.asgi.starlette import metrics
+import fastapi
+import uvicorn
+from fastapi import Request
+from fastapi.exceptions import RequestValidationError
+from fastapi.middleware.cors import CORSMiddleware
+from fastapi.responses import JSONResponse, StreamingResponse, Response
+
+from vllm.engine.arg_utils import AsyncEngineArgs
+from vllm.engine.async_llm_engine import AsyncLLMEngine
+from vllm.engine.metrics import add_global_metrics_labels
+from protocol import (
+    CompletionRequest, CompletionResponse, CompletionResponseChoice,
+    CompletionResponseStreamChoice, CompletionStreamResponse,
+    ChatCompletionRequest, ChatCompletionResponse,
+    ChatCompletionResponseChoice, ChatCompletionResponseStreamChoice,
+    ChatCompletionStreamResponse, ChatMessage, DeltaMessage, ErrorResponse,
+    LogProbs, ModelCard, ModelList, ModelPermission, UsageInfo)
+from vllm.logger import init_logger
+from vllm.outputs import RequestOutput
+from vllm.sampling_params import SamplingParams
+from vllm.transformers_utils.tokenizer import get_tokenizer
+from vllm.utils import random_uuid
+import vllm.model_executor.layers.sampler as sampler
+from outlines.serve.vllm import (
+    JSONLogitsProcessor,
+    RegexLogitsProcessor,
+    _patched_apply_logits_processors,
+)
+
+
+TIMEOUT_KEEP_ALIVE = 5  # seconds
+
+sampler._apply_logits_processors = _patched_apply_logits_processors
+
+logger = init_logger(__name__)
+served_model = None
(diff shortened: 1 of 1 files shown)


lapp0:
Thanks for the PR! I see two issues this would close

https://github.com/outlines-dev/outlines/issues/512

https://github.com/outlines-dev/outlines/issues/594

outlines/serve/openai_server.py
```diff
@@ -0,0 +1,785 @@
+import argparse
```
lapp0, line 1: Do we need a reference to vLLMs license since this is derived from

https://github.com/vllm-project/vllm/blob/b72af8f1eded6f5838be29eb6093ab0e0e0c240c/vllm/entrypoints/openai/api_server.py#L10

outlines/serve/openai_server.py
```diff
@@ -0,0 +1,785 @@
...
+import asyncio
+import codecs
+import json
+import time
+from http import HTTPStatus
+from typing import AsyncGenerator, Dict, List, Optional, Tuple, Union
+
+from aioprometheus import MetricsMiddleware
```
lapp0, line 9: Please ensure any new dependencies are added are part of `pyproject.toml`

lapp0 (commented):
Good job integrating vllms openai module with outlines!

Could you document usage and availability in `vllm.md` (or a new markdown file `openai_endpoint.md`)?

jamestwhedbee:
I'd expect this to extend OpenAI's existing response_format field to include the schema, rather than use an extra_body field. Thoughts?

rlouf:
Thank you for opening a PR, this is impressive and I will need a few days to review it.

7flash:
Wow looks amazing, so I understand now after I deploy outlines to my server, then I can just add another line into my litellm config pointing to outlines, and then it just works compatible with my existing UI?

docs/reference/vllm.md
```diff
@@ -50,6 +50,50 @@ curl http://127.0.0.1:8000/generate \
...
+        "model": "./Mistral-7B-Instruct-v0.2-GPTQ",
+        "prompt": "Hello!",
+        "schema": {"type": "integer"},
+        "max_tokens": 5
+        }'
+```
+To use with Langchain or with OpenAI add the argument `extra_body`:
+
+```bash
+llm = ChatOpenAI( openai_api_key = "EMPTY",
```
rlouf, line 90: You would need to add an import here to make it clear that `ChatOpenAI` is not a class implemented in Outlines.

outlines/serve/openai_server.py
rlouf: Wouldn't they be installed when installing vLLM?

lapp0: Yes, I should have pulled before comparing to me local `openai_server.py`...

rlouf:
@Soufiane-Ra could you solve the formatting issues?

outlines/serve/openai_server.py
```diff
@@ -0,0 +1,836 @@
...
+            usage=usage,
+        )
+
+        return response
+
+    # Streaming response
+    return StreamingResponse(
+            completion_stream_generator(), media_type="text/event-stream"
+        )
+    else:
```
danielchalef, line 499: @Soufiane-Ra Great work! We'd love to use this. It looks like there is another indentation error here. Could you fix it? Thx

findalexli:
Following this and would love to see any update

olaf-beh:
> I'd expect this to extend OpenAI's existing response_format field to include the schema, rather than use an extra_body field. Thoughts?

I cannot agree. To be openai compatible we have to use `extra_body`. They  specified `extra_body` field exactly for the use case to pass custom application fields.

E.g. if you use openai library for sending requests to `vllm.engine.openai.entrypoints.openai.apiserver` and you use an extra JSON field which is not specified by official openai API, then the openai python library will raise an error (I tried that a couple of weeks ago). The same is true for the other, previous vLLM extensions by the way. So in my eyes every extension must be wrapped into `extra_body`. Otherwise we 

1. are not openai compatible
2. cannot us the official openai python library with vLLM apiserver (at least not with vLLM extensions)

Just my two cents ...

rlouf:
Thank you for contributing! In the meantime vLLM integrated Outlines, and you can now use structured generation directly from there.

PR #599, opened 2024-02-01
introduce WHITESPACE regex constant

jc-louis:
To harmonize with other regexes that are already in constants.

This will also allow to do monkey patching of this regex, to do experiments related to https://github.com/outlines-dev/outlines/issues/484

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -15,6 +15,7 @@ INTEGER = r"(0|[1-9][0-9]*)"
 NUMBER = rf"(-)?({INTEGER})(\.[0-9]+)?([eE][+-][0-9]+)?"
 BOOLEAN = r"(true|false)"
 NULL = r"null"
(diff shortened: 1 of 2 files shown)


rlouf:
This is a very good idea, we might want to be able to pass that as an argument in the near future as well, from our internal experiments it looks like it has a substantial effect on the generation.

rlouf:
It looks like a warning that’s raised in `transfomers` is making the tests fail.

Soufiane-Ra:
+1 for changing outlines/fsm/json_schema.py from whitespace = r"[\n ]* to whitespace = r"[\n ]?" (or just deleting whitespace altogether) to avoid endless generation of new lines

rlouf:
It is not clear from our experiments that this is actually better: https://github.com/outlines-dev/outlines/issues/484. This PR will help a lot with experiments as one can now modify the value of `fsm.json_schema.WHITESPACE`. 

@jc-louis thank you for the PR!

PR #600, opened 2024-02-01
Update openai_compatible.py

7flash:
Fixing error

KeyError: 'Could not automatically map gpt4 to a tokeniser. Please use `tiktoken.get_encoding` to explicitly get the tokeniser you expect.'


rlouf:
Thank you for spotting this and correcting it!

PR #601, opened 2024-02-01
Update index.md - fix model

7flash:
Fixing an error

```
  File "<string>", line 4, in <module>
  File "/Users/gur/miniconda3/envs/jan25/lib/python3.10/site-packages/outlines/models/openai.py", line 128, in __init__
    raise ValueError(
ValueError: Invalid model_name. Check openai models list at https://platform.openai.com/docs/models
```


rlouf:
Thank you for catching this AND correcting it 🙂

PR #602, opened 2024-02-01
Update self_consistency.py - print actual answer

7flash:
```
{64: 1, 67: 9}
The answer is 67 with probability of 90.0%
```

* also reduces number of samples to 10 in order to avoid unexpected spendings when running this example


rlouf:
Thank you! I would change the wording a little bit to be "The *most likely* answer, etc."

PR #613, opened 2024-02-05
Unpin Transformers

lapp0:
Confirming the issue with torch with CI


rlouf:
Thank you !

> In `torch>=1.13.1,!=2.0.*`, the `OSError` described in #568 is resolved.

Any way we can simplify this by forcing `torch >= X`?

> We also should upgrade `outlines` version, as you cannot directly install `outlines[serve]` due to the dependency conflict between `vllm==0.3.0` (requires `transformers>=4.37.0`) and `outlines==0.0.26` (requires `transfomers==4.36.2`)

You mean cutting a new release once this has been merged?

Issue #624, opened 2024-02-08
Docstring of `RegexLogitsProcessor` is incorrect


sethkimmel3:
### Describe the issue as clearly as possible:

Related to #536, the JSONLogitsProcessor requires the same `adapt_tokenizer` modifications found [here](https://github.com/outlines-dev/outlines/blob/7fae436345e621a955e1e6ea610f74cf59f9466f/outlines/serve/vllm.py#L80), otherwise it throws this error: `AttributeError: 'LLM' object has no attribute 'tokenizer'`.

### Steps/code to reproduce the bug:

```python
from vllm import LLM
from outlines.serve.vllm import JSONLogitsProcessor, _patched_apply_logits_processors
from pydantic import BaseModel
from vllm import SamplingParams
import vllm.model_executor.layers.sampler as sampler

sampler._apply_logits_processors = _patched_apply_logits_processors

llm = LLM('mistralai/Mistral-7B-Instruct-v0.1')

class AviationPrompt(BaseModel):
    is_aviation_related: bool

logits_processor = JSONLogitsProcessor(AviationPrompt, llm)
results = llm.generate(user_prompts, sampling_params=SamplingParams(max_tokens=100, logits_processors=[logits_processor]))
```

### Expected result:

```shell
No errors
```

### Error message:

```shell
Traceback (most recent call last):
  File "/pkg/modal/_container_entrypoint.py", line 372, in handle_input_exception
    yield
  File "/pkg/modal/_container_entrypoint.py", line 530, in run_inputs
    res = imp_fun.fun(*args, **kwargs)
  File "/root/api.py", line 169, in generate
    logits_processor = JSONLogitsProcessor(AviationPrompt, self.llm)
  File "/usr/local/lib/python3.10/site-packages/outlines/serve/vllm.py", line 122, in __init__
    super().__init__(regex_string, llm)
  File "/usr/local/lib/python3.10/site-packages/outlines/serve/vllm.py", line 54, in __init__
    tokenizer = self.adapt_tokenizer(llm.tokenizer.tokenizer)
AttributeError: 'LLM' object has no attribute 'tokenizer'
Traceback (most recent call last):
  File "/pkg/modal/_container_entrypoint.py", line 372, in handle_input_exception
    yield
  File "/pkg/modal/_container_entrypoint.py", line 530, in run_inputs
    res = imp_fun.fun(*args, **kwargs)
  File "/root/api.py", line 487, in run_batch_inference
[...]

PR #625, opened 2024-02-08
Allow JSON Schema Constraining to a Single Whitespace

lapp0:
Allow disabling `multiple_ws` argument in JSON constrained generation. Results in a maximum of one consecutive syntactic whitespace (no impact on string literals).

Fixes https://github.com/outlines-dev/outlines/issues/624

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -1,7 +1,7 @@
 import inspect
 import json
 import re
-from typing import Callable, Union
+from typing import Callable, Optional, Union
 
 from jsonschema.protocols import Validator
 from pydantic import BaseModel, create_model
@@ -38,7 +38,9 @@ format_to_regex = {
 }
 
 
-def build_regex_from_object(object: Union[str, Callable, BaseModel]):
+def build_regex_from_object(
+    object: Union[str, Callable, BaseModel], whitespace_pattern: Optional[str] = None
+):
     """Turn a JSON schema into a regex that matches any JSON object that follows
     this schema.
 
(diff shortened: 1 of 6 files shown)


rlouf:
I think it would be cleaner to allow users to pass the regex for the whitespace directly and default to the current `WHITESPACE` value.

outlines/fsm/json_schema.py
```diff
@@ -105,8 +110,16 @@ def to_regex(resolver: Resolver, instance: dict):
         An object that resolves references to other instances within a schema
     instance
         The instance to translate
+    multiple_ws
+        determines whether to include multiple consecutive whitespace in JSON syntax
     """
 
+    # determine whitespace pattern
+    if multiple_ws:
+        whitespace_pattern = WHITESPACE
```
rlouf, line 119: Pass `whitespace_pattern` as an argument directly. This give fine-grained control without having to patch the `WHITESPACE` variable.

lapp0: If we do this, I think it makes sense to validate that their pattern is a proper subset of the legal json whitespace characters `[\t\r\n ]*`. Fortunately that's exactly what `interegular` is designed to do.

rlouf: I think it's completely fine to let users do what they want here; it is not worth the extra code.

outlines/fsm/json_schema.py
```diff
@@ -83,10 +100,12 @@ def build_regex_from_object(object: Union[str, Callable, BaseModel]):
     resolver = registry.resolver()
 
     content = schema.contents
-    return to_regex(resolver, content)
+    return to_regex(resolver, content, whitespace_pattern)
 
 
-def to_regex(resolver: Resolver, instance: dict):
```
rlouf, line 89: We should let users pass this parameter to `outlines.generate.json` as well.

rlouf:
I fixed the merge conflicts and added documentation. Thank you for contributing!

Issue #649, opened 2024-02-12
Device error for JSON generation with outlines 0.0.29


streitl:
### Describe the issue as clearly as possible:

I'm trying to use outlines to generate json objects based on a Pydantic model.
I am loading my model on a GPU, but when I try to generate an object, I get an error saying that pytorch also found tensors on the CPU.

### Steps/code to reproduce the bug:

```python
import outlines
import torch

device: str = "cuda"
rng = torch.Generator(device=device)
rng.manual_seed(90302)

outlines_model = outlines.models.transformers(model_id, device=device)
generator = outlines.generate.json(outlines_model, MyPydanticModel, whitespace_pattern="")

outlines_output = generator("test prompt", rng=rng)
```

### Expected result:

```shell
A MyPydanticModel instance.
```

### Error message:

```shell
Traceback (most recent call last):
  File "/my_example.py", line ..., in <module>
    outlines_output = generator("test prompt", rng=rng)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/outlines/generate/api.py", line 205, in __call__
    last_state = next(states)
                 ^^^^^^^^^^^^
  File "lib/python3.11/site-packages/outlines/generate/generator.py", line 83, in sequence_generator
    token_ids = update_token_ids(token_ids, next_token_ids, ancestors)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "lib/python3.11/site-packages/outlines/generate/generator.py", line 198, in update_token_ids
    token_ids = torch.index_select(token_ids, 0, ancestors)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument index in method wrapper_CUDA__index_select)
```

### Outlines/Python version information:

<details>

```
outlines 0.0.29
Python 3.11.7 (main, Dec  8 2023, 18:56:58) [GCC 11.4.0]
accelerate==0.26.1
aiofiles==23.2.1
aiohttp==3.9.3
aiosignal==1.3.1
[...]

PR #651, opened 2024-02-12
Ensure Ancestors on Correct Device During Sampling

lapp0:
Fixes https://github.com/outlines-dev/outlines/issues/649

### DIFF
--- a/outlines/samplers.py
+++ b/outlines/samplers.py
@@ -147,7 +147,7 @@ class MultinomialSampler:
 
         logprobs = torch.nn.functional.log_softmax(altered_next_token_logits, dim=-1)
         ancestors = torch.arange(
-            altered_next_token_logits.shape[0], device=altered_next_token_logits.device
+            altered_next_token_logits.shape[0], device=next_token_logits.device
         )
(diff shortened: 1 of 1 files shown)


rlouf:
You would need to do this for every sampling method

lapp0:
Good catch. Beam and greedy needs changes as well. 

Smoke tested, can confirm for all 3 samplers the device is set correctly:

```
>>> next_token_ids.device, ancestors.device, weights.device
(device(type='cuda', index=0), device(type='cuda', index=0), device(type='cuda', index=0))
```

outlines/samplers.py
```diff
@@ -66,7 +66,9 @@ def __call__(
         logprobs = torch.nn.functional.log_softmax(next_token_logits, dim=-1)
         next_token_ids = torch.argmax(logprobs, dim=-1, keepdim=True)
 
-        ancestors = torch.arange(next_token_logits.shape[0])
+        ancestors = torch.arange(
+            next_token_logits.shape[0], device=next_token_logits.device
```
rlouf, line 70: nit: let's be consistent across samplers and use `next_token_logits` for each sampler :)

rlouf:
Thanks for the fix!

Issue #656, opened 2024-02-13
Issues with CUDA and the examples in the tutorial 


raphaelchinchilla:
### Describe the issue as clearly as possible:

I am trying to get outlines to work on CUDA. 

I just modified the example in the README to force the model to go to CUDA (otherwise it would go to the CPU).

However, I get an error. It seems that the tokenizer is not sending the prompt to the CUDA.

### Steps/code to reproduce the bug:

```python
from enum import Enum
from pydantic import BaseModel, constr

import outlines
import torch


class Weapon(str, Enum):
    sword = "sword"
    axe = "axe"
    mace = "mace"
    spear = "spear"
    bow = "bow"
    crossbow = "crossbow"


class Armor(str, Enum):
    leather = "leather"
    chainmail = "chainmail"
    plate = "plate"


class Character(BaseModel):
    name: constr(max_length=10)
    age: int
    armor: Armor
    weapon: Weapon
    strength: int


model = outlines.models.transformers("mistralai/Mistral-7B-Instruct-v0.2",device="cuda")

# Construct structured sequence generator
generator = outlines.generate.json(model, Character)

# Draw a sample
rng = torch.Generator(device="cuda")
rng.manual_seed(789001)

character = generator("Give me a character description", rng=rng)

print(repr(character))
# Character(name='Anderson', age=28, armor=<Armor.chainmail: 'chainmail'>, weapon=<Weapon.sword: 'sword'>, strength=8)

character = generator("Give me an interesting character description", rng=rng)

print(repr(character))
# Character(name='Vivian Thr', age=44, armor=<Armor.plate: 'plate'>, weapon=<Weapon.crossbow: 'crossbow'>, strength=125)
```

### Expected result:

```shell
The one on the README
```

### Error message:

```shell
On line 198 of outlines/generate/generator.py

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cudo:0 and cpu! (when checking argument for argument index in method wrapper_CUDA__index_select)
```

### Outlines/Python version information:

Version information
<details>
```
(command output here)
```
</details>

rlouf:
I just released `0.0.30`, could you update and try again ?

raphaelchinchilla:
It now works with device="cuda" but does not work with device="auto". The error I get is

On line 198 of outlines/generate/generator.py

[...]

PR #660, opened 2024-02-14
Put `ancestors` on the same device as the KV cache

raphaelchinchilla:
Addressed the issue reported in #656 with the solution proposed by @rlouf . 

I am really not sure whether this is the best solution.

### DIFF
--- a/outlines/generate/generator.py
+++ b/outlines/generate/generator.py
@@ -265,7 +265,7 @@ def reorder_kv_cache(
     for cache_item in kv_cache:
         new_cache_item: Tuple = tuple()
         for layer in cache_item:
-            layer = torch.index_select(layer, 0, ancestors)
+            layer = torch.index_select(layer, 0, ancestors.to(layer.device))
             new_cache_item += (layer,)
         new_kv_cache += (new_cache_item,)
 


lapp0 (commented):
Thanks for contributing!

I think we should synchronize the device to `ancestors` since I believe (correct me if I'm wrong) it's possible that the kv_cache layers are on different devices.


Also could you please smoke test this with all 3 samplers on a 2 GPU system with `device="auto"` set?

raphaelchinchilla:
> Thanks for contributing!

I am not sure whether this is really a contribution, I just did what @rlouf suggested :)

> I think we should synchronize the device to ancestors since I believe (correct me if I'm wrong) it's possible that the kv_cache layers are on different devices.

Honestly, I understand very little about the package (I started using it yesterday), and I do not feel I can actually give my opinion on whether you are wrong about it or not. It does feel like too simple of a contribution.

> Also could you please smoke test this with all 3 samplers on a 2 GPU system with `device="auto"` set?

Sorry, which 3 samplers are you talking about?

rlouf:
I just checked `transformers` and that's how they do it. Might not be optimal but if it solves @raphaelchinchilla'a problem let's merge it.

PR #673, opened 2024-02-16
Add Particle Filter

rlouf:
See Nicolas Chopin's [book](https://link.springer.com/book/10.1007/978-3-030-47845-2) for a really nice introduction to the topic. The algorithms consists in carrying $N$ particles for each sequence in the batch, and at each step to:

1. Sample a new token for each particle using the next-token logits.
2. Resample the particles.

We use the multinomial resampling function in this first PR, although it is known to have very large variance. To make the implementation easier we combine (1) and (2) in a single step, similarly to what we do with beam search.

Note that there is a subtlety when doing structured generation. We can think of the simple following scheme to sample from the distribution of sequences that follow the structure:

1. Move particles by one step using the *unbiased* next-token logits;
2. Set the weight of each invalid particle to $-\infty$
3. Resample

But this can be very inefficient. Instead, we move particles using a specific proposal: using the *biased* next-token logits. Since this is not exactly sampling from the original distribution we need to resample the particles using the factor $P_i  / \tilde{P}_i$ as a weight where $P_i$ is the unbiased probability of token $i$ and $P_i$ the biased probability of token $i$ (importance sampling).

**Note:** I am wondering if we should correct the Beam Search algorithm as well.

### DIFF
--- a/outlines/generate/generator.py
+++ b/outlines/generate/generator.py
@@ -77,7 +77,7 @@ def sequence_generator(
         allowed_tokens = get_allowed_tokens(fsms, fsm_states)
         biased_logits = bias_logits(logits, allowed_tokens)
         next_token_ids, ancestors, sequence_weights = sampler(
-            biased_logits, sequence_weights, rng
+            logits, biased_logits, sequence_weights, rng
         )
 
         token_ids = update_token_ids(token_ids, next_token_ids, ancestors)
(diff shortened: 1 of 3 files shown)


lapp0:
Considering posteriors during inference is extremely important for constrained generation and infilling!

rlouf:
Doing this I started to wonder if we shouldn't see and implement greedy and multinomial sampling as particular cases of more general samplers (resp. a form of beam search and a form of particle filtering).

PR #696, opened 2024-02-21
fix actions

lapp0:
Release Docker dispatch: https://github.com/lapp0/outlines/actions/runs/7994419887
- "Fails successfully": Got to the point where it only auth errors. `Error: buildx failed with: ERROR: denied: requested access to the resource is denied`

Not testing fetch PyPi. Changes are minimal between this version and the previous *working* main.

```
git diff e99d92d024dbf6f6bff10a9c3954f326cf4a0cd3 -- .github/workflows/release_pypi.yaml .github/workflows/release.yml | cat
diff --git a/.github/workflows/release.yml b/.github/workflows/release_pypi.yaml
similarity index 92%
rename from .github/workflows/release.yml
rename to .github/workflows/release_pypi.yaml
index e6bf1b1..597ebb7 100644
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release_pypi.yaml
@@ -1,16 +1,17 @@
-name: Release
+name: Release PyPi
 
 on:
   release:
     types:
       - created
-
 jobs:
   release-job:
     name: Build and publish on PyPi
     runs-on: ubuntu-latest
+    environment: release
     steps:
-    - uses: actions/checkout@v2
+    - name: Checkout
+      uses: actions/checkout@v2
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
```

### DIFF
--- a/.github/workflows/release_docker.yml
+++ b/.github/workflows/release_docker.yml
@@ -16,7 +16,6 @@ jobs:
     runs-on: ubuntu-latest
     environment: release
     steps:
-
     - name: Checkout
       uses: actions/checkout@v4
     - name: Log in to Docker Hub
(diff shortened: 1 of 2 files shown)


rlouf:
Now works, thank you for the quick fix!

Issue #670, opened 2024-02-16
Add `outlines.generate.fsm()` API entrypoint


rlouf:
As suggested in https://github.com/outlines-dev/outlines/issues/666#issuecomment-1947685486, we should allow users to pass any DFA (with vocabulary = characters) to generate text that is in the vocabulary of that DFA.

lapp0:
We should allow the passing of an `interegular.FSM` because the desired operations requested in the linked issue, and those requested by users in other issues are not possible with a `RegexFSM`.

We should modify `RegexFSM`, perhaps to have `RegexFSM.from_interegular_fsm`

https://github.com/outlines-dev/outlines/blob/main/outlines/fsm/fsm.py#L94-L125

rlouf:
> We should modify `RegexFSM`, to have `RegexFSM.from_interegular_fsm`

This is a hint that we have a naming problem here. And we might as well fix it until more libraries depend on Outlines' internals.

lapp0:
Maybe `TokenIndexFSM`

miftahmoha:
Can someone please explain to me, I'm confused. 

Does the user have to create the FSM with `interegular` and pass it to a method `outlines.generate.fsm(fsm=interegular_fsm)` which returns the `SequenceGenerator`?

I guess `RegexFSM.from_interegular_fsm` should turn an `interegular_fsm` object into a `RegexFSM` object so that it can be processed by `SequenceGenerator`?

I'm interested in working on it if it's possible.

lapp0:
>Does the user have to create the FSM with interegular and pass it to a method outlines.generate.fsm(fsm=interegular_fsm) which returns the SequenceGenerator?

Yes. `interegular` allows you to perform helpful operations on FSMs - concatenation, OR / XOR, intersection, negation, etc. `RegexFSM` doesn't have these capabilities.  

>I guess RegexFSM.from_interegular_fsm should turn an interegular_fsm object into a RegexFSM object so that it can be processed by SequenceGenerator?

Yes exactly!

>I'm interested in working on it if it's possible.

That's great news! 

[The current `RegexFSM.__init__` logic](https://github.com/outlines-dev/outlines/blob/main/outlines/fsm/fsm.py#L94-L125) involves conversion from pattern -> `interegular.FSM` -> initialization of Outlines `RegexFSM`. Ideally `from_interegular_fsm` wouldn't repeat logic with `__init__`, but instead they'd share as much logic as possible using a helper function.

[...]

Issue #666, opened 2024-02-15
Adding blacklist of words LLM should never use in generation


dhuynh95:
### Presentation of the new feature

I want to ban some words from being generated, such as "might" or "hypothetically," as they are considered foul language for my use case.

This is a big use case in AI as there is a lot of literature on how to not have the LLM talk about specific topics.

For instance, https://arxiv.org/abs/2402.07896 looks at it using DPO but has no guarantees.

### Where does it fit in Outlines?

I think this could be a generation mode where we provide a list of words we don't to be ever used.
This could look like:

```python
import outlines

model = outlines.models.transformers("mistralai/Mistral-7B-v0.1")

prompt = "Don't talk about pink elephants. What is an elephant?"
answer = outlines.generate.negative_choice(model, ["is", "are"]
print(answer)
```

### Are you willing to open a PR?

I could try to help but I don't know how to get started.

Thanks for the help!

lapp0:
I think this is addressed by https://github.com/outlines-dev/outlines/issues/156

It's helpful to know that this is needed so we can prioritize it, thanks for submitting!

rlouf:
> I believe this can be implemented with `interegular` subtraction (`base_fsm - is_are_fsm`) 
> 
> 
> 
> https://github.com/MegaIng/interegular/blob/master/interegular/fsm.py#L721-L726

That's a really good point!

dhuynh95:
Thanks for the answer! Is it something that would be straightforward? I don't know too much how the low-level components work but I guess I could use some middle-level functions to create the full FSM, derive a FSM from the different words to ban, subtract them using interegular, and patch it to outlines?

lapp0:
IMO we should create something like `outlines.generate.fsm()`, there's a lot of feature requests that would benefit from this flexibility. 

An immediate solution is to use the greenery library to create a pattern, which **may** work for your use case.

```
>>> import greenery
>>> str(greenery.parse(".*") - greenery.parse("pink|elephant"))
'(([^ep]|e([^l]|l([^e]|e([^p]|p([^h]|h([^a]|a([^n]|n[^t]))|hant.))))|p([^i]|i([^n]|n([^k]|k.)))).*|e(l(e(p(h(an?)?)?)?)?)?|p(in?)?)?'
```

Issue #156, opened 2023-06-21
Forbid sequences in generations


rlouf:
I propose to forbid sequences in generated completions using a `not_in` keyword argument to `Completion`. This can be implemented using masking and re-weighing.

PR #699, opened 2024-02-21
Add outlines.generate.fsm() API entrypoint

miftahmoha:
This adds the outlines.generate.fsm() API entrypoint following https://github.com/outlines-dev/outlines/issues/670.

#### Tests:
tests

### DIFF
--- a/outlines/fsm/fsm.py
+++ b/outlines/fsm/fsm.py
@@ -94,7 +94,7 @@ class RegexFSM(FSM):
     def __init__(self, regex_string: str, tokenizer):
         @cache()
         def create_states_mapping(
-            regex_string: str, cacheable_vocabulary: Tuple[Tuple[str, int]]
+            regex_string: str, cacheable_vocabulary: Tuple[Tuple[str, int], ...]
         ) -> Tuple[dict, set]:
             """Create the variables related to the mapping between states and tokens
             The parameters of the function are used for caching purpose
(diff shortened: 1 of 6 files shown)


outlines/fsm/fsm.py
```diff
@@ -182,6 +182,47 @@ def next_state(self, state: FSMState, token_id: int) -> FSMState:
...
         return FSMState(next_state)
 
+    @classmethod
+    def from_interegular_fsm(
+        cls, interegular_fsm: interegular.fsm.FSM, tokenizer: "Tokenizer"
+    ):
+        from_interegular_instance = cls.__new__(cls)
+
+        @cache()
+        def create_states_mapping_from_interegular_fsm(
```
lapp0, line 192: As discussed here, `@cache` doesn't work with FSMs. 

https://github.com/outlines-dev/outlines/issues/670#issuecomment-1958266151

We should leave this uncached for now. There needs to be an upstream fix in `interegular` which makes `AnythingElse` have a consistent hash before we can cache.

We should leave this uncached for now.

lapp0 (commented):
Could you please review the test cases `tests/fsm/test_regex.py` and introduce test cases for this new method?

It's probably sufficient to simply verify that `RegexFSM.from_interegular_fsm(interegular.parse_pattern(pattern)) == RegexFSM(pattern)` for a few `pattern`s.

miftahmoha:
@rlouf Done. 

According to https://github.com/outlines-dev/outlines/issues/670#issuecomment-1949402672, `interegular.FSM` offers operations that `RegexFSM` doesn't seem to have. For the examples, we've got https://github.com/outlines-dev/outlines/issues/666 and https://github.com/outlines-dev/outlines/issues/156 that we can use.

What do you think @lapp0?

lapp0:
@miftahmoha yes, we should document those use cases. This perfectly solves those problems. 

Use a pattern but disallow certain keywords.

Example: A list of strings, but none of those strings can be "pink" or "elephant"
```
import interegular

list_of_strings_pattern = """\["[^"\s]*"(?:,"[^"\s]*")*\]"""
pink_elephant_pattern = """.*(pink|elephant).*"""

list_of_strings_fsm = interegular.parse_pattern(list_of_strings_pattern).to_fsm()
pink_elephant_fsm = interegular.parse_pattern(pink_elephant_pattern).to_fsm()



interegular.parse_pattern("
```

miftahmoha:
@lapp0 @rlouf I added some documentation.

rlouf:
Thank you so much for the valuable addition @miftahmoha !

PR #717, opened 2024-02-29
Allow specifying own Hugging Face tokenizer instance

shawnz:
This gives the option of specifying your own `PreTrainedTokenizer` instance rather than specifying a model name and having outlines construct the tokenizer for you. This might be convenient for situations where you are already using the LLM in other parts of your application and only need outlines for a specific use case.

Example of use:

```python
from transformers import AutoTokenizer, AutoModelForCausalLM
from outlines.models.transformers import TransformerTokenizer, Transformer

model_name = "hf-internal-testing/tiny-random-GPTJForCausalLM"
hf_tokenizer = AutoTokenizer.from_pretrained(model_name)
hf_model = AutoModelForCausalLM.from_pretrained(model_name)
tokenizer = TransformerTokenizer(hf_tokenizer)
model = Transformer(hf_model, tokenizer)
```

As compared to before this change:

```python
from transformers import AutoTokenizer, AutoModelForCausalLM
from outlines.models.transformers import TransformerTokenizer, Transformer

model_name = "hf-internal-testing/tiny-random-GPTJForCausalLM"
hf_tokenizer = AutoTokenizer.from_pretrained(model_name)
hf_model = AutoModelForCausalLM.from_pretrained(model_name)
tokenizer = TransformerTokenizer(model_name) # Can't make use of hf_tokenizer here, even if
                                             # you already instantiated it for other reasons!
model = Transformer(hf_model, tokenizer)
```

This also corrects a type annotation on the `outlines.models.transformers.Transformer` class which incorrectly stated that the `tokenizer` argument was of type `transformers.PreTrainedTokenizer` when actually it should be of type `outlines.models.transformers.TransformerTokenizer`. In order to fix the annotation, the classes also had to be rearranged so that the name of the other class would be defined at the right time.

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -5,7 +5,7 @@ import torch
 from outlines.models.tokenizer import Tokenizer
 
 if TYPE_CHECKING:
-    from transformers import PreTrainedModel, PreTrainedTokenizer
+    from transformers import PreTrainedModel, PreTrainedTokenizerBase
 
 __all__ = ["transformers"]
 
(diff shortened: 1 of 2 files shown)


rlouf:
This part of the interface has been bugging me for a while, thank you for helping!

PR #727, opened 2024-03-04
Clean the `transformers` and `llamacpp` interfaces

### LINKED ISSUE #723: Let users pass `transformers` and `llama_cpp` models directly
We keep the high-level API which should work for most use cases, but allow users to initialize models with these libraries and initialise Outlines models using the following API:

```python
from outlines import models

model = models.transformers("model_name")
models = models.llamacpp("path")


import transformers

llm = transformers.AutoModelForCausalLM("model_name")
tokenizer = transformers.AutoTokenizer("model_name")
llm = models.Transformers(model, tokenizer)


from llama_cpp import Llama

llm = Llama(model_path="path")
model = models.LlamaCpp(model)

```

rlouf:
Closes #723

### DIFF
--- a/docs/reference/models/llamacpp.md
+++ b/docs/reference/models/llamacpp.md
@@ -4,12 +4,41 @@
 
     You need to install the `llama-cpp-python` library to be able to use these models in Outlines.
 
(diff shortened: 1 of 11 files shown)


rlouf:
This is good to merge. Let's now adopt the same pattern in #593.

PR #728, opened 2024-03-04
Feat/add transformers integration

### LINKED ISSUE #713: Add integration with Hugging Face `transformers`
### Presentation of the new feature
It should be possible to use the `transformers` package for inference of generative models, and simply add structured generation from `outlines` as a "plugin" rather than needing to wrap all models in `outlines`-specific classes, as it seems like is the current approach.

Instead, `transformers` supports a `prefix_allowed_tokens_fn` argument in the `generate` method, which is a function that returns the allowed tokens to be generated at a given step. The `outlines` package could thus have a simple function/class, which can be given as this argument, analogous to the current vLLM integration with `logits_processors`.

### Where does it fit in Outlines?
Allows easier integration into inference frameworks that people are mostly using, making `outlines` more useful to many people.

### Are you willing to open a PR?
I would be willing to implement this in a PR, yes. The implementation would be very similar to the current vLLM integration. If I am going to do this, I would need some guidance on preferred directory structures, however. The vLLM integration is inside the `serve` directory. Should vLLM and this `transformers` integration be moved into a separate `integrations` directory, perhaps?

saattrupdan:
This adds integration with the `transformers` package, by supplying a `prefix_allowed_tokens_fn` that can be supplied to both `transformers` pipelines and directly to `generate` methods of generative models from the `transformers` package.

The `transformers` integration has been put in an `integrations` top-level directory, and the existing vLLM integration has also been moved into that directory. We keep a reference in the previous `serve/vllm.py` module, to preserve backwards compatibility.

One notable difference between vLLM and transformers is that transformers include the input in the generated sequences. Since this can mess up the FSMs, we keep track of the input tokens, and use these both to identify when we change to new samples (resetting the FSM) as well as to know which token ID prefix to remove, before getting the next state from the FSM.

Also added an example with the transformers integration.

Closes #713


outlines/serve/vllm.py
```diff
@@ -1,143 +1,4 @@
...
-        else:
-            raise ValueError(
-                f"Cannot parse schema {schema}. The schema must be either "
-                + "a Pydantic object, a dictionary or a string that contains the JSON "
-                + "Schema specification"
-            )
-        regex_string = build_regex_from_schema(schema_str, whitespace_pattern)
-        super().__init__(regex_string, llm)
+from outlines.integrations.vllm import (  # noqa[F401]
+    JSONLogitsProcessor,
```
rlouf, line 2: It would be preferable to also change the imports in `outlines.serve.serve`

rlouf:
Could you rebase your branch on `main` to remove the merge commits? I haven't had time to test it locally yet, but will get around to it soon. It would be great to also move the integration logic that's in `outlines.models.llamacpp.py` in `outlines.integrations`, if you have time to do it.

saattrupdan:
> It would be great to also move the integration logic that's in `outlines.models.llamacpp.py` in `outlines.integrations`, if you have time to do it.

Sure thing. Is it correctly understood that this concerns `LogitsProcessor`, `RegexLogitsProcessor` and `CFGLogitsProcessor` in the `models.llamacpp` module that should be moved to `integrations.llamacpp`?

> Could you rebase your branch on main to remove the merge commits?

Sure thing, I'll do that when the llamacpp integration is set up.

saattrupdan:
@rlouf The LlamaCpp integration has been moved now, and have also adapted the imports in the `models.llamacpp` module to point to `integrations.llamacpp` now. Did a rebase as well. Hopefully it's ready for a merge now!

rlouf:
Will take a look early next week!

outlines/generate/api.py
```diff
@@ -101,7 +101,7 @@ def strip_stop_sequences(
 
         return sequence
 
-    def format_sequence(self, sequence: str) -> str:
+    def format_sequence(self, sequence: str) -> Any:
```
rlouf, line 104: Is it possible to avoid using `Any`?

saattrupdan: It is, yep! It would just be a big union, due to all the different type casts, but I can add that instead.

saattrupdan: Changed this now.

rlouf:
I just fixed merged conflicts due to a big refactor of the `FSM` interface. I will take a closer look at the PR tomorrow.

Kamakshi8104:
Can this be used with multimodal models too?

rlouf:
> Can this be used with multimodal models too?

It should!

Kamakshi8104:
thanks!! Can't wait to try it out!!

rlouf:
I was able to run the example locally, and tried the vLLM integration as well; llama.cpp is covered by the integration tests. Great job!

@Kamakshi8104 would you like to add an example using a multimodal model to the docs?

Kamakshi8104:
Yes I can add an example🙂. I am in the middle of exams so will get started on it this coming week👍

jeremyallenjacobson:
> Yes I can add an example🙂. I am in the middle of exams so will get started on it this coming week👍

Any update on this? Very interested in a seeing a multimodal example, thanks!

PR #733, opened 2024-03-06
[Draft] Add option to ignore fields order

leloykun:
# Description

Predibase recently published [an article](https://predibase.com/blog/lorax-outlines-better-json-extraction-with-structured-generation-and-lora) wherein they discussed an issue with the fact that Outlines imposes an ordering constraint on object fields. See below:
image

This PR just adds an option to turn this constraint off. This does not change the default API interface at all.

There are probably a lot of smarter ways to do this, but this is good enough of a hack.

TODOs:
- [ ] Add tests
- [ ] Add docs

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -1,7 +1,8 @@
 import inspect
 import json
 import re
-from typing import Callable, Optional
+from itertools import permutations
+from typing import Any, Callable, Dict, Iterable, List, Optional
 
 from jsonschema.protocols import Validator
 from pydantic import create_model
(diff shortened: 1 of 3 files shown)


rlouf:
Thank you! My worry with this technique is that it will make the regexes much longer and therefore increase compilation time for all schemas. Would you mind taking a couple examples and giving the length of the resulting regexes with and without your change?

leloykun:
Closing this as I have a better idea in mind

Maximilian-Winter:
@leloykun Would you mind telling me your idea? I have a problem with ordering. Can I somehow make it like in the schema?

PR #745, opened 2024-03-14
Check if the given token is a string

silverriver:
Some model use bytes as their tokens, such as Qwen (see: https://huggingface.co/Qwen/Qwen-7B/blob/ef3c5c9c57b252f3149c1408daf4d649ec8b6c85/tokenization_qwen.py#L136 )

### DIFF
--- a/outlines/integrations/utils.py
+++ b/outlines/integrations/utils.py
@@ -56,7 +56,11 @@ def adapt_tokenizer(tokenizer: PreTrainedTokenizerBase) -> PreTrainedTokenizerBa
         string = tokenizer.convert_tokens_to_string([token])
 
         # A hack to handle missing spaces to HF's Llama tokenizers
-        if type(token) is str and token.startswith(SPIECE_UNDERLINE) or token == "<0x20>":
+        if (
+            type(token) is str
+            and token.startswith(SPIECE_UNDERLINE)
+            or token == "<0x20>"
+        ):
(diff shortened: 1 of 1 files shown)


rlouf:
Thank you! Is this solving an issue you encountered? Also would you mind running `pre-commit` to fix code style?

silverriver:
> Thank you! Is this solving an issue you encountered? Also would you mind running `pre-commit` to fix code style?

Yeap, this solves my issue when I am using the Qwen model family, such as https://huggingface.co/Qwen/Qwen-1_8B.

Here is a code snip:
```python
import vllm
from outlines.serve.vllm import RegexLogitsProcessor

tokenizer = AutoTokenizer.from_pretrained('Qwen/Qwen-1_8B', trust_remote_code=True)
prompt = "Answer with A, B, C, or D."

llm = vllm.LLM(model="Qwen/Qwen-1_8B", trust_remote_code=True)
logits_processor = RegexLogitsProcessor(regex_string=r'[ABCD]', llm=llm)
outputs = llm.generate([prompt],
            sampling_params=vllm.SamplingParams(
                max_tokens=5, logits_processors=[logits_processor]
            ),
        )
```

I have also found out that the Qwen model family use a non-standard tokenizer implementation, i.e., they return byte sequences as their tokens, not strings. However, I think it is always benifical to consider more corner cases.

PR #775, opened 2024-03-28
Small typo fix in cookbook example

mkretsch327:
I noticed an import path must have changed and not been reflected in the cookbook examples for 'Classification' in the docs. I didn't see other examples. This PR only is for the example I noticed.


rlouf:
Thank you!

Issue #776, opened 2024-03-29
JSON Integers cannot be negative?


posionus:
Hello,
I'm wondering if there is a particular reason why it seems JSON integers cannot be negative.

See:
https://github.com/outlines-dev/outlines/blob/aed9d21bb622e3579c3296a73a0240e36d04180d/outlines/fsm/json_schema.py#L14

"No negatives" seems to be explicitly tested for:
https://github.com/outlines-dev/outlines/blob/aed9d21bb622e3579c3296a73a0240e36d04180d/tests/fsm/test_json_schema.py#L65

This seems incorrect to me, but maybe I'm missing something.

rlouf:
This is indeed wrong. Would you be willing to submit a PR?

PR #777, opened 2024-03-29
allow json ints to be negative

posionus:
https://github.com/outlines-dev/outlines/issues/776


posionus:
@rlouf not sure why the one check is failing. Doesn't seem related to anything I changed, but I could be wrong

rlouf:
No it's not unrelated, I need to investigate.

Issue #667, opened 2024-02-15
Continuous generation in Outlines


rlouf:
I am opening this issue to roughly sketch the next big milestone for Outlines, tentatively called "continuous generation". There are many rough edges still, and open questions.

The first goal is to allow sampling of sequences like these:

```python
from outlines import generate, models

model = models.transformers("mistalai/Mistral-7B-v0.1")
generator = generate.text(model)

sequence = "What are the most popular types of vehicles?\n"
for i in range(6):
      sequence += f"{i}, "
      sequence += generator(sequence, stop_at=["\n"])
      sequence += "\n"
```

By "sampling these sequences" I mean being able to run, for instance, beam search and optimize the sequence as a whole rather than each generation separately. 

All we have to do is to return a `Sequence` object instead of a string, with the following attributes and methods:

```python
class Sequence:
    token_ids: torch.Tensor
    weights: torch.Tensor
    kv_cache: Tuple
    tokenizer: Tokenizer

    def __str__(self):
        return tokenizer.decode(token_ids)
```

`Sequence` should have the same *feel* as a string. Besides being able to print it, we should be able to slice it, add it to another string, another sequence, etc. and carry on:

```python
class Sequence:
    ...
    def __getitem__(self, key):
        if isinstance(key, int):
            # Just return the character? There's not much more we can do here.
        if isintance(key, slice):
           # Different behavior depending on whether `start` is 0. If `start = 0` we can 
           # keep part of the KV Cache.  Otherwise we need to re-compute the KV 
           # Cache i.e. consider the `Sequence` as a new prompt. 
           #
           # We will likely need to split tokens. For instance if we call `sequence[:10]` and 
           # 10 is the letter `m` in `formida`. In this case we can encode and append `afor` 
           # to the previous token ids. Edge cases should automatically be handled when 
           # aligning prompt and generation. 
    def __add__(self, other):
         if isinstance(other, str):
            # Signal that KV cache + logprob need to be re-computed
         if isinstance(other, Sequence):
             # Concatenate token_ids
             # Concatenate logprobs
             # Signal that KV Cache after `other` needs to be recomputed
```

This should be enough to bring Outlines at feature-parity with other DSLs, while not being a DSL.

cpfiffer:
It may also be interesting to get the join token likelihood, if available. I'm not _super_ familiar with outlines but I'd love to be able to compare `Sequence`s probabilistic.

jeffreyftang:
[...]

PR #781, opened 2024-04-01
One small step towards continuous generation for Outlines

miftahmoha:
This PR is for solving https://github.com/outlines-dev/outlines/issues/667.

last_tests

last_pre_commit

# API

```python
import outlines
from outlines.generate import continuous

generator_c = continuous(generator)

response = generator_c(prompt, max_tokens=30)
```

## `continuous`

`continuous` wraps any `SequenceGenerator` object, it could be:

- `outlines.generate.choice`
- `outlines.generate.text`
- `outlines.generate.json`
- `...`

The `continuous` wrapper allows the generator to save the state of a `Sequence`, it means that, if you _continuously_ generate a sequence as shown:

```python
import outlines
from outlines.generate import continuous

generator_c = continuous(generator)

response_1 = generator_c(prompt, max_tokens=100)
response_2 = generator_c(response)
```

KV Cache _(under some conditions)_ will be saved. Algorithms such as beam search could be used to optimize the whole sequence rather than separately.

```python
import outlines
from outlines.generate import continuous

generator = outlines.generate.text(model, sampler=BeamSearchSampler(3))
generator_c = continuous(generator)

response_1 = generator_c(prompt, max_tokens=100)
response_2 = generator_c(response)
```

You can mix different types of `SequenceGenerator` objects:

```python
import outlines
from outlines.generate import continuous

generator_text = outlines.generate.text(model)
generator_choice  = outlines.generate.choice(model, ["Positive", "Negative"])

generator_text_c = continuous(generator_text)
generator_choice_c = continuous(generator_choice)

response_1 = generator_text_c(prompt, max_tokens=100)
response_2 = generator_choice_c(response_1)
```

Once a prompt is given to the `continuous` wrapper, it becomes a `SequenceState` object.

```python
class SequenceState:
    token_ids: torch.Tensor
    weights: torch.Tensor
    attention_masks: torch.Tensor
    kv_cache: torch.Tensor
    tokenizer: "Tokenizer"
```

### Indexing

Each `SequenceState` has three dimensions `SequenceState[batch_key: Union[int, slice], sample_key: Union[int, slice], ids_size_key: Union[int, slice]]`.

However, there are three cases where this is handled differently:

1. `batch_size == 1` and `sample_size == 1`
   `SequenceState[ids_size_key: Union[int, slice]]`, instead of `SequenceState[0, 0, ids_size_key: Union[int, slice]]`.

2. `batch_size == 1`
   `SequenceState[sample_key: Union[int, slice], ids_size_key: Union[int, slice]]`, instead of `SequenceState[0, sample_key: Union[int, slice], ids_size_key: Union[int, slice]]`.

3. `sample_size == 1`
   `SequenceState[batch_key: Union[int, slice], ids_size_key: Union[int, slice]]`, instead of `SequenceState[batch_key: Union[int, slice], 0, ids_size_key: Union[int, slice]]`.

### Operations

You can apply two operations on a SequenceState:

- Slicing

- Adding _(`SequenceState` to a `SequenceState` and `SequenceState` to a prompt)_

#### Adding

1. _`SequenceState` to a `SequenceState`_

**This won't save the first part of the KV Cache for the moment**, but it does accumulate the weights between both sequences.

I don't have an idea how to implement it, the KV Cache implementation from **HuggingFace** accepts either **(1)** a `None` value or **(2)** a KV Cache with a context size less than one than the one for the `token_ids`.

I've just done an experiment where I use the model to compute _(or complete)_ the KV Cache for the second sequence using the model to satisfy **(2)**. The function is called `complete_kv_cache_from_token_ids`, it's not implemented because it's slow.

2. _`SequenceState` to a prompt_

This will reinitialize everything.

##### Conditions under which KV Cache is saved:

1. The slice considers only one element `(batch_size_after_the_slice == 1, sample_size_after_the_slice == 1)`, slicing more than one element will reset the KV Cache. The condition includes the base case where `(batch_size == 1, sample_size == 1)`.

2. The slice starts from the first index for the prompt `(SequenceState[..., :M], SequenceState[..., 0:M])`.

There are some technical intricacies that don't allow saving KV Cache even under 1. and 2., see `[NOTE] [SPECIAL CASE]` flags in `token_level_slice_from_string_level_slice` utility.

It's also one of the reasons to not go wander to get KV Cache work if `batch_size > 1` and `num_samples > 1`. The tradeoff _complexity-usefulness_ seems just way off to me.

#### Using `list(SequenceState)`

`list(SequenceState)` allows to convert the `SequenceState` object into a list of strings.

# Exceptions

Three types of exceptions could be raised while using the `continuous` wrapper.

1. **SampleMismatch**: This is (1) raised when the sequence's samples are sliced, then thrown to the wrapper _(a mismatch between the number of samples in the sequence and the one in the generator)_ and (2) two sequences with different number of samples are added.

2. **BatchMismatch**: This is raised when two sequences of different batch sizes are added.

3. **SlicingError**: This is raised when the slice doesn't allow the KV Cache to be saved, it is handled through resetting the KV Cache.

# FLAGs

You will see multiple flags that I've put in the code comments:

- **[NOTE]**: Those are general notes, explaining how I approached some problems.

- **[QUESTION]**: Those are questions that I had when I was coding certain mechanisms.

- **[POTENTIAL BUG]**: Those are lines of code that could potentially trigger bugs.

# Modifications

1. `GenerationState` returns `attention_masks` as well.

2. `sequence_generator` takes `kv_cache` as a keyword argument with the value `None` as a default.

# `test_continuous.py`

Those are some tests I added for the different parts of the code.

PS: _I use the name `SequenceState` instead of `Sequence` just because it made the coding more obvious to me, tell me if you want to switch it back to `Sequence`._


rlouf:
This is cool, I really like the `continuous` function to a point I'd consider doing the same thing for `stream`. I'll review the PR soon, please bear with me.

PR #786, opened 2024-04-04
Add SGLang as a user of outlines

merrymercy:


rlouf:
Good idea, thank you!

PR #791, opened 2024-04-08
Fix missing text module from json documentation

ibndias:
Update the json function calling tutorial example that still use deprecated text module.


rlouf:
Thank you for spotting this and submitting a patch! Greatly appreciated 🙏

PR #798, opened 2024-04-11
Switched order of AzureAsyncOpenAI to AsyncAzureOpenAI

gfade:
Switched order of AzureAsyncOpenAI to AsyncAzureOpenAI

to match name in repo 

https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py


rlouf:
Good catch, thank you for contributing!

PR #803, opened 2024-04-11
Add cookbook to run Outlines on the cloud with Modal

rlouf:
Based on the [example notebook](https://github.com/modal-labs/modal-examples/blob/main/06_gpu_and_ml/llm-structured/outlines_generate.py) in Modal's example repository. I also added a script that can be run directly in the `/examples` folder.

### DIFF
--- a/docs/cookbook/deploy-using-modal.md
+++ b/docs/cookbook/deploy-using-modal.md
@@ -1,0 +1,126 @@
+# Run Outlines using Modal
+
+[Modal](https://modal.com/) is a serverless platform that allows you to easily run code on the cloud, including GPUs. It can come very handy for those of us who don't have a monster GPU at home and want to be able to quickly and easily provision, configure and orchestrate cloud infrastructure.
+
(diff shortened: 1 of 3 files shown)


charlesfrye (commented):
Charles from Modal here. This looks like a great demonstration of how Modal's platform and Outlines' tooling can work effectively together!

I have a few comments on framing, but you're free to take or leave them.

> and don't have experience or time to set up the necessary cloud infrastructure.

I would instead say "want to be able to quickly and easily provision, configure, and orchestrate cloud infrastructure". Even folks with experience often end up able to move more quickly with Modal, because we don't just simplify (no k8s ingress config, no certs), we also make things faster (faster container builds and boots, hot-reloading development servers).

> here an A100 with 80Gb memory

Should be GB.

> `modal token set`

charlesfrye:
And let me know if you have any questions!

rlouf:
Thank you so much @charlesfrye !

PR #814, opened 2024-04-13
Fix string format quotes

navster888:
The string format regexes setup in https://github.com/outlines-dev/outlines/pull/567 were not properly wrapped in quotes, resulting in invalid jsons being produced.

Updating the DATE_TIME, DATE, TIME, and UUID regexes to properly include `"` wrapping the values.

TODO: Adding additional tests that verify the produced regex is able to parse nested values, which is how this bug was originally noticed.  Will revisit later this afternoon, struggling to get my python environment configured correctly in order to run tests locally.


navster888:
@rlouf apologies for the delay was out of town this week. Still have not been able to get a working dev setup locally, however  given this change seems simple I'm wondering if we can rely on the checks in CI. Either way I believe this PR is ready for review, will give local setup another go once I'm back home tomorrow and see if I can figure things out in the meantime, cheers.

rlouf:
I updated the code in #831 and merged. Thank you for contributing!

PR #832, opened 2024-04-21
Convert vocabulary types and load model concurrently

### LINKED ISSUE #768: Speed up index construction by converting vocabulary types while loading the model
Because we use Numba to compile the index we need to [convert the vocabulary types](https://github.com/outlines-dev/outlines/blob/803439a83d15ea4fcc452e1c367ffb41feb39ac2/outlines/fsm/regex.py#L761), which takes a non-negligible amount of time every time the script is run. A simple way to go around this is to execute this function in a separate thread while model is being loaded. We may also be able to make Numba cache JIT-compiled function by compiling the index for a trivial regex.

rlouf:
Converting vocabulary types and JIT-compiling Numba functions can represent a substantial amount of the compile time for very simple regular expressions. In particular, type conversion needs to happen every time the code is run and takes a few seconds at each session. Here we perform these operations while the model weights are downloaded and loaded on GPU, thus removing some of the overhead associated with the index compilation.

Closes #768.

### DIFF
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -5,7 +5,6 @@ import interegular
 from lark import Lark
 
 from outlines import grammars
-from outlines.caching import cache
 from outlines.fsm.regex import (
     create_fsm_index_tokenizer,
     make_byte_level_fsm,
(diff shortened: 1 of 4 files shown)


rlouf:
A few API tweaks are necessary to implement this properly. First, the code that transforms a regex into an index should be decomposed in:

1. A function that takes a regex as an argument and returns a byte-level deterministic FSM;
2. A function that converts the vocabulary types
3. A function that takes the converted vocabulary, tokenizer and returns an index

We then initialize model wrappers (e.g. `outlines.models.vLLM`) with the model instance and the converted vocabulary. Initializing functions (e.g. `outlines.model.vllm`) accept a model name or a model class.

PR #836, opened 2024-04-25
Add cookbook to run Outlines with BentoML and BentoCloud

larme:


larme:
Sorry, need to correct some wording, close for now

PR #840, opened 2024-04-26
Add azure_openai to models __init__.py

lassiraa:
Currently, importing azure_openai as documented in https://outlines-dev.github.io/outlines/reference/models/openai/ does not work. Importing azure_openai as documented results in "AttributeError: module 'outlines.models' has no attribute 'azure_openai'"

Adding azure_openai to __init__.py fixes that behaviour. Currently, only way to import azure_openai is with `from outlines.models.openai import azure_openai`


rlouf:
Thank you for contributing!

PR #848, opened 2024-04-30
Add cookbook to run Outlines on the cloud with BentoML

larme:
Similar to #803, I made an example of running outlines with BentoML locally or in the cloud. Any feedback is welcomed. Thanks!


rlouf:
Thank you for contributing! You will need to run `pre-commit` locally and fix formatting issues before we can merge the PR.

PR #852, opened 2024-04-30
Remove the need to copy all tokens during basic generation

brandonwillard:
This change removes the need to copy the entire list of token IDs during basic generation.  (N.B. I'm not a fan of the use of `None` to indicate the full vocabulary, but it's simple and does the job.)

More can be done to improve this situation dramatically (e.g. return the original logits untouched, update the original logits array in-place using the disallowed token IDs in other cases, etc.), but a change like this is a good start.


rlouf:
Not a huge fan of `None` either, but I don't have a better alternative to suggest. Otherwise looks good.

Note that this is a breaking change for libraries downstream.

PR #863, opened 2024-05-04
Adds option for JSON schema optimization

leloykun:
Pydantic's `.model_json_schema()` and `get_schema_from_signature` don't actually make optional fields/arguments optional in the json schema. This forces the model to output the keys even when the values are `null` anyway--slowing down inference the larger the schema is & the more optional fields there is.

---

For example, for this Pydantic class:

```python
class Test(BaseModel):
    field_a: int
    field_b: Optional[int]
    field_c: None
```

`.model_json_schema()` builds this schema:

```json
{
    "properties": {
        "field_a": {"title": "Field A", "type": "integer"},
        "field_b": {
            "anyOf": [{"type": "integer"}, {"type": "null"}],
            "title": "Field B",
        },
        "field_c": {"title": "Field C", "type": "null"},
    },
    "required": ["field_a", "field_b", "field_c"],
    "title": "Test",
    "type": "object",
}
```

`optimize_schema` in this PR reduces this to:

```json
{
    "properties": {
        "field_a": {"title": "Field A", "type": "integer"},
        "field_b": {"title": "Field B", "type": "integer"},
    },
    "required": ["field_a"],
    "title": "Test",
    "type": "object",
}
```

---

Likewise, `get_schema_from_signature` converts this function:

```python
    if b is None:
        return a
    return a + b
```

to this schema:

```json
{
    "properties": {
        "a": {"title": "A", "type": "integer"},
        "b": {
            "anyOf": [{"type": "integer"}, {"type": "null"}],
            "title": "B",
        },
    },
    "required": ["a", "b"],
    "title": "Arguments",
    "type": "object",
}
```

`optimize_schema` reduces this to:

```json
{
    "properties": {
        "a": {"title": "A", "type": "integer"},
        "b": {"title": "B", "type": "integer"},
    },
    "required": ["a"],
    "title": "Arguments",
    "type": "object",
}
```

---

I decided to add a flag, `enable_schema_optimization`, and set it to `False` by default because it further restricts the support distribution and thus might break models finetuned without this setting.

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -97,7 +97,7 @@ def build_regex_from_schema(
     return to_regex(resolver, content, whitespace_pattern)
 
 
-def is_null_type(instance: dict):
+def _is_null_type(instance: dict):
     if "type" in instance and (instance["type"] == "null" or instance["type"] is None):
         return True
     if "const" in instance and (
@@ -107,9 +107,9 @@ def is_null_type(instance: dict):
     return False
 
 
-def any_of_list_has_null_type(any_of_list: list[dict[str, str]]):
-    for subinstance in any_of_list:
-        if is_null_type(subinstance):
+def _has_null_type(instance_list: list[dict]):
+    for instance in instance_list:
+        if _is_null_type(instance):
             return True
     return False
 
@@ -134,12 +134,10 @@ def optimize_schema(instance):
                     subinstance_type == "array" and subinstance.get("minItems", 0) == 0
                 ):
                     new_optional_keys.add(key)
-            elif "anyOf" in subinstance and any_of_list_has_null_type(
-                subinstance["anyOf"]
-            ):
+            elif "anyOf" in subinstance and _has_null_type(subinstance["anyOf"]):
                 any_of_list = subinstance.pop("anyOf")
                 filtered_any_of_list = list(
-                    filter(lambda d: is_null_type(d), any_of_list)
+                    filter(lambda d: _is_null_type(d), any_of_list)
                 )
                 if len(filtered_any_of_list) == 0:
                     keys_to_remove.add(key)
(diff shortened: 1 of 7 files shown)


eitanturok:
There seems to be another potential bug here. Given the function
```python
    if b is None:
        return a
    return a + b
```
the function `get_schema_from_signature` outputs `"title": "Arguments"` both when `optimize_schema` is used and when it is not used. It seems like the output should have `"title": "test_add"`.

Perhaps I should raise this in a separate issue.

leloykun:
@eitanturok I don't think this is a bug cuz we don't use the title field when building the FSM (& when generating outputs)

Can you provide an example where this breaks something?

eitanturok:
@leloykun

I'm using outlines to make my models better at function calling and this current setup causes me some issues.

At a high level, I take the generated schema and use it 1) for the system prompt and 2) to create a regex. I input this schema into the system prompt so it knows which functions it has access to. But if the json schema does NOT contain the function's name, the model won't know how to call it.

Here is an example:
```python

    if b is None:
        return a
    return a + b
    
schema_json = get_schema_from_signature(tool)
schema_str = json.dumps(schema_json).strip()
schema_regex = build_regex_from_schema(schema_str, whitespace_pattern)

system_prompt = f"You are an expert at function calling and have access to the following tools: {function_schema}."
system_prompt += "Please call one of these functions."
system_prompt = system_prompt.format(schema_str)

generator = generate.regex(model, schema_regex)
```

If the function name is not included in the schema generated from `get_schema_from_signature` then this causes issues to arise.

leloykun:
@eitanturok, I think we should raise this as a separate issue

I'm thinking of replacing this line in `get_schema_from_signature`
```python
model = create_model("Arguments", **arguments)
```
with
```python
model = create_model(fn.__name__, **arguments)
```
or
```python
model = create_model(fn.get("__name__", "Arguments"), **arguments)
```
just to be safer

eitanturok:
I was thinking the same thing. I'll raise this a separate issue.

eitanturok:
Raised the issue in #878. Future discussions should take place there.

PR #864, opened 2024-05-05
fix code rendering

eitanturok:
Make the code render as Python.


rlouf:
Thank you for spotting this and submitting a fix!!

PR #866, opened 2024-05-05
Ignore errors caused by import warnings from `huggingface_hub` & `pyairports`

leloykun:
`pytest` seems to be pretty strict with these warnings.

This PR ignores errors caused by import warnings from `huggingface_hub` & `pyairports`


rlouf:
Thank you for fixing this!

PR #873, opened 2024-05-07
Update create_states_mapping function to include tokenizer parameter

### LINKED ISSUE #872: State mapping cache ignores the tokenizer used to build the state machine

https://github.com/outlines-dev/outlines/blob/4f8433d8d6633b0780c3a6c27981f9adffbe49f5/outlines/fsm/guide.py#L115

The cached function actually depends on the regex **and** the tokenizer. The tokenizer is not a parameter of the function, though, which leads to cached state maps being shared across different tokenizers, which leads to errors.

```python
import outlines

regex = r"((25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(25[0-5]|2[0-4]\d|[01]?\d\d?)"

model = outlines.models.transformers("stabilityai/stablelm-2-zephyr-1_6b")

    model,
    regex,
)

print(structured)

model = outlines.models.transformers("microsoft/phi-2")
    model,
    regex,
)

print(structured)
```

### Expected result:

```shell
Both generations should conform to the regex.
```

### Error message:

_No response_

Version information
<details>

```
0.0.41
Python 3.11.4 (main, Nov 24 2023, 14:45:29) [Clang 15.0.0 (clang-1500.0.40.1)]
aiohttp==3.9.3
aiosignal==1.3.1
annotated-types==0.6.0
anyio==3.7.1
appdirs==1.4.4
appnope==0.1.4
argon2-cffi==23.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==2.4.1
async-lru==2.0.4
attrs==23.2.0
Babel==2.14.0
beautifulsoup4==4.11.1
bleach==6.1.0
boto3==1.24.53
botocore==1.27.96
certifi==2024.2.2
cffi==1.16.0
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==3.0.0
colorama==0.4.6
comm==0.2.1
contourpy==1.2.1
cycler==0.12.1
datasets==2.15.0
debugpy==1.8.1
decorator==5.1.1
defusedxml==0.7.1
dill==0.3.7
diskcache==5.6.3
distro==1.9.0
dnspython==2.5.0
easyocr==1.7.1
einops==0.7.0
environs==9.5.0
executing==2.0.1
faiss-cpu==1.7.4
fastjsonschema==2.19.1
filelock==3.13.1
fonttools==4.51.0
fqdn==1.5.1
frozendict==2.4.0
frozenlist==1.4.1
fsspec==2023.10.0
grpcio==1.56.0
h11==0.14.0
html5lib==1.1
httpcore==1.0.2
httpx==0.26.0
huggingface-hub==0.19.4
idna==3.6
imageio==2.34.1
interegular==0.3.3
ipykernel==6.29.2
ipython==8.21.0
ipywidgets==8.1.2
isoduration==20.11.0
jedi==0.19.1
Jinja2==3.1.3
jmespath==1.0.1
joblib==1.3.2
json5==0.9.14
jsonpointer==2.4
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
jupyter==1.0.0
jupyter-console==6.6.3
jupyter-events==0.9.0
jupyter-lsp==2.2.2
jupyter_client==8.6.0
jupyter_core==5.7.1
jupyter_server==2.12.5
jupyter_server_terminals==0.5.2
jupyterlab==4.0.12
jupyterlab_pygments==0.3.0
jupyterlab_server==2.25.2
jupyterlab_widgets==3.0.10
kiwisolver==1.4.5
lark==1.1.9
lazy_loader==0.4
llmware==0.2.3
llvmlite==0.42.0
lxml==4.9.3
MarkupSafe==2.1.5
marshmallow==3.20.2
matplotlib==3.8.4
matplotlib-inline==0.1.6
mistune==3.0.2
mpmath==1.3.0
multidict==6.0.5
multiprocess==0.70.15
multitasking==0.0.11
nbclient==0.9.0
nbconvert==7.16.0
nbformat==5.9.2
nest-asyncio==1.6.0
networkx==3.2.1
ninja==1.11.1.1
nltk==3.8.1
notebook==7.0.8
notebook_shim==0.2.3
numba==0.59.1
numpy==1.26.4
openai==1.12.0
opencv-python-headless==4.9.0.80
outlines==0.0.41
overrides==7.7.0
packaging==23.2
pandas==2.2.0
pandocfilters==1.5.1
parso==0.8.3
pdf2image==1.16.0
pexpect==4.9.0
pgvector==0.2.4
pillow==10.2.0
platformdirs==4.2.0
prometheus-client==0.19.0
prompt-toolkit==3.0.43
protobuf==4.25.2
psutil==5.9.8
psycopg==3.1.17
psycopg-binary==3.1.17
ptyprocess==0.7.0
pure-eval==0.2.2
pyarrow==15.0.0
pyarrow-hotfix==0.6
pyclipper==1.3.0.post5
pycparser==2.21
pydantic==2.6.1
pydantic_core==2.16.2
Pygments==2.17.2
pymilvus==2.3.0
pymongo==4.5.0
pyparsing==3.1.2
pytesseract==0.3.10
python-bidi==0.4.2
python-dateutil==2.8.2
python-dotenv==1.0.1
python-json-logger==2.0.7
pytz==2024.1
PyYAML==6.0.1
pyzmq==25.1.2
qtconsole==5.5.1
QtPy==2.4.1
referencing==0.33.0
regex==2023.12.25
requests==2.31.0
rfc3339-validator==0.1.4
rfc3986-validator==0.1.1
rpds-py==0.17.1
s3transfer==0.6.2
safetensors==0.4.2
# Editable install with no version control (sandbox==0.1.0)
-e /Users/breno/src/py/sandbox
scikit-image==0.23.2
scikit-learn==1.4.0
scipy==1.12.0
Send2Trash==1.8.2
sentence-transformers==2.2.2
sentencepiece==0.1.99
shapely==2.0.4
six==1.16.0
sniffio==1.3.0
soupsieve==2.5
sseclient-py==1.8.0
stack-data==0.6.3
sympy==1.12
tabulate==0.9.0
terminado==0.18.0
threadpoolctl==3.2.0
tifffile==2024.5.3
timm==0.9.16
tinycss2==1.2.1
tokenizers==0.19.1
torch==2.2.0
torchvision==0.17.0
tornado==6.4
tqdm==4.66.2
traitlets==5.14.1
transformers==4.40.2
types-python-dateutil==2.8.19.20240106
typing_extensions==4.9.0
tzdata==2024.1
ujson==5.9.0
uri-template==1.3.0
urllib3==1.26.18
wcwidth==0.2.13
webcolors==1.13
webencodings==0.5.1
websocket-client==1.7.0
Werkzeug==3.0.1
widgetsnbextension==4.0.10
Wikipedia-API==0.6.0
word2number==1.1
xxhash==3.4.1
yarl==1.9.4
yfinance==0.2.28
```
</details>

### Context for the issue:

_No response_

br3no:
Fixes #872

### DIFF
--- a/outlines/caching.py
+++ b/outlines/caching.py
@@ -52,7 +52,8 @@ def cache(key_function: Optional[Callable] = None):
     ----------
     key_function
       A callable function used to generate a unique key for each function call. It's
-      called with the arguments of the decorated function as arguments
+      called with the arguments of the decorated function as arguments and returns an
+      iterable of values that are used to create the cache key.
     Returns
     -------
       A decorator function that can be applied to other functions.
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -111,8 +111,21 @@ class RegexGuide(Guide):
     initial_state = 0
 
     def __init__(self, regex_string: str, tokenizer):
-        @cache()
-        def create_states_mapping(regex_string: str, tokenizer) -> Tuple[dict, set, set]:
+        @cache(
+            key_function=lambda regex_string, tokenizer: hash(
+                (
+                    regex_string,
+                    tokenizer.eos_token,
+                    tokenizer.eos_token_id,
+                    tokenizer.pad_token_id,
+                    tuple(sorted(tokenizer.vocabulary.items())),
+                    tuple(sorted(tokenizer.special_tokens)),
+                ),
+            )
+        )
+        def create_states_mapping(
+            regex_string: str, tokenizer: "Tokenizer"
+        ) -> Tuple[dict, set, set]:
             """Create the variables related to the mapping between states and tokens
             The parameters of the function are used for caching purpose
             """
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -112,15 +112,13 @@ class RegexGuide(Guide):
 
     def __init__(self, regex_string: str, tokenizer):
         @cache(
-            key_function=lambda regex_string, tokenizer: hash(
-                (
-                    regex_string,
-                    tokenizer.eos_token,
-                    tokenizer.eos_token_id,
-                    tokenizer.pad_token_id,
-                    tuple(sorted(tokenizer.vocabulary.items())),
-                    tuple(sorted(tokenizer.special_tokens)),
-                ),
+            key_function=lambda regex_string, tokenizer: (
+                regex_string,
+                tokenizer.eos_token,
+                tokenizer.eos_token_id,
+                tokenizer.pad_token_id,
+                tuple(sorted(tokenizer.vocabulary.items())),
+                tuple(sorted(tokenizer.special_tokens)),
             )
         )
         def create_states_mapping(
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -112,7 +112,7 @@ class RegexGuide(Guide):
 
     def __init__(self, regex_string: str, tokenizer):
         @cache()
-        def create_states_mapping(regex_string: str) -> Tuple[dict, set, set]:
+        def create_states_mapping(regex_string: str, tokenizer) -> Tuple[dict, set, set]:
             """Create the variables related to the mapping between states and tokens
             The parameters of the function are used for caching purpose
             """
@@ -142,7 +142,7 @@ class RegexGuide(Guide):
             self.states_to_token_maps,
             self.empty_token_ids,
             fsm_finals,
-        ) = create_states_mapping(regex_string)
+        ) = create_states_mapping(regex_string, tokenizer)
         self.eos_token_id = tokenizer.eos_token_id
         self.final_states = fsm_finals | {-1}
 
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+from unittest.mock import MagicMock, call
 
 import diskcache
 import pytest
@@ -41,6 +42,21 @@ def test_cache(refresh_environment):
         memory.clear()
 
 
+@pytest.fixture
+def test_cache_custom_key_function(refresh_environment):
+    """Initialize a temporary cache and delete it after the test has run."""
+    with tempfile.TemporaryDirectory() as tempdir:
+        os.environ["OUTLINES_CACHE_DIR"] = tempdir
+        import outlines
+
+        memory = outlines.get_cache()
+        assert memory.directory == tempdir
+
+        yield outlines.caching.cache(key_function=lambda x: (sorted(x.keys())))
+
+        memory.clear()
+
+
 def test_get_cache(test_cache):
     import outlines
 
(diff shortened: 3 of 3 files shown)


brandonwillard (commented):
I don't know how well the `tokenizer` instance will hash/serialize.  We need to at least make sure that there's a test confirming that we get cache hits when appropriate.

br3no:
The class inherits `Hashable`
 https://github.com/outlines-dev/outlines/blob/4f8433d8d6633b0780c3a6c27981f9adffbe49f5/outlines/models/tokenizer.py#L7

But yeah, a test would be good. Is there somewhere a documentation for contributors that could help me setup the dev env?

ekagra-ranjan:
Passing the tokenizer directly to cache key will not work since each instance of the same tokenizer object will be hashed to different values so there will be cache miss. I didnt see this PR earlier so I raised a PR with my local fix inn #876  which caches based on tokenizer name or path string.

br3no:
I've changed the PR to explicitly set the cache key computation using the `key_function` parameter of the `@cache()` decorator. This is in line with the discussion in #876.

@brandonwillard what do you think?

brandonwillard:
> I've changed the PR to explicitly set the cache key computation using the `key_function` parameter of the `@cache()` decorator. This is in line with the discussion in #876.
> 
> @brandonwillard what do you think?

We need tests that confirm that the hashing function works as intended.

br3no:
@brandonwillard I have added unit tests for the `cache` decorator when used in conjunction with the `key_function`. Let me know if this is sufficient for you.

brandonwillard (commented):
> @brandonwillard I have added unit tests for the `cache` decorator when used in conjunction with the `key_function`. Let me know if this is sufficient for you.

I like where you're going with those tests!  I was originally referring to tests for the choice of key/hash function for the tokenizers, though.  

We need to confirm that this choice of key will work between sessions and across different tokenizer types (at the very least the `transformer` types).  Since you're using all the members/fields of the generic `Tokenizer` interface, it's probably fine; if not, we need to update the interface and/or force subclasses to implement this key function themselves.

Speaking of which, we should first consider/try updating the `__hash__` and `__eq__` methods of `Tokenizer`to follow this key.  That would make things a lot more consistent; however, for the few times we need to compute a hash, it will be a little expensive.  Since the interface already implies some form of immutability via `Hashable`, and the tokenizers should be immutable in this scenario conceptually, we could reasonably cache the hash value and mostly remove that concern.

Anyway, if we fix `Tokenizer`, then we only need to add that extra argument to `create_states_mapping` and a direct test guaranteeing that the disk cache is hit for&mdash;say&mdash;an example using GPT2's tokenizer (at the very least).

br3no:
> We need to confirm that this choice of key will work between sessions and across different tokenizer types (at the very least the transformer types). Since you're using all the members/fields of the generic Tokenizer interface, it's probably fine; 

The `key_function` will return the fields that, together, "define" a particular tokenizer (these fields are all used in later code, so we can safely assume that it's okay to use them for this purpose). Assuming the hash is computed correctly, I believe we can also safely assume that the caching will work since we verified in a unit test, that the `key_function` is used the way it should.

> if not, we need to update the interface and/or force subclasses to implement this key function themselves.

I agree this would be the best solution, but this would require a large refactoring. The different integrations work differently with the tokenizers. The only implementation of the `Tokenizer` protocol class is https://github.com/outlines-dev/outlines/blob/97ec37d9038750101152582e5df3d7315b2759b5/outlines/models/transformers.py#L57 For vLLM the tokenizer is patched with the right methods and members. In this case, we would need to also patch the hashing function into the tokenizer. I think this is not something we should do.

> Speaking of which, we should first consider/try updating the `__hash__` and `__eq__` methods of Tokenizerto follow this key.

The situation described above is also the reason why this is, unfortunately, no solution at the moment.

I haven't looked into the matter in depth, but I believe there are some refactoring opportunities to consolidate the usage of tokenizers, aligned with what is done in the `transformers` integration. If all integrations would provide implementations of the `Tokenizer` protocol, then we could make the hash computation much simpler and also easier to test.

----

I opened the original issue because I realized that the cache uses only the regex as key, leading to errors because different tokenizers share the same state machine. Since the cache is persistent, I believe many users will face this issue. The symptoms of this problem are hard to read; the LLMs will generate gibberish. 

I think the right thing to do now would be to merge this PR to make sure people don't get into this problem. This is not a new feature, it's a bug. And then open a new issue to address the refactoring needs described above and maybe clean up the change introduced in this PR.

brandonwillard (commented):
I've created a PR (i.e. #911) that synthesizes the things we've talked about and outlines the kind of test we need before merging anything that closes #872.

The `outlines.caching.cache` tests here are useful, so we can update this PR so that it only introduces those changes (in one separate commit), or we can merge the changes from #911 into this PR if you want to finish that work here.

br3no:
@brandonwillard great work on #911. If I read it correctly, there is no longer a `key_function` argument in the `cache` decorator, right? So the tests here don't make sense if #911 is merged.

I don't mind closing this PR; I'd just like to have the behavior fixed. #911 seems like a better increment than this PR here.

PR #874, opened 2024-05-07
Remove broken final state loop

### LINKED ISSUE #856: Endless generation bug popped up during migration to `Guide` in vLLM integration

While updating the vLLM integration with outlines (cf. https://github.com/vllm-project/vllm/pull/4109) to move to the new `Guide` API a test started failing. The symptom is that a regular expression that matches IPv4 addresses does not stop generating.

The original regular expression was: `r"((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)"`
The issue can still be observed with the much simpler expression: `r"((25|(2|1\d|\d))\.){3}(25|(2|1\d|\d))"`

The issue was discussed in a call with @rlouf.

Things we tried, that didn't change the result:
* move to outlines 0.0.36
* change the `RegexGuide` code and remove the `byte_fsm` replacing it with `regex_pattern.to_fsm()`

```python
#
# To reproduce the error you need to pull the correct vllm fork, where the change to `Guides` is being worked on (https://github.com/br3no/vllm/tree/3715-outlines-guides) 
# 

import asyncio
import re
import pytest
from vllm.engine.arg_utils import AsyncEngineArgs
from vllm.engine.async_llm_engine import AsyncLLMEngine

from vllm.entrypoints.openai.protocol import CompletionRequest
from vllm.entrypoints.openai.serving_completion import OpenAIServingCompletion
from unittest.mock import Mock

pytestmark = pytest.mark.asyncio


async def test_guided_decoding_with_outlines():
    guided_regex = (r"((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.){3}"
              r"(25[0-5]|(2[0-4]|1\d|[1-9]|)\d)")

    args = AsyncEngineArgs(
        model = "HuggingFaceH4/zephyr-7b-beta",
        dtype = "bfloat16",
        max_model_len = 8192,
        enforce_eager = True,
        max_num_seqs = 128)
    
    engine = AsyncLLMEngine.from_engine_args(args)

    openai_serving_completion = OpenAIServingCompletion(
        engine, [args.model])
    
    raw_request_mock = Mock()
    future_false = asyncio.Future()
    future_false.set_result(False)
    raw_request_mock.is_disconnected = Mock(return_value=future_false)
    completion = await openai_serving_completion.create_completion(
        CompletionRequest(
            model=args.model,
            prompt=f"Give an example IPv4 address with this regex: {guided_regex}",
            n=1,
            temperature=1.0,
            max_tokens=20,
            guided_regex=guided_regex,
            guided_decoding_backend="outlines"
        ),
        raw_request_mock
    )

    for choice in completion.choices:
        assert len(choice.text) <= 15
        assert re.match(guided_regex, choice.text) is not None
```

### Expected result:

```shell
The test should not fail. Instead it fails generating IPs like this: `10.0.0.1282552550000`.
```

### Error message:

_No response_

Version information
<details>

```
0.0.39
Python 3.8.18 (default, Oct  2 2023, 15:02:11) 
[GCC 9.4.0]
accelerate==0.27.2
ai2-olmo==0.2.5
aiohttp==3.9.3
aioprometheus==23.12.0
aiosignal==1.3.1
annotated-types==0.6.0
antlr4-python3-runtime==4.9.3
anyio==4.2.0
async-timeout==4.0.3
attrs==23.2.0
awscli==1.32.83
boto3==1.34.83
botocore==1.34.83
cached_path==1.6.2
cachetools==5.3.3
certifi==2024.2.2
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==3.0.0
cmake==3.29.2
codespell==2.2.6
colorama==0.4.4
cupy-cuda12x==12.1.0
diskcache==5.6.3
distro==1.9.0
docutils==0.16
einops==0.7.0
exceptiongroup==1.2.0
fastapi==0.109.2
fastrlock==0.8.2
filelock==3.13.1
frozenlist==1.4.1
fsspec==2024.2.0
google-api-core==2.18.0
google-auth==2.29.0
google-cloud-core==2.4.1
google-cloud-storage==2.16.0
google-crc32c==1.5.0
google-resumable-media==2.7.0
googleapis-common-protos==1.63.0
h11==0.14.0
hiredis==2.3.2
httpcore==1.0.4
httptools==0.6.1
httpx==0.27.0
huggingface-hub==0.20.3
idna==3.6
importlib-resources==6.1.1
importlib_metadata==7.0.2
iniconfig==2.0.0
interegular==0.3.3
isort==5.13.2
Jinja2==3.1.3
jmespath==1.0.1
joblib==1.3.2
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
lark==1.1.9
libnacl==2.1.0
llvmlite==0.41.1
lm-format-enforcer==0.9.3
markdown-it-py==3.0.0
MarkupSafe==2.1.5
mdurl==0.1.2
mpmath==1.3.0
msgpack==1.0.7
multidict==6.0.5
mypy==1.9.0
mypy-extensions==1.0.0
nest-asyncio==1.6.0
networkx==3.1
ninja==1.11.1.1
numba==0.58.1
numpy==1.24.4
nvidia-cublas-cu12==12.1.3.1
nvidia-cuda-cupti-cu12==12.1.105
nvidia-cuda-nvrtc-cu12==12.1.105
nvidia-cuda-runtime-cu12==12.1.105
nvidia-cudnn-cu12==8.9.2.26
nvidia-cufft-cu12==11.0.2.54
nvidia-curand-cu12==10.3.2.106
nvidia-cusolver-cu12==11.4.5.107
nvidia-cusparse-cu12==12.1.0.106
nvidia-nccl-cu12==2.18.1
nvidia-nvjitlink-cu12==12.3.101
nvidia-nvtx-cu12==12.1.105
omegaconf==2.3.0
openai==1.13.3
orjson==3.9.13
outlines==0.0.39
packaging==23.2
peft==0.9.0
pillow==10.3.0
pkgutil_resolve_name==1.3.10
pluggy==1.4.0
prometheus_client==0.20.0
proto-plus==1.23.0
protobuf==4.25.2
psutil==5.9.8
py==1.11.0
py-cpuinfo==9.0.0
pyasn1==0.6.0
pyasn1_modules==0.4.0
pydantic==2.6.1
pydantic_core==2.16.2
Pygments==2.17.2
pynvml==11.5.0
pytest==8.0.2
pytest-asyncio==0.23.5
pytest-forked==1.6.0
pytest-rerunfailures==13.0
pytest-shard==0.1.2
python-dateutil==2.9.0.post0
python-dotenv==1.0.1
PyYAML==6.0.1
quantile-python==1.1
ray==2.9.2
redis==5.0.3
referencing==0.33.0
regex==2023.12.25
requests==2.31.0
rich==13.7.1
rpds-py==0.17.1
rsa==4.7.2
ruff==0.1.5
s3transfer==0.10.1
safetensors==0.4.2
scipy==1.10.1
sentencepiece==0.1.99
six==1.16.0
sniffio==1.3.0
starlette==0.36.3
sympy==1.12
tensorizer==2.9.0a0
tiktoken==0.6.0
tokenizers==0.19.1
toml==0.10.2
tomli==2.0.1
torch==2.1.2
tqdm==4.66.1
transformers==4.40.0
triton==2.1.0
types-PyYAML==6.0.12.12
types-requests==2.31.0.6
types-setuptools==69.1.0.20240308
types-urllib3==1.26.25.14
typing_extensions==4.9.0
urllib3==1.26.18
uvicorn==0.27.0.post1
uvloop==0.19.0
-e git+ssh://git@github.com/br3no/vllm.git@e6ffb1af2e904436473568f9d43131c998e55063#egg=vllm
watchfiles==0.21.0
websockets==12.0
xformers==0.0.23.post1
yapf==0.32.0
yarl==1.9.4
zipp==3.17.0
```
</details>

### Context for the issue:

_No response_

br3no:
Fixes #856

The code this PR removes introduces an artificial and erroneous loop transition in every final state that is always traversed, regardless of the generation.

The comment doesn't make sense in my opinion, as the `if` above just handles exactly this case.

Removing this piece of code fixes the bug that surfaced in the upgrade of outlines in the vLLM integration.

### DIFF
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -193,7 +193,7 @@ class RegexGuide(Guide):
         The new state of the guide.
 
         """
-        if token_id == self.eos_token_id:
+        if token_id == self.eos_token_id or state not in self.states_to_token_maps:
             return -1
 
         last_token_to_end_state = self.states_to_token_maps[state]
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -195,10 +195,6 @@ class RegexGuide(Guide):
         """
         if token_id == self.eos_token_id:
             return -1
-        elif (
-            state in self.final_states
-        ):  # Necessary because we keep generating EOS tokens when finished
-            return state
 
         last_token_to_end_state = self.states_to_token_maps[state]
         next_state = last_token_to_end_state.get(token_id)
--- a/tests/fsm/test_fsm.py
+++ b/tests/fsm/test_fsm.py
@@ -82,9 +82,7 @@ def test_regex_final_state():
     assert fsm.is_final_state(state)
 
     state = fsm.next_state(state=5, token_id=103)
-    assert state == 5
-
-    assert fsm.is_final_state(-1)
+    assert fsm.is_final_state(state)
 
 
 def test_cfg():
--- a/tests/fsm/test_guide.py
+++ b/tests/fsm/test_guide.py
@@ -180,9 +180,7 @@ def test_regex_final_state():
     assert fsm.is_final_state(state)
 
     state = fsm.get_next_state(state=5, token_id=103)
-    assert state == 5
-
-    assert fsm.is_final_state(-1)
+    assert fsm.is_final_state(state)
 
 
 def test_cfg():


br3no:
A related matter: https://github.com/outlines-dev/outlines/blob/4f8433d8d6633b0780c3a6c27981f9adffbe49f5/outlines/generate/generator.py#L94

This code is fundamentally broken, in my opinion, because it always stops generation when a final state is reached, regardless of outgoing transitions it may have. Instead, the condition for stopping should be that a stop-token has been generated. Right?

rlouf:
Is there a minimal reproducing example we could add as a test?

br3no:
The issue does not show in the `transformers` integration because of the line I posted in the last comment. So the most minimal example at the moment would be the code provided in https://github.com/outlines-dev/outlines/issues/856.

br3no:
I believe this should do it now.

br3no:
I've looked into the breaking tests:

```
____________________________ test_regex_final_state ____________________________

    def test_regex_final_state():
        """Make sure that the FSM stays in the final state as we keep generating"""
    
        class MockTokenizer:
            vocabulary = {"`": 101, ".": 102, "\n": 103, "eos": 104}
            special_tokens = {"eos"}
            eos_token_id = 104
    
            def convert_token_to_string(self, token):
                return token
    
        regex_str = r"`\n(\.\n)?`\n"
        tokenizer = MockTokenizer()
    
        with pytest.warns(UserWarning):
            fsm = RegexFSM(regex_str, tokenizer)
    
        state = fsm.next_state(state=4, token_id=103)
        assert state == 5
        assert fsm.is_final_state(state)
    
        state = fsm.next_state(state=5, token_id=103)
>       assert state == 5
E       assert -1 == 5

tests/fsm/test_fsm.py:85: AssertionError
____________________________ test_regex_final_state ____________________________

    def test_regex_final_state():
        """Make sure that the FSM stays in the final state as we keep generating"""
    
        class MockTokenizer:
            vocabulary = {"`": 101, ".": 102, "\n": 103, "eos": 104}
            special_tokens = {"eos"}
            eos_token_id = 104
    
            def convert_token_to_string(self, token):
                return token
    
        regex_str = r"`\n(\.\n)?`\n"
        tokenizer = MockTokenizer()
        fsm = RegexGuide(regex_str, tokenizer)
    
        state = fsm.get_next_state(state=4, token_id=103)
        assert state == 5
        assert fsm.is_final_state(state)
    
        state = fsm.get_next_state(state=5, token_id=103)
>       assert state == 5
E       assert -1 == 5

tests/fsm/test_guide.py:183: AssertionError
```

I believe these tests are testing an assumption that is fundamentally wrong. Final states **can** have outbound transitions, including into non-terminal states.
I have attached an svg file with a rendering of the state machine described by the state to token map for the following very simple regular expression:

```
"(12){1,3}"
```
state_machine

---

I'm wondering if the right thing to do would be to remove these tests, or what we would want to test instead.

br3no:
I have changed the tests to verify that if we are in a final state with no outbound transitions, a new generation will lead to us staying in a final state.

lapp0 (commented):
Basically the same as https://github.com/outlines-dev/outlines/pull/884

Should we leave `test_fsm.py` as is? Otherwise looks good.

br3no:
> Basically the same as #884
> 
> Should we leave `test_fsm.py` as is? Otherwise looks good.

Do you mean reverting the changes in `test_fsm.py`? If so, this will break the build. In essence both `test_fsm.py` and `test_guide.py` are testing the same thing, since they share the underlying implementation.

lapp0:
Thanks, looks good to me!

ekagra-ranjan:
Damn! I was facing this issue on Fri and spent a couple of days to figure the solution only to find that a PR was merged recently :)

ekagra-ranjan:
@br3no can you pls share how did you generate the FSM plot here? https://github.com/outlines-dev/outlines/pull/874#issuecomment-2100235459

ekagra-ranjan:
Shouldn;t `state` be always present in `self.states_to_token_maps`? If nor then we should throw an error or letting it pass silently?

outlines/fsm/guide.py
```diff
@@ -193,12 +193,8 @@ def get_next_state(self, state: int, token_id: int) -> int:
         The new state of the guide.
 
         """
-        if token_id == self.eos_token_id:
+        if token_id == self.eos_token_id or state not in self.states_to_token_maps:
```
ekagra-ranjan, line 196: Shouldn;t `state` be always present in `self.states_to_token_maps`? If not then we should throw an error or letting it pass silently?

br3no:
@ekagra-ranjan sure. I used graphviz for that. Here's an example:

```python
import outlines
from graphviz import Digraph

def draw_state_machine(graph: dict, final_states: set, tokenizer):
    dot = Digraph()

    # Add nodes
    for state in graph:
        if state in final_states:
            dot.node(str(state), str(state), color='salmon', style='filled', fillcolor='salmon')
        else:
            dot.node(str(state), str(state), color='lightblue', style='filled', fillcolor='lightblue')

    # Prepare edge labels by aggregating transitions between the same nodes
    edge_labels = {}
    for state, transitions in graph.items():
        for transition, end_state in transitions.items():
            if end_state not in graph:
                # Add end states not in the state map
                dot.node(str(end_state), str(end_state), color='salmon', style='filled', fillcolor='salmon')
            label = tokenizer.decode(int(transition))
            edge_key = (str(state), str(end_state))
            if edge_key not in edge_labels:
                edge_labels[edge_key] = label
            else:
                # Append new label to existing label, separated by a comma
                edge_labels[edge_key] += ", " + label

    # Add edges with aggregated labels
    for (start_state, end_state), label in edge_labels.items():
        dot.edge(start_state, end_state, label=label)

    # Render and view the graph
    dot.render('state_machine', view=True, format='svg')

tokenizer_zephyr = AutoTokenizer.from_pretrained("stabilityai/stablelm-2-zephyr-1_6b")

regex = r"(12){1,2}"

model = outlines.models.transformers("stabilityai/stablelm-2-zephyr-1_6b")

generator_zephyr = outlines.generate.regex(
    model,
    regex,
)

draw_state_machine(generator_zephyr.fsm.states_to_token_maps, generator_zephyr.fsm.final_states, tokenizer_zephyr)
```

outlines/fsm/guide.py
ekagra-ranjan, line 196: @br3no  I was wondering if we really need the 2nd condition `state not in self.states_to_token_maps` ? The condition basically checks for states which do not have outgoing edges which would be a part of final states in the FSM. [This block of code](https://github.com/outlines-dev/outlines/blob/7863f8e8bbaeb71c9d2434636a2d63bfe6dd7d39/outlines/fsm/regex.py#L835-L841) add EOS as an edge to such states which makes them have atleast one outgoing edges and no states in FSM will be absent in the `states_to_token_maps`. Wdyt?

br3no: @ekagra-ranjan yes, we do need this second condition.

The block of code you linked to does not add an EOS outbound transition to these states. It only adds transitions to final states which are present in `states_to_token_maps`. But these states are not present there. `states_to_token_subsets.get(state)` will return `None` for these states.

I'm not really knowledgeable about the way Outlines (and interegular) build the state machines out of regexes. The matter of fact is that the `states_to_token_maps` does **not** contain all states that are reachable. I have noticed this while debugging the code for some example regexes. 

This is not a problem in principle, as these states are considered to be final and `states_to_token_subsets.get(state) is None` is used all over the code to handle this special case (as in the block you linked to). 

I actually believe this could be improved and Outlines would profit from removing this special case that needs to be thought of all over the place and could lead to bugs. But this is, as I said, not a problem in principle.

PR #875, opened 2024-05-07
Use actual function name in `get_schema_from_signature`

leloykun:
As discussed with @eitanturok in https://github.com/outlines-dev/outlines/pull/863

---

Currently, `get_schema_from_signature` converts the following function:
```python
    if b is None:
        return a
    return a + b
```
to the schema:
```python
{
    "properties": {
        "a": {"title": "A", "type": "integer"},
        "b": {"title": "B", "type": "integer"},
    },
    "required": ["a"],
    "title": "Arguments",
    "type": "object",
}
```
The issue is that the title of the schema is always "Arguments", regardless of the name of the function. And in some cases, this breaks the performance of the model as discussed by @eitanturok

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -375,6 +375,10 @@ def get_schema_from_signature(fn: Callable) -> str:
         else:
             arguments[name] = (arg.annotation, ...)
 
-    model = create_model("Arguments", **arguments)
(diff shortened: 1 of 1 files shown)


leloykun:
now a duplicate of https://github.com/outlines-dev/outlines/pull/875

PR #876, opened 2024-05-07
Add tokenizer to cache key

ekagra-ranjan:
Fixes: #872

### DIFF
--- a/outlines/integrations/utils.py
+++ b/outlines/integrations/utils.py
@@ -52,6 +52,17 @@ def adapt_tokenizer(tokenizer: PreTrainedTokenizerBase) -> PreTrainedTokenizerBa
     tokenizer.vocabulary = tokenizer.get_vocab()
     tokenizer.special_tokens = set(tokenizer.all_special_tokens)
 
+    # tokenizer.name_or_path is one of the keys for `create_states_mapping()` cache
+    if hasattr(tokenizer, "name_or_path"):
+        tokenizer.name_or_path = tokenizer.name_or_path
+    elif hasattr(tokenizer, "name"):
+        tokenizer.name_or_path = tokenizer.name
+    elif hasattr(tokenizer, "path"):
+        tokenizer.name_or_path = tokenizer.path
+    else:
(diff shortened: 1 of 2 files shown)


outlines/integrations/utils.py
```diff
@@ -52,6 +52,17 @@ def adapt_tokenizer(tokenizer: PreTrainedTokenizerBase) -> PreTrainedTokenizerBa
...
 
+    # tokenizer.name_or_path is one of the keys for `create_states_mapping()` cache
+    if hasattr(tokenizer, "name_or_path"):
+        tokenizer.name_or_path = tokenizer.name_or_path
+    elif hasattr(tokenizer, "name"):
+        tokenizer.name_or_path = tokenizer.name
+    elif hasattr(tokenizer, "path"):
+        tokenizer.name_or_path = tokenizer.path
+    else:
+        print("[WARNING]: tokenizer has no name or path. Setting cache key to `unknown`")
```
brandonwillard, line 63: Use `warnings`.

outlines/integrations/utils.py
```diff
@@ -52,6 +52,17 @@ def adapt_tokenizer(tokenizer: PreTrainedTokenizerBase) -> PreTrainedTokenizerBa
     tokenizer.vocabulary = tokenizer.get_vocab()
     tokenizer.special_tokens = set(tokenizer.all_special_tokens)
 
+    # tokenizer.name_or_path is one of the keys for `create_states_mapping()` cache
+    if hasattr(tokenizer, "name_or_path"):
```
brandonwillard, line 56: To what do all these cases correspond?  Regardless, we'll need tests for them.

brandonwillard (changes requested):

br3no:
@ekagra-ranjan, nice.

Would an alternative be to devise a function that will compute a unique hash given the current class members? This would not require changing the API.

brandonwillard:
> Would an alternative be to devise a function that will compute a unique hash given the current class members? This would not require changing the API.

Yes, that would be much better.

brandonwillard:
Closing this since it was covered by #911, which also required fixes unrelated to #872.

PR #878, opened 2024-05-08
Extract function name in `get_schema_from_signature`

eitanturok:
Currently, `get_schema_from_signature` creates a json where `"title": "Arguments"`. This PR changes this so that the `"title"` will instead be mapped to the function's name.

Here is an example. I have this code snippet:
```python
from outlines.fsm.json_schema import get_schema_from_signature

    if b is None:
        return a
    return a + b

schema = get_schema_from_signature(test_add)
print(schema)
```
Before this PR, this outputted
```python
{'properties': {'a': {'title': 'A', 'type': 'integer'}, 'b': {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'title': 'B'}}, 'required': ['a', 'b'], 'title': 'Arguments', 'type': 'object'}
```
and after this PR, this outputted
```python
{'properties': {'a': {'title': 'A', 'type': 'integer'}, 'b': {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'title': 'B'}}, 'required': ['a', 'b'], 'title': 'test_add', 'type': 'object'}
```
(Scroll to see the `'title'` key.)

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import re
+import warnings
 from typing import Callable, Optional
 
 from jsonschema.protocols import Validator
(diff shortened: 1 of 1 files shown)


eitanturok:
Initial discussion occurred in #863.

eitanturok:
In this PR, I ran

```python
from outlines.fsm.json_schema import get_schema_from_signature

    if b is None:
        return a
    return a + b

schema = get_schema_from_signature(test_add)
print(schema)
```
and got the output
```python
{'properties': {'a': {'title': 'A', 'type': 'integer'}, 'b': {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'title': 'B'}}, 'required': ['a', 'b'], 'title': 'test_add', 'type': 'object'}
```

Notice that we have `'title': 'test_add'` which is the desired behavior.

leloykun:
ah, I also raised a (now duplicate) PR 😅 

https://github.com/outlines-dev/outlines/pull/875

I'll close mine

Issue #853, opened 2024-04-30
Disable regex caching in tests


brandonwillard:
### Describe the issue as clearly as possible:

Tests like [`test_regex_vocabulary_error`](https://github.com/outlines-dev/outlines/blob/4934425cc511b7f7fcfff76744edb8f1452feb8f/tests/fsm/test_guide.py#L27) can fail due to the caching of [`create_states_mapping`](https://github.com/outlines-dev/outlines/blob/4934425cc511b7f7fcfff76744edb8f1452feb8f/outlines/fsm/guide.py#L114).

Caching should probably be disabled somewhat generally in those test modules (except for tests that are specifically for caching, of course).

### Steps/code to reproduce the bug:

```python
NA
```

### Expected result:

```shell
NA
```

### Outlines/Python version information:

4934425cc511b7f7fcfff76744edb8f1452feb8f

lapp0:
@brandonwillard could you please ensure https://github.com/outlines-dev/outlines/pull/886 resolves the issue on your end or provide reproduction steps?

brandonwillard:
> @brandonwillard could you please ensure #886 resolves the issue on your end or provide reproduction steps?

There's no need for that; we shouldn't be caching the operations we're testing when we aren't testing the caching itself.

PR #886, opened 2024-05-10
disable cache for all tests except `test_cache.py` and benchmark tests

lapp0:
Fixes https://github.com/outlines-dev/outlines/issues/853

### DIFF
--- a/outlines/caching.py
+++ b/outlines/caching.py
@@ -7,7 +7,10 @@ from typing import Callable, Optional
 import cloudpickle
 from diskcache import Cache
 
-_caching_enabled = True
+if os.environ.get("OUTLINES_DISABLE_CACHE", "") in ("", "0"):
+    _caching_enabled = True
+else:
+    _caching_enabled = False
 
 
 @functools.lru_cache(1)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,8 @@
+import os
+
+import pytest
+
+
+@pytest.fixture(scope="session", autouse=True)
+def set_env():
+    os.environ["OUTLINES_DISABLE_CACHE"] = "1"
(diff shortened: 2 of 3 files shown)


outlines/caching.py
```diff
@@ -7,7 +7,10 @@
 import cloudpickle
 from diskcache import Cache
 
-_caching_enabled = True
+if os.environ.get("OUTLINES_DISABLE_CACHE", "") in ("", "0"):
```
brandonwillard, line 10: Why do we need an environment variable for this?  

Also, can't we use the existing `disable_cache` function?

lapp0: It may be beneficial to introduce an environment variable that allows users disable cache, even outside of pytest.

I could also simply make the `disable_cache` fixure patch this variable and have no changes to `caching.py` if we don't want to expose this.

brandonwillard (commented):
To move this forward, we need a description of the approach and some justification(s) for why we should (or need to) use it over the alternatives (e.g. see the approaches in the test suite).

lapp0:
We need a general solution that is applied to all test cases in `tests/conftest.py`. This will ensure new test cases don't accidentally reintroduce the issue you highlighted.

I agree about the environment variable being a problem. The cleanest solution is to clear the cache before each test run. This won't require any special handling for benchmarks or `test_cache.py`

brandonwillard:
> We need a general solution that is applied to all test cases in `tests/conftest.py`. This will ensure new test cases don't accidentally reintroduce the issue you highlighted.
> 
> I agree about the environment variable being a problem. The cleanest solution is to clear the cache before each individual test run. This won't require any special handling for benchmarks or `test_cache.py`

Clearing the cache is another necessarily "global" approach with similar drawbacks.

Ideally, we would have something that works, still allows the tests to be run in parallel, and introduces a minimal amount of new code/logic (even if said code can be multi-purposed).

lapp0:
Thanks for reviewing. Indeed, we want to ensure the code remains robust. I'll close this for now.

PR #888, opened 2024-05-10
change int regex

eitanturok:
I also noticed that we have two different definitions of `INTEGER`. In `json_schema.py` we have 
```python
INTEGER = r"(-)?(0|[1-9][0-9]*)"
```
and in `types.py` we have
```python
INTEGER = r"[+-]?(0|[1-9][0-9]*)"
```
I changed the definition in `json_schema.py` to be the same as that in `types.py` because `INTEGER` in `types.py` allows both `+` and `-` in front of the number and the definition in `json_schema.py` does not allow this.

More generally, it may be helpful to store these types in one single location. Perhaps we should keep all the types in `types.py` and we can import these types into `json_schema.py` when we need it. That way we won't have diverging definitions again.

What do you think?

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -13,6 +13,7 @@ from referencing.jsonschema import DRAFT202012
 STRING_INNER = r'([^("\\\x00-\x1f\x7f-\x9f)]|\\\\)'
 STRING = f'"{STRING_INNER}*"'
 INTEGER = r"(-)?(0|[1-9][0-9]*)"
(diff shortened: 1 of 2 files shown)


rlouf:
That is because Python and JSON have different definitions of what counts as a valid integer. Check [the JSON specification](https://www.json.org/json-en.html).

eitanturok:
Oh interesting, I did not know that.

I wish the definition of an integer would be consistent. Shaking my head... 

Okay, I'm going to close this PR then.

Issue #820, opened 2024-04-15
Encountering `RuntimeError: Cannot convert token ` �` (29333) to bytes:  �` for some model vocabularies when using llama.cpp


abetlen:
### Describe the issue as clearly as possible:

Ran into this tokenizer issue in [`regex.py`](https://github.com/outlines-dev/outlines/blob/0bb6d1d3fcb358797e8d3d869acbfd167749b3b2/outlines/fsm/regex.py#L793) with some models (`Qwen1.5`, `Phi-2`) but not others (`OpenHermes-2.5-Mistral-7B`) when using llama.cpp.

This honestly might be my fault with something I'm doing in the `llama-cpp-python` but I'm not familiar enough with the outlines codebase to tell.

### Steps/code to reproduce the bug:

```python
from outlines import models, generate

model = models.llamacpp("Qwen/Qwen1.5-0.5B-Chat-GGUF", "*q8*.gguf")
generator = generate.choice(model, ["skirt", "dress", "pen", "jacket"])
answer = generator("Pick the odd word out: skirt, dress, pen, jacket")

print(answer)
```
```


### Expected result:

```shell
`skirt`
```

### Error message:

```shell
Exception: Cannot convert token ` �` (29333) to bytes:  �
Traceback (most recent call last):
  File "/home/andrei/Documents/llms/llama_cpp/server/errors.py", line 171, in custom_route_handler
    response = await original_route_handler(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andrei/Documents/llms/.direnv/python-3.11.9+/lib/python3.11/site-packages/fastapi/routing.py", line 278, in app
    raw_response = await run_endpoint_function(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andrei/Documents/llms/.direnv/python-3.11.9+/lib/python3.11/site-packages/fastapi/routing.py", line 191, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andrei/Documents/llms/llama_cpp/server/app.py", line 451, in create_chat_completion
    ] = await run_in_threadpool(llama.create_chat_completion, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #890, opened 2024-05-12
[Issue #820] Cannot convert token � (29333) to bytes: � for some model vocabularies when using llama.cpp

desaxce:
Attempt at solving issue #820 - RuntimeError: Cannot convert token � (29333) to bytes: �

With additional characters in the `gpt2_bytes_to_unicode()` map, I don't get errors anymore.
However, it doesn't seem good to add ad-hoc characters like that. We should test on many more kinds of tokenizers.

Benchmarks pass:
image

Side note: Had to switch to Linux for development because of the `vllm` dependency, see their [Requirements](https://docs.vllm.ai/en/latest/getting_started/installation.html#requirements).

### DIFF
--- a/outlines/fsm/regex.py
+++ b/outlines/fsm/regex.py
@@ -726,8 +726,8 @@ re_replacement_seq = re.compile(r"^▁*�+$")
 @lru_cache()
 def gpt2_bytes_to_unicode():
     """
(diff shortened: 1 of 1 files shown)


rlouf:
Thank you for opening a PR! Could we add a small test (maybe with the llama.cpp integration tests) that fails on `main` but passes with this change?

rlouf:
This issue was addressed  in #892, closing. Thank you for opening a PR!

Issue #456, opened 2023-12-20
Handle regexps with lookaround


benlipkin:
### What behavior of the library made you think about the improvement?

Outlines currently uses `interegular` to compile regexps to FSMs, but not all features of python regex are supported, in particular lookarounds.

I ran into this issue when trying to test the following grammar:
```
json_grammar = r"""
    ?start: value

    ?value: object
          | array
          | string
          | SIGNED_NUMBER      -> number
          | "true"             -> true
          | "false"            -> false
          | "null"             -> null

    array  : "[" [value ("," value)*] "]"
    object : "{" [pair ("," pair)*] "}"
    pair   : string ":" value

    string : ESCAPED_STRING

    %import common.ESCAPED_STRING
    %import common.SIGNED_NUMBER
    %import common.WS

    %ignore WS
"""
```

In the `CFGFSM` constructor, terminal patterns are compiled to regexps so that they can later be used to initialize `RegexFSM` proposals, e.g., `RegexFSM(terminal.pattern.to_regexp())`. When the supplied regex includes lookarounds, we get an error such as the following:
```
interegular.patterns.Unsupported: lookbacks are not implemented
```

### How would you like it to behave?

It would seem there are at least a couple options here:
1. Work with `interegular` to expand support to lookarounds. This is on the TODOs outlined in their [README](https://github.com/MegaIng/interegular), and perhaps could use a push.
2. Compile regex with lookaround to regex without lookaround? In theory, lookaround does not change the expressive power and this is all closed within regular languages. However, I do not know if any (efficient) algorithm exists to make this conversion (and in some cases the resulting regexp might be quite complex). Maybe the authors know more about this.

brandonwillard:
Yeah, we could at the very least offer replacements for the `common.*` definitions that don't unnecessarily use look-arounds and other non-regular features.

lapp0:
I don't see any non-regular examples in `common.lark` other than `ESCAPED_STRING` and `C_COMMENT`. https://github.com/lark-parser/lark/blob/ab11ead8316978d1d36f35adb6fc88bdae43c2c4/lark/grammars/common.lark

lapp0:
`interegular` uses a finite state machine which cannot handle non-regular features such as look-arounds. 

We could convert non-regular lark terminals to lark rules with multiple regular lark terminals, but this seems like overkill. I think we should just document the fact that "Outlines-lark" is a subset of lark which disallows non-regular terminals.

[...]

PR #897, opened 2024-05-17
Prevent Illegal Look-Around for OneOf in JSONSchema

### LINKED ISSUE #823: Seems like passing a discriminated union crashes

Seems like discriminated unions crashes outlines

```python
class Cat(BaseModel):
    pet_type: Literal['cat']
    meows: int


class Dog(BaseModel):
    pet_type: Literal['dog']
    barks: float


class Model(BaseModel):
    pet: Union[Cat, Dog] = Field(..., discriminator='pet_type')
    n: int

json_schema = Model.schema_json()

# and then use the schema with a vLLM endpoint
```
```


### Expected result:

```shell
To be able to pass a discriminated union
```

### Error message:

```shell
InternalServerError: Internal Server Error
```

Version information
<details>
```
(command output here)
```
</details>

### Context for the issue:

This would allow for multiple tool calling with reliability, because JSON mode is not enough and Json Schema only allwo

lapp0:
Fixes #823

This comment details the issues error: https://github.com/outlines-dev/outlines/issues/823#issuecomment-2116490949

The reproduction code provided results in a json schema with `OneOf[pets]`:

```
class Model(BaseModel):
    pet: Union[Cat, Dog] = Field(..., discriminator='pet_type')
```

Before this PR: `OneOf` uses negative lookaheads to assert that only one schema member is included. This is illegal in `interegular`, more details available here: https://github.com/outlines-dev/outlines/issues/456

After `OneOf` uses or-joined non-capturing groups which don't have the same issues with `interegular`.

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -195,13 +195,7 @@ def to_regex(
             to_regex(resolver, t, whitespace_pattern) for t in instance["oneOf"]
         ]
 
-        xor_patterns = []
-        # json schema validation ensured there is no overlapping schemas in oneOf
-        for subregex in subregexes:
-            other_subregexes = filter(lambda r: r != subregex, subregexes)
-            other_subregexes_str = "|".join([f"{s}" for s in other_subregexes])
-            negative_lookahead = f"(?!.*({other_subregexes_str}))"
-            xor_patterns.append(f"({subregex}){negative_lookahead}")
+        xor_patterns = [f"(?:{subregex})" for subregex in subregexes]
 
         return rf"({'|'.join(xor_patterns)})"
 
--- a/tests/fsm/test_json_schema.py
+++ b/tests/fsm/test_json_schema.py
@@ -1,9 +1,10 @@
 import json
 import re
-from typing import List
+from typing import List, Literal, Union
 
+import interegular
 import pytest
-from pydantic import BaseModel, constr
+from pydantic import BaseModel, Field, constr
 
 from outlines.fsm.json_schema import (
     BOOLEAN,
@@ -321,7 +322,7 @@ def test_match_number(pattern, does_match):
                 "title": "Foo",
                 "oneOf": [{"type": "string"}, {"type": "number"}, {"type": "boolean"}],
             },
-            rf"(({STRING})(?!.*({NUMBER}|{BOOLEAN}))|({NUMBER})(?!.*({STRING}|{BOOLEAN}))|({BOOLEAN})(?!.*({STRING}|{NUMBER})))",
+            rf'((?:"{STRING_INNER}*")|(?:{NUMBER})|(?:{BOOLEAN}))',
             [
                 ("12.3", True),
                 ("true", True),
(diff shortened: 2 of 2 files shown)


rlouf:
Looks good, thank you for the fix!

outlines/fsm/json_schema.py
```diff
@@ -195,13 +195,7 @@ def to_regex(
...
         ]
 
-        xor_patterns = []
-        # json schema validation ensured there is no overlapping schemas in oneOf
-        for subregex in subregexes:
-            other_subregexes = filter(lambda r: r != subregex, subregexes)
-            other_subregexes_str = "|".join([f"{s}" for s in other_subregexes])
-            negative_lookahead = f"(?!.*({other_subregexes_str}))"
-            xor_patterns.append(f"({subregex}){negative_lookahead}")
+        xor_patterns = [f"(?:{subregex})" for subregex in subregexes]
```
ekagra-ranjan, line 198: @lapp0 what is the difference bw oneOf and anyOf in this script? Both concat the subregex with `|`

ekagra-ranjan: Do you also think additional conditions like just checking for `^` and just checking for `$` needs to be added to this conditon? https://github.com/outlines-dev/outlines/blob/7863f8e8bbaeb71c9d2434636a2d63bfe6dd7d39/outlines/fsm/json_schema.py#L248-L249

ekagra-ranjan: And how about adding `array` along with `object` to the types to ignore [here](https://github.com/outlines-dev/outlines/blob/7863f8e8bbaeb71c9d2434636a2d63bfe6dd7d39/outlines/fsm/json_schema.py#L344-L347)? Array like object requires many more parameters like `items` which is not provided in the list of types which is the reason why object is ignored I believe since properties cannot be mentioned when type is given as a list

lapp0: @ekagra-ranjan actually reviewing the test cases, I think `anyOf` is wrong. I'd appreciate a second set of eyes on this. Would you agree that 

`oneOf` is correct (uses exclusive or), guarantees only one match

```
>>> print(re.fullmatch(r"((?:foo)|(?:bar))", "foo"))
<re.Match object; span=(0, 3), match='foo'>
>>> print(re.fullmatch(r"((?:foo)|(?:bar))", "foobar"))
None
```

`anyOf` is incorrect, allows only one match

```
>>> print(re.fullmatch(r"((foo)|(bar))", "foo"))
<re.Match object; span=(0, 3), match='foo'>
>>> print(re.fullmatch(r"((foo)|(bar))", "foobar"))
None
```

`allOf` is **somewhat** correct, requires generation of all schemas, but enforces an order which it shouldn't:

```
    elif "allOf" in instance:
        subregexes = [
            to_regex(resolver, t, whitespace_pattern) for t in instance["allOf"]
        ]
        subregexes_str = [f"{subregex}" for subregex in subregexes]
        return rf"({''.join(subregexes_str)})"
```

lapp0: For your other two comments, I appreciate your insights into json schema  -> regex. It's definitely a work in progress and it's great that you're looking deep into it's behavior and identifying problems. I think it's out of scope with this PR. Could you create a separate issue so others can provide input as well?

PR #902, opened 2024-05-17
add args to prompt

eitanturok:
In the outlines docs, we have the example
```python
import outlines

def my_tool(arg1: str, arg2: int):
    """Tool description.

    The rest of the docstring
    """
    pass

@outlines.prompt
def tool_prompt(question, tool):
    """{{ question }}

    COMMANDS
    1. {{ tool | name }}: {{ tool | description }}, args: {{ tool | args }}

    {{ tool | source }}
    """

prompt = tool_prompt("Can you do something?", my_tool)
print(prompt)
```
However, when I tried running this code, it did not work because the `args` filter used in `{{ tool | args }}` was not implemented. I implemented the `args` filter so now this example works.

Now the args filter will output all of the arguments with the type annotations and default values (if they are provided).
Example:
```python
from typing import List

def foo(x, y: str, z: List[int]=[1, 2, 3]):
    pass

@outlines.prompt
def tool_prompt(fn):
    """My args: {{ fn | args }}"""

prompt = tool_prompt(foo)
print(prompt)
```
which outputs
```python
My args: x, y: str, z: List[int] = [1, 2, 3]
```

### DIFF
--- a/docs/reference/prompting.md
+++ b/docs/reference/prompting.md
@@ -223,7 +223,7 @@ Several projects (e.g.[Toolformer](https://arxiv.org/abs/2302.04761), [ViperGPT]
     Can you do something?
 
     COMMANDS
-    1. my_tool: Tool description, args: arg1:str, arg2:int
+    1. my_tool: Tool description., args: arg1: str, arg2: int
 
     def my_tool(arg1: str, arg2: int):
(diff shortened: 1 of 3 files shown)


rlouf:
Thank you so much for contributing!

PR #903, opened 2024-05-18
Introduce PR Benchmark Workflow

### LINKED ISSUE #883: Check benchmarks in CI
We need an Actions step that automatically compares benchmark results against current `main` and notifies us when there is a significant regression *before* a PR is merged.

We've used [`github-actions-benchmark`](https://github.com/benchmark-action/github-action-benchmark) in the past, but it seemed overly complicated, was more about serving a website with benchmark results (which is complexity we don't need right now), and wasn't actually able to perform the desired checks on each PR&mdash;at least not without some undesirable trade-offs.

If anyone knows of an existing workflow we can adapt, or alternative Action, please tell us!

lapp0:
Fixes #883

# Changes

This change-set configures `asv` within the repo, along with the `asv_benchmark_pr.yml` workflow to comment benchmark comparisons in each open PR. 

`tests/benchmarks/` has been moved to `benchmarks/` and converted from `pytest-benchmark` to `asv` format.

### Behavior

- Comparison is between PR `HEAD` and `outlines-dev/outlines@main`  `HEAD`. 
- The use of `--interleave-rounds -a repeat=3` in `asv continuous` mitigates variance due to environmental factors described in #883, but triples the runtime compared to a single pass.
- ["The median time from all samples collected in all roudns is used as the final measurement result."](https://asv.readthedocs.io/en/stable/benchmarks.html)
- Total benchmark workflow runtime (`repeat=3`): 23 minutes (should be close to test run time - 10 minutes)
- Runs once per push within open PRs.
- Creates a single comment per PR, and edits the comment when workflow is re-run within the same PR.

# Examples

- Times differ between 1% and 4% due to random variation: https://github.com/lapp0/outlines/pull/16#issuecomment-2118777376

- Demo of Benchmark Output for PR with Performance Regression: https://github.com/lapp0/outlines/pull/18#issuecomment-2118860838

# Out of Scope

With this infrastructure we can create useful historical performance dashboards such as https://asv-runner.github.io/asv-collection/pandas/ This requires a stable, dedicated machine. If we have an appropriate machine, the code changes are relatively easy.

# Repo Configuration Work

For this workflow we need to set up a token, `GH_TOKEN`, with appropriate permissions:

- contents: read 
  - for retrieving compared revisions
- pull-requests: read and write
  - for commenting

Then create a new `asv-benchmarks` environment, and a secret with key `GH_TOKEN`.

# TODO:

- [x] `asv` configuration
- [x] PR comment workflow
- [x] migrate benchmarks from `pytest-benchmark` to `asv`
- [ ] harden workflow security (e.g. a PR with a new workflow using `GH_TOKEN` could spam the repo using the pull-requests write permissions)
- [ ] update docs
- [ ] Optimize workflow run time (setup is majority of time, not benchmark execution)
- [ ] receive commentary

@rlouf  / @brandonwillard could you please share your thoughts on features / changes you'd like to see before this is ready for review?


benchmarks/asv.conf.json
```diff
@@ -0,0 +1,20 @@
...
+    "repo": "..",
+    "branches": [
+	"HEAD"
+    ],
+    "build_command": [
+        "pip install .[test]",
+	"python -m build --wheel -o {build_cache_dir} {build_dir}"
+    ],
+    "environment_type": "virtualenv",
+    "show_commit_url": "https://github.com/lapp0/outlines/commit/",
```
lapp0, line 14: `https://github.com/outlines-dev/outlines/commit/`

brandonwillard:
Can we do without the token/permissions if we get rid of the PR comment feature?  We really don't need a comment added to a PR if there's a workflow run with the output that we can check.  

Also, we need to confirm that this approach will block merging of the benchmarks don't pass.  Ideally we won't need to run the benchmarks on every commit, but would still need them to be run in order to merge.  In that scenario, it would be up to the maintainers to add a tag that runs the benchmarks.

.github/workflows/asv_benchmark_pr.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: Save comparison of PR against main branch
+      run: |
+        # prepare main branch for comparison
+        git remote add upstream https://github.com/${{ github.repository }}.git
+        git fetch upstream main
+
+        # Run benchmarks, writing comment contents to ./output
+        asv continuous upstream/main HEAD \
+            --factor 1.1 --sort ratio --split --interleave-rounds -a repeat=3
+        asv compare upstream/main HEAD --factor 1.1 --sort ratio --split | tee output
```
kc611, line 52: The `output` contents should be used to raise an error which would fail this Github Action and prevent a merge in case the benchmark goes beyond the `--factor` threshold.

.github/workflows/asv_benchmark_pr.yml
```diff
@@ -0,0 +1,72 @@
...
+        python -m lf_asv_formatter --asv_version "$(echo asv --version)"
+        printf "Benchmark Suite Results:\n\n" >> comment_body
+        cat output >> comment_body
+
+    # from https://github.com/hombit/load_ztfdr_for_tape/blob/9acf7c83/.github/workflows/asv-pr.yml
+    - name: Find benchmarks comment
```
kc611, line 58: As @brandonwillard mentioned if the comments are the only things that require the read/write permissions, they can be dropped for now (they can be a part of a separate PR though where we they can be further discussed)

.github/workflows/asv_benchmark_pr.yml
```diff
@@ -0,0 +1,72 @@
...
+      run: asv machine --machine gh-runner --yes
+
+    - name: Save comparison of PR against main branch
+      run: |
+        # prepare main branch for comparison
+        git remote add upstream https://github.com/${{ github.repository }}.git
+        git fetch upstream main
+
+        # Run benchmarks, writing comment contents to ./output
+        asv continuous upstream/main HEAD \
```
kc611, line 50: This maybe out of scope for this PR but just out of curiosity: It is possible to configure `asv` such that the machines on every run are somewhat similar. This makes it such that the benchmark comparision history is accurately built on the historical web browser window thing that `asv` has. (https://asv.readthedocs.io/en/stable/using.html#machine-information). 

Maybe that could also help out by allowing us to run the benchmarks only once per run and compare them with the existing `main` results which were stored as an artifact.

kc611 (commented):
Hi @lapp0 :slightly_smiling_face: 

This approach looks really good for `outlines`'s benchmarking requirements. Here are few of my thoughts on this PR.

.github/workflows/asv_benchmark_pr.yml
```diff
@@ -0,0 +1,72 @@
...
+  contents: read  # Read access for repository contents
+  pull-requests: write  # Write access for pull requests
+
+env:
+  PYTHON_VERSION: "3.10"
+  WORKING_DIR: ${{ github.workspace }}/benchmarks
+
+jobs:
+  benchmark-pr:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
lapp0: Thanks!

.github/workflows/asv_benchmark_pr.yml
lapp0: I agree. We should create a separate issue which implements the Out of Scope section of the PRs description

benchmarks/asv.conf.json
```diff
@@ -0,0 +1,20 @@
+{
+    "version": 1,
+    "project": "Outlines",
+    "project_url": "https://outlines-dev.github.io/outlines/",
+    "repo": "..",
+    "branches": [
+	"HEAD"
+    ],
+    "build_command": [
+        "pip install .[test]",
```
brandonwillard, line 10: This should be in `install_command` instead, no?  I tried it locally with `"install_command": ["in-dir={env_dir} python -mpip install '{wheel_file}[test]'"]`, but it fails with no useful output.  

We need to figure this out before going forward.

lapp0: I ran into issues using a canonical `install_command` / `build_command`, and instead opted to follow this workaround: https://github.com/airspeed-velocity/asv/issues/903#issuecomment-1276110657

brandonwillard: OK, then we should change it to something like 
```
    "build_command": [
        "in-dir={env_dir} python -mpip install .[test]",
        "python setup.py build",
        "PIP_NO_BUILD_ISOLATION=false python -mpip wheel --no-deps --no-index -w {build_cache_dir} {build_dir}",
    ],
```
which reflects the current build command defaults.

lapp0:
Thanks for the feedback!

As requested I've
- Removed comment functionality, instead results are shown in the `BENCHMARK RESULTS` section of workflow.
- Updated the workflow to ensure it runs only when workflow is manually dispatched OR for every commit when `run_benchmarks` label is applied.
- Updated the `asv.conf.json` `build_command` to be consistent with current build command defaults.
- Forced workflow failure if performance decline for any test exceeds 10%


[Rendered Docs](https://github.com/lapp0/outlines/blob/introduce-asv-ci-workflow-883/docs/community/contribute.md#benchmarking)


Requires https://github.com/outlines-dev/outlines/pull/929 merge first for tests to pass.

Example: Terrible performance regression resulting in failure:
- PR: https://github.com/lapp0/outlines/pull/18
- Workflow: https://github.com/lapp0/outlines/actions/runs/9324556694/job/25670013480?pr=18
- `BENCHMARK RESULTS` section:

benchmarks/asv.conf.json
brandonwillard: This change needs to be made.

benchmarks/bench_numba_compile.py
```diff
@@ -0,0 +1,37 @@
+import importlib
+
+import interegular
+import numba
+
+import outlines
+
+from .common import clear_outlines_cache, setup_tokenizer
+
+outlines.disable_cache()
```
brandonwillard, line 10: We need to contain global state changes like this in the set-up/tear-down of these benchmarks, or&mdash;better yet&mdash;avoid them completely.

brandonwillard (changes requested):
Just some small comments/updates and we can merge this.

lapp0:
@brandonwillard I've introduced and tested a new `outlines.caching.cache_disabled` context manager. This decorator is applied to all benchmark tests.

brandonwillard (approved):

PR #915, opened 2024-05-23
Use TQDM to track index compilation progress

### LINKED ISSUE #810: Add a progress bar for compilation
Compilation can take a few seconds, and it would be a huge improvement to the UI if we let the users know what is happening and the progress of the compilation progress. We should be able to get by with `tqdm` added to [this function](https://github.com/outlines-dev/outlines/blob/5f98bd31e9a4583eec93cf72f12bfd371a43ae12/outlines/fsm/regex.py#L695). Even if we don't know exactly how many state the compiler needs to traverse the total number of states in the FSM should give an upper bound.

lapp0:
Fixes #810 

## Changes

Use `tqdm` progress bar within `regex.py`'s `create_fsm_index_end_to_end()` so the program doesn't silently hang with no indication of status.

Update pre-commit hook to ensure `types-tqdm` is installed for `mypy` type checking.

## Example

```
$ python3 810.py 
Compiling FSM index for all state transitions: 62%|██████████████████           | 146/235 [00:12<00:00, 18.96it/s]
```

Contents of `810.py`:

```python
import interegular

from outlines.fsm.regex import (
    make_deterministic_fsm,
    make_byte_level_better_fsm,
    create_fsm_index_tokenizer
)


regex_str = "(?:(?:[0-9](?:(?:_)?[0-9])*(?:e|E)(?:(?:\\+|\\-))?[0-9](?:(?:_)?[0-9])*|(?:[0-9](?:(?:_)?[0-9])*\\.(?:[0-9](?:(?:_)?[0-9])*)?|\\.[0-9](?:(?:_)?[0-9])*)(?:(?:e|E)(?:(?:\\+|\\-))?[0-9](?:(?:_)?[0-9])*)?)|[0-9](?:(?:_)?[0-9])*)(?:J|j)|(?:[0-9](?:(?:_)?[0-9])*(?:e|E)(?:(?:\\+|\\-))?[0-9](?:(?:_)?[0-9])*|(?:[0-9](?:(?:_)?[0-9])*\\.(?:[0-9](?:(?:_)?[0-9])*)?|\\.[0-9](?:(?:_)?[0-9])*)(?:(?:e|E)(?:(?:\\+|\\-))?[0-9](?:(?:_)?[0-9])*)?)|0(?:x|X)(?:(?:_)?(?:[0-9]|[a-f]|[A-F]))+|0(?:b|B)(?:(?:_)?[0-1])+|0(?:o|O)(?:(?:_)?[0-7])+|(?:(?i:([ubf]?r?|r[ubf])('([^\\\\']|.)*?'))|(?i:([ubf]?r?|r[ubf])(\"([^\\\"]|.)*?\")))|(?:(?:\r?\n[\t ]*|#[^\n]*))+|[1-9](?:(?:_)?[0-9])*|\\\\[\t \x0c]*\r?\n|continue|nonlocal|assert|global|import|lambda|return|async|await|break|class|False|match|raise|while|yield|case|from|None|pass|True|with|def|del|for|not|try|if|[^\\W\\d]\\w*|#[^\n]*|[\t \x0c]+|\\.\\.\\.|@|\\{|\\(|\\[|\\-|\\+|\\*|\\~"

regex_pattern = interegular.parse_pattern(regex_str)
# Not reduced, so that there are many states
regex_fsm, _ = make_deterministic_fsm(regex_pattern.to_fsm())
bytes_fsm = make_byte_level_better_fsm(regex_fsm, keep_utf8=True)

num_fsm_states = len(regex_fsm.states)
assert num_fsm_states == 220

num_bytes_fsm_states = len(bytes_fsm.states)
assert num_bytes_fsm_states == 235

tokenizer = AutoTokenizer.from_pretrained("gpt2")
tokenizer = TransformerTokenizer(tokenizer)

states_to_token_subsets, empty_token_ids = create_fsm_index_tokenizer(
    bytes_fsm, tokenizer
)
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,3 +30,4 @@ repos:
     - id: mypy
       args: [--allow-redefinition]
       exclude: ^examples/
+      additional_dependencies: [types-tqdm]
--- a/outlines/fsm/regex.py
+++ b/outlines/fsm/regex.py
@@ -26,6 +26,7 @@ from interegular.fsm import (
     anything_else,
 )
 from numba.typed.typedobjectutils import _nonoptional
+from tqdm import tqdm
 
 if TYPE_CHECKING:
     from outlines.models.tokenizer import Tokenizer
@@ -692,6 +693,12 @@ def create_fsm_index_end_to_end(
     seen: Set[int] = set()
     next_states = {fsm_info.initial}
 
+    pbar = tqdm(
+        total=len(set(fsm_info.transitions.values()))
+        + 1,  # all transitions plus initial
+        desc="Compiling FSM index for all state transitions",
+    )
+
     while next_states:
         start_state = next_states.pop()
 
@@ -713,7 +720,11 @@ def create_fsm_index_end_to_end(
             if end_state not in seen:
                 next_states.add(end_state)
 
-        seen.add(start_state)
+        if start_state not in seen:
+            pbar.update(1)
+            seen.add(start_state)
+
+    pbar.close()
 
     return states_to_token_subsets
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
    "referencing",
    "jsonschema",
    "requests",
+   "tqdm"
 ]
 dynamic = ["version"]
 


rlouf:
Looks great, thank you!

Issue #908, opened 2024-05-20
Are we able to structure JSON output into a single line with just one whitespace?


timothylimyl:
### Presentation of the new feature

Output JSON without wasting tokens on whitespaces and linebreaks.

Example output: `{"name": : "Tim" , "age" : 25 , "interest" : "llm" }`

### Where does it fit in Outlines?

Structured Generation

### Are you willing to open a PR?

Yes.

lapp0:
Please pass `whitespace_token=r'[ ]?' to `outlines.generate.json()`

rlouf:
I've been wondering if we should restrict the default pattern a little more to accept a maximum of 4 white spaces and one line break? That seems like a reasonable default that should cover most of what the model has seen during training.

lapp0:
> I've been wondering if we should restrict the default pattern a little more to accept a maximum of 4 white spaces and one line break? That seems like a reasonable default that should cover most of what the model has seen during training.

My bike-shedding: typically with newlines there will be indentation involving 8, 12, or 16 spaces. We should set the default whitespace pattern to be `r'[ ]?'` and to make the json output a single line.

rlouf:
Fair. We can give it a try and see if we still get complaints from users.

timothylimyl:
> I've been wondering if we should restrict the default pattern a little more to accept a maximum of 4 white spaces and one line break? That seems like a reasonable default that should cover most of what the model has seen during training.

Where do you see that training data for JSON is under the scheme of 4 whitespaces and one line break?

The reason I am asking about single liner (flat) JSON output is to save tokens.  My intuition is that a well-formatted JSON is meant for humans and not LLM, LLM can deal with flat JSON structure in both input and output.  A single line JSON which is very long and potentially nested can be very difficult for human to read but the equivalent is most probably not true for LLM.

timothylimyl:
@rlouf @lapp0 Thanks for making the change.

Issue #690, opened 2024-02-20
Length constraint causes infinite looping of generation


piercefreeman:
### Describe the issue as clearly as possible:

I saw some generation irregularities similar to https://github.com/outlines-dev/outlines/issues/450, so benchmarked it a bit further. It seems like there are certain classes of constraints that cause an infinite loop of the state machine, where the model keeps generating "valid" tokens but they don't contribute to the overall string content.

This script uses a pydantic schema that compiles to a valid regex and state machine, which has length constraints in place: `\{[\n ]*"poem"[\n ]*:[\n ]*"(?:[^"\\\x00-\x1f\x7f-\x9f]|\\.){150,200}"[\n ]*\}`. Each line of the sequence_generator is logged separately in the format:

```
{loop} {value.token_ids.shape[1]} {generated text}
```

Within the loop, runs 0 and 1 complete successfully. On run 2 it stalls with the same maximum length output (`Once upon a cosmic dance,\nIn the vast expanse of numeric romance,\nWhere digits intertwined in grand design,\nAnd infinity was but a linear line,\n\nNumbers whispered sweet secrets to the stars,\nYielded h`).

Even as this generated text stays the same, the value.token_ids keep growing. The tokens evaluate to 28705 (`▁`) and 13 (`<0x0A>`), over and over again:

```bash
86 153 tensor(28705, device='cuda:0') "poem": "Once upon a cosmic dance,\nIn the vast expanse of numeric romance,\nWhere digits intertwined in grand design,\nAnd infinity was but a linear line,\n\nNumbers whispered sweet secrets to the stars,\nYielded h"
87 154 tensor(13, device='cuda:0') "poem": "Once upon a cosmic dance,\nIn the vast expanse of numeric romance,\nWhere digits intertwined in grand design,\nAnd infinity was but a linear line,\n\nNumbers whispered sweet secrets to the stars,\nYielded h"
88 155 tensor(28705, device='cuda:0') "poem": "Once upon a cosmic dance,\nIn the vast expanse of numeric romance,\nWhere digits intertwined in grand design,\nAnd infinity was but a linear line,\n\nNumbers whispered sweet secrets to the stars,\nYielded h"
89 156 tensor(13, device='cuda:0') "poem": "Once upon a cosmic dance,\nIn the vast expanse of numeric romance,\nWhere digits intertwined in grand design,\nAnd infinity was but a linear line,\n\nNumbers whispered sweet secrets to the stars,\nYielded h"
```

[...]

Issue #450, opened 2023-12-19
Some inferences take forever to complete


gaspard-dv:
## Issue description

The issue was raised by other people on Discord too.

To quote one of them:
> I'm running the same query 10 times (with equivalent prompts and output sizes), but some inferences are taking abnormally longer than others. 

<details>
<summary>their screenshot</summary>

![image](https://github.com/outlines-dev/outlines/assets/72385491/47c59310-da9d-4c05-8d7a-9de082ab4393)
</details>

## Repro

I made a reproduction code snippet that can run in Google Colab (w/ free T4 GPU):

<details>
<summary>💻 Code snippet</summary>

```bash
pip install outlines==0.0.13 transformers datasets optimum auto-gptq accelerate
```

```python
from outlines import models
from outlines.text.generate import json, continuation
from json import dumps
from time import perf_counter
import torch


prompt = """<|system|>
You are a friendly AI assistant.
You're specialized in mathematics and open source Github repositories.
Your answers must be concise and factual.</s>
<|user|>
Write a very long poem</s>
<|assistant|>
"""
output_format = {
    "type": "object",
    "properties": {
        "poem": {"type": "string"}
    }
}
model = models.transformers("TheBloke/zephyr-7B-beta-GPTQ", device="cuda")
```

```python
rng = torch.Generator(device="cuda")
rng.manual_seed(789001)

errors = []
for i in range(20):
  start_time = perf_counter()
  try:
    sequence = json(model, dumps(output_format))(prompt, rng=rng)
    poem = sequence.get('poem')
    elapsed_time = round(perf_counter() - start_time)
    n_characters_per_second = len(poem) // elapsed_time
    print(f"{i}\t{elapsed_time}\t{n_characters_per_second}\t{poem[:30]}..")
  except Exception as e:
    errors.append(e)
    print(f"{i}\t{elapsed_time}\tINFERENCE FAILED")
```
</details>

<details>
<summary>📃 Output</summary>

```
0	14	76	In the vastness of cosmic spac..
1	14	INFERENCE FAILED
2	769	0	In this universe, a vast expan..
3	389	0	In ancient lands, where skies ..
4	16	67	In the depths of the cosmos, w..
5	35	70	In the stillness of the mornin..
6	32	60	In a universe vast and unceasi..
7	13	77	75000 lines of blank verse, hi..
8	22	69	In a land of purest light, Who..
9	34	59	A cosmic dance of stars, a sym..
[...]

PR #916, opened 2024-05-23
Use less problematic whitespace token

### LINKED ISSUE #839: Provide more restrictive defaults for white space patterns in JSON
Repeated new lines is a common issue when generating JSON so many users end up setting the value of the `whitespace_pattern` argument. We should probably set a more restrictive limit on the number of consecutive white spaces and new lines.

lapp0:
Fixes #839 #908 #690 #450

## Problem

A major problem, especially with smaller language models, is the repetition problem. 

For example, let's say a model is generating json and must provide 12 space tokens for indentation in json output. Often a language model will assign a high probability to a 13th space token, and do the same for a 14th space, and then enter an infinite space generation loop.

This is a problem with NLG that has been known for half a decade, but only has mitigations (mirostat, repetition penalty, using hundreds of billions of weights, etc), no absolute solutions (except for **structured generation**)

## Solution

For structured json generation, we set a sane default whitespace pattern of `r"[ ]?"`. This removes all newlines and indentation. It disallows any syntactic whitespace beyond a single space separator.

Users can still set the argument `whitespace_pattern=` if they want different behavior

### DIFF
--- a/docs/reference/json.md
+++ b/docs/reference/json.md
@@ -36,10 +36,10 @@ print(result)
 
 !!! Note "JSON and whitespaces"
 
-    By default Outlines lets model choose the number of linebreaks and white spaces used to structure the JSON. Small models tend to struggle with this, in which case we recommend to set the value of the parameter `whitespace_pattern` to the empty string:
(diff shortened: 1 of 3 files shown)


timothylimyl:
Thanks for making the PR!

Issue #913, opened 2024-05-23
`json_schema.py`: If schema has an array with no `items` key, the pattern is invalid


lapp0:
### Describe the issue as clearly as possible:

In `json_schema.py`, if the below line is reached, the pattern is illegal.

https://github.com/outlines-dev/outlines/blob/main/outlines/fsm/json_schema.py#L299

In short, there are 5 right-paren, and 4 left-paren

### Steps/code to reproduce the bug:

```python
>>> from outlines.fsm.json_schema import build_regex_from_schema
>>> import interegular
>>> pattern = build_regex_from_schema("""{"type": "array"}""")
>>> interegular.parse_pattern(pattern)
Traceback (most recent call last):
  File "/home/andrew/p/outlines/.myenv/lib/python3.11/site-packages/interegular/patterns.py", line 486, in parse
    return super(_ParsePattern, self).parse()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andrew/p/outlines/.myenv/lib/python3.11/site-packages/interegular/utils/simple_parser.py", line 63, in parse
    raise NoMatch(self.data, max(self._expected), self._expected[max(self._expected)])
interegular.utils.simple_parser.NoMatch: Can not match at index 269. Got ')?[\\n', expected any of ['?', '(', '[', '\\', '.', '$', '^', "<Any 1 except ('^', '+', ')', '.', '*', '?', '(', '|', '$', '[', '\\\\')>", '|'].
Context(data[-10:+10]): '\\)*")){0,})?[\\n ]*\\]'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/andrew/p/outlines/.myenv/lib/python3.11/site-packages/interegular/patterns.py", line 730, in parse_pattern
    out = p.parse()
          ^^^^^^^^^
  File "/home/andrew/p/outlines/.myenv/lib/python3.11/site-packages/interegular/utils/simple_parser.py", line 38, in w
    return m(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/andrew/p/outlines/.myenv/lib/python3.11/site-packages/interegular/patterns.py", line 488, in parse
    raise InvalidSyntax
interegular.patterns.InvalidSyntax
```

### Expected result:

```shell
Successfully compiled pattern
```

### Outlines/Python version information:

Version information
<details>
```
[...]

PR #914, opened 2024-05-23
Allow json schema of `{}`, resulting in unconstrained json value

lapp0:
Fix #749 

`{}` wasn't handled at all in `json_schema.py`, resulting in a `ValueError`.

Now it follows the json schema spec: allow **anything**: bool, int, object, etc, etc.

PR #919, opened 2024-05-25
Fix invalid regex in unconstrained arrays for json_schema.py

lapp0:
Fixes https://github.com/outlines-dev/outlines/issues/913

Enables complete version of PR https://github.com/outlines-dev/outlines/pull/914 which allows arrays

### Problems
- Unconstrained arrays (arrays without any `items` types specified) would always result in an illegal regex
- Unconstrained arrays couldn't contain objects or arrays

### Change
- Fix Unconstrained array pattern so it doesn't produce an illegal pattern
- Allow objects and arrays within arrays, but use "mock" `depth` key to prevent infinite recursion

### DIFF
--- a/outlines/fsm/json_schema.py
+++ b/outlines/fsm/json_schema.py
@@ -297,15 +297,22 @@ def to_regex(
                 # Here we need to make the choice to exclude generating list of objects
                 # if the specification of the object is not given, even though a JSON
                 # object that contains an object here would be valid under the specification.
-                types = [
+                legal_types = [
                     {"type": "boolean"},
                     {"type": "null"},
                     {"type": "number"},
                     {"type": "integer"},
                     {"type": "string"},
                 ]
-                regexes = [to_regex(resolver, t, whitespace_pattern) for t in types]
-                return rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}){allow_empty}{whitespace_pattern}\]"
+                depth = instance.get("depth", 2)
+                if depth > 0:
+                    legal_types.append({"type": "object", "depth": depth - 1})
+                    legal_types.append({"type": "array", "depth": depth - 1})
+
+                regexes = [
+                    to_regex(resolver, t, whitespace_pattern) for t in legal_types
+                ]
+                return rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}{allow_empty}{whitespace_pattern}\]"
 
         elif instance_type == "object":
             # pattern for json object with values defined by instance["additionalProperties"]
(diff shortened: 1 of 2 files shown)


outlines/fsm/json_schema.py
```diff
@@ -297,15 +297,22 @@ def to_regex(
...
-                return rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}){allow_empty}{whitespace_pattern}\]"
+                depth = instance.get("depth", 2)
+                if depth > 0:
+                    legal_types.append({"type": "object", "depth": depth - 1})
+                    legal_types.append({"type": "array", "depth": depth - 1})
+
+                regexes = [
+                    to_regex(resolver, t, whitespace_pattern) for t in legal_types
+                ]
+                return rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}{allow_empty}{whitespace_pattern}\]"
```
ekagra-ranjan, line 315: @lapp0 - I think this is missing a parentheses to denote that we want the entire array to be optional and not just the elements which start with `,`, i.e. 
```
rf"\[({whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}){allow_empty}{whitespace_pattern}\]"
```
Note the starting bracket after `rf"\[` and closing bracket before `{allow_empty}`. Wdyt?

lapp0: I've refactored a lot of this code so we can do constrained yaml. Does this make sense?

```
    def format_object_with_additional_properties(self, value_pattern, min_properties=None, max_properties=None):
        if max_properties == 0:
            return rf"\{{{self.ws}\}}"

        inner = self._regex_repeat_elem(
            elem_pattern=f"{STRING}{self.ws}:{self.ws}{value_pattern}",
            separator_pattern=f"{self.ws},{self.ws}",
            min_elem=min_properties,
            max_elem=max_properties,
            pad=self.ws
        )
        return rf'{{{inner}}}'
```

```
    def _regex_repeat_elem(self, elem_pattern, separator_pattern, min_elem=None, max_elem=None, pad=""):
        """Creates a pattern allowing between min_elem and max_elem occurrences of elem_pattern"""
        if max_elem == 0:
            return ""

        base_pattern = f"{elem_pattern}"
        suffix_pattern = f"(?:{separator_pattern}{elem_pattern})"

        min_suffix_repeats = "" if min_elem is None else max(0, int(min_elem) - 1)
        max_suffix_repeats = "" if max_elem is None else max_elem - 1

        if max_suffix_repeats == 0:
            pattern = base_pattern
        else:
            pattern = f"{base_pattern}({suffix_pattern}){{{min_suffix_repeats},{max_suffix_repeats}}}"

        padded_pattern = f"({pad}{pattern}{pad})"

        if not min_elem:
            return f"({padded_pattern}|{pad})"
        else:
            return padded_pattern
```

tests/fsm/test_json_schema.py
```diff
@@ -719,6 +719,46 @@ def test_format(schema, regex, examples):
...
+            {
+                "title": "Foo",
+                "type": "object",
+            },
+            [
+                ("{}", True),
+                ('{"a": 1, "b": null}', True),
+                ('{"a": {"z": {"g": 4}}, "b": null}', True),
+                ("1234", False),  # not an object
+                ('["a", "a"]', False),  # not an array
```
ekagra-ranjan, line 733: @lapp0 are we missing a test case for having arrays inside object?

outlines/fsm/json_schema.py
ekagra-ranjan: Sorry, I didnt get you. I just meant that right now the regex is this:
```
rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}{allow_empty}{whitespace_pattern}\]"
```

and I think it should be this:
```
rf"\[({whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}){allow_empty}{whitespace_pattern}\]"
```

Does this make sense?

tests/fsm/test_json_schema.py
lapp0: Thanks for the comments! It's important that we create patterns that follow json-schema spec expectations. I'll ensure the two conditions you've referenced are tested for in the refactor here https://github.com/lapp0/outlines/pull/36

Issue #549, opened 2024-01-17
Run-time structured generation benchmarks


lapp0:
Initialization benchmarks are introduced in https://github.com/outlines-dev/outlines/pull/542

We should extend these benchmarks to measure the performance of inference. 

# Goal

Outlines shouldn't be a bottleneck for most inference. A reasonable goal can be set based on
- [1,200 output tok/s on H100 w/ Llama 2 70B on Optimum-Nvidia](https://huggingface.co/blog/optimum-nvidia)

Benchmarks will help us achieve and maintain that goal.

# What must be benchmarked

- End-to-end CFG ~~and Regex~~ guided generation
- Guided generation with Ray IPC logits processing with vLLM (related https://github.com/outlines-dev/outlines/issues/524)

# Proposed method

It's annoying to need a GPU to run tests. We shouldn't do actual inference in performance benchmarks.
- 1) Create a mock inference engine
- 2) Simple benchmark to ensure unguided mock inference engine takes infinitesimally small time
- 3) Guided benchmarks that show true throughput of outlines

rlouf:
At this point I think that it would only make sense to benchmark the CFG-guided generation. Regex-guided generation is only a dictionary call at each step, so there really isn't anything we could do that would move the needle.

PR #587, opened 2024-01-25
Add More Grammars to `outlines.grammars`, Benchmark and Verify Their Integrity

lapp0:
Overview:
- Adds grammars to `outlines.grammars` and fixes some grammars
- Adds benchmarks for `CFGFSM`  (`pytest --benchmark-cfg`)
- Fixes `UnexpectedToken` error resulting from a partially complete terminal being lexed as a full terminal (see `fsm.py`)

There are a lot of files changed in this PR, but they're mostly new lark grammars, lark grammar improvements, and many sample test files.
- Python changes: 
  - Add test cases
  - update `outlines/grammars.py` with new grammars
  - Prevent UnexpectedToken if terminal is partially generated and misrecognized as being a different terminal in `fsm.py`
- Lark changes:
  - Fix grammars
  - Add grammars  
- Documentation:
  - Add documentation on creating grammars ([Rendered](https://github.com/lapp0/outlines/blob/benchmark-cfg-generation/docs/reference/creating_grammars.md))
  - Add documentation on using `outlines.grammars`
- Add test generation files to `outlines/tests/benchmark/cfg_samples/`

Fixes:
- Fixes https://github.com/outlines-dev/outlines/issues/549 (PR introduces full FSM benchmarks)
- Fixes https://github.com/outlines-dev/outlines/issues/456 (PR fixes `ESCAPED_STRING` in `common.lark` and documents lookaround behavior)
- Fixes https://github.com/outlines-dev/outlines/issues/580 (PR updates `arithmetic.lark` to not result in deep recursion)
- Fixes https://github.com/outlines-dev/outlines/issues/633
- Fixes https://github.com/outlines-dev/outlines/issues/636
- Fixes https://github.com/outlines-dev/outlines/issues/634
- Fixes https://github.com/outlines-dev/outlines/issues/663

Prerequisites
- [x] https://github.com/outlines-dev/outlines/pull/570 (Necessary for correctness of grammars)
- [x] https://github.com/outlines-dev/outlines/pull/562 (Necessary infrastructure for using `outlines.grammars`)

Create / Fix Grammars
- [x] `outlines.grammars.csv`
- [x] Fix `ESCAPED_STRING` in `outlines.grammars.json`
- [x] `outlines.grammars.lark`
- [x] `outlines.grammars.sql_select` (maybe reach out to https://motherduck.com/blog/duckdb-text2sql-llm/) 
- [x] remove `outlines.grammars.lisp`
[...]

PR #925, opened 2024-05-28
Added benchmarks for larger regex fsm and runtime benchmarks for the same

kc611:
Step towards #549,

This PR adds a lager regex benchmark and a few runtime benchmarks.

### DIFF
--- a/tests/benchmark/test_benchmark_regex_fsm.py
+++ b/tests/benchmark/test_benchmark_regex_fsm.py
@@ -1,10 +1,13 @@
 import pytest
+from transformers import AutoTokenizer
 
 import outlines
+from outlines.fsm.fsm import RegexFSM
+from outlines.fsm.guide import RegexGuide  # noqa: E402
+from outlines.models.transformers import TransformerTokenizer
 
 outlines.disable_cache()
 
-from outlines.fsm.guide import RegexGuide  # noqa: E402
 
 regex_samples = {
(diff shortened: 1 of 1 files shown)


tests/benchmark/test_benchmark_regex_fsm.py
```diff
@@ -30,3 +34,32 @@ def test_benchmark_regex_to_fsm(
...
+    benchmark.pedantic(dict_run_through, rounds=8, setup=setup_large_fsm)
+
+
+def test_benchmark_regex_fsm_states(benchmark, tokenizer, ensure_numba_compiled):
+    def dict_run_through(fsm):
+        for start_state in fsm.states_to_token_maps.keys():
+            for token in fsm.allowed_token_ids(start_state):
+                # Access every end state individually, to get an amortized
+                # cost of dictionary accesses.
+                next_state = fsm.next_state(start_state, token)
```
brandonwillard, line 62: As mentioned previously, let's split these into to separate benchmarks.

lapp0:
Thanks for this work, it's really important that we prevent performance degradation! 

Related / conflicting PR which moves tests from `pytest-benchmark` to ASV: https://github.com/outlines-dev/outlines/pull/903

If you're looking to add additional benchmarks after this PR, you might consider adapting some benchmarks from https://github.com/outlines-dev/outlines/pull/587 there's a lot of benchmarks in there for CFG generation.

lapp0:
Hi @kc611 the ASV infrastructure is in `main`. Could you please move these to `benchmarks/`?

 Here is an example :https://github.com/outlines-dev/outlines/blob/main/benchmarks/bench_regex_guide.py

Issue #918, opened 2024-05-24
`mlx` library integration (via `mlx-lm`)


pngwn:
An additional library integration for mlx.

**Context**

[`mlx`](https://github.com/ml-explore/mlx) is an ML framework that support high-performance inference on apple silicon (amongst other things). It has now has a rich ecosystem and vibrant community of users.

**Request**

The [`mlx-lm`](https://github.com/ml-explore/mlx-examples/blob/main/llms/README.md#python-api) python library provides some simple utilities for text generation and it would be great if there were an integration with outlines.

**Extra detail**

[`mlx-lm` supports a `logit_bias` param in its top level generate function](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/utils.py#L232) 

Related to #806

pngwn:
There is a separate library that integrates outlines with mlx. I thought I would post it here in case it is useful:

https://github.com/sacha-ichbiah/outlines-mlx

I have also pinged the author.

namin:
In the meantime, note that it's possible to use MLX via the OpenAI API. Of course, then, one is limited to generators `choice` and `text`, but I guess this is better than nothing.

This is what I did to get it working:
```
# see https://github.com/ml-explore/mlx-examples/pull/810 which must be merged
# tested as follows
# mlx_lm.server --model mlx-community/gemma-1.1-7b-it-4bit --port 11435 --chat-template=CHAT_TEMPLATE
# where CHAT_TEMPLATE is as in the tokenizer_config below

from openai import AsyncOpenAI
from outlines.models.openai import OpenAI, OpenAIConfig

from mlx_lm.tokenizer_utils import load_tokenizer
from pathlib import Path
model_path = Path("mlx-community/gemma-1.1-7b-it-4bit")
[...]

Issue #855, opened 2024-05-02
Logits Processors `Guide` integration will be buggy when `len(tokens) > 1` in a `Write` instruction


br3no:
### Describe the issue as clearly as possible:

See:
https://github.com/outlines-dev/outlines/blob/d6a2b7908065d420456118723f69908c4094c1f8/outlines/integrations/vllm.py#L110

Here the `tokens` field of the next instruction is treated equally regardless of whether it is of type `Generate` or `Write`. 

If a `Write` instruction has a `tokens` field with length > 1, this means we will accept any of the next ff-tokens as the token in the next step. This is incorrect.

### Steps/code to reproduce the bug:

```python
The bug will only appear once the ff-tokens have length > 1.
```

### Expected result:

```shell
N.a.
```

### Outlines/Python version information:

Version information
<details>

```
python -c "import sys; print('Python', sys.version)"
pip freeze
0.0.39
Python 3.8.18 (default, Oct  2 2023, 15:02:11) 
[GCC 9.4.0]
accelerate==0.27.2
ai2-olmo==0.2.5
aiohttp==3.9.3
aioprometheus==23.12.0
aiosignal==1.3.1
annotated-types==0.6.0
antlr4-python3-runtime==4.9.3
anyio==4.2.0
async-timeout==4.0.3
attrs==23.2.0
awscli==1.32.83
boto3==1.34.83
botocore==1.34.83
cached_path==1.6.2
cachetools==5.3.3
certifi==2024.2.2
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==3.0.0
cmake==3.29.2
codespell==2.2.6
colorama==0.4.4
cupy-cuda12x==12.1.0
diskcache==5.6.3
distro==1.9.0
docutils==0.16
einops==0.7.0
exceptiongroup==1.2.0
fastapi==0.109.2
fastrlock==0.8.2
filelock==3.13.1
frozenlist==1.4.1
fsspec==2024.2.0
google-api-core==2.18.0
google-auth==2.29.0
google-cloud-core==2.4.1
google-cloud-storage==2.16.0
google-crc32c==1.5.0
google-resumable-media==2.7.0
googleapis-common-protos==1.63.0
h11==0.14.0
hiredis==2.3.2
httpcore==1.0.4
httptools==0.6.1
httpx==0.27.0
huggingface-hub==0.20.3
idna==3.6
importlib-resources==6.1.1
importlib_metadata==7.0.2
[...]

Issue #789, opened 2024-04-05
RegexPrefixAllowedTokens does not work for batch


randomcodelookup:
### Describe the issue as clearly as possible:

In these lines of code, https://github.com/outlines-dev/outlines/blob/3a41b0edf389ceba49126e811ed0c1b23b50f235/outlines/integrations/transformers.py#L87-L100

it seems that the batch id is not used at all, and there is an error triggered when batch is used and input_ids is empty.
https://github.com/outlines-dev/outlines/blob/3a41b0edf389ceba49126e811ed0c1b23b50f235/outlines/integrations/transformers.py#L116

Do you know how this can be fixed? Thanks! @saattrupdan @rlouf

### Steps/code to reproduce the bug:

```python
# Based on https://github.com/outlines-dev/outlines/blob/3a41b0edf389ceba49126e811ed0c1b23b50f235/examples/transformers_integration.py#L6
from pydantic import BaseModel
from transformers import  AutoModelForCausalLM, AutoProcessor

from outlines.integrations.transformers import JSONPrefixAllowedTokens


class Person(BaseModel):
    first_name: str
    surname: str


processor = AutoProcessor.from_pretrained("mistralai/Mistral-7B-v0.1")
model = AutoModelForCausalLM.from_pretrained("mistralai/Mistral-7B-v0.1")
prefix_allowed_tokens_fn = JSONPrefixAllowedTokens(
    schema=Person, tokenizer_or_pipe=processor.tokenizer,
)

inputs = processor(...)
model.generate(
    **inputs,
    prefix_allowed_tokens_fn=prefix_allowed_tokens_fn,
)
```

### Expected result:

```shell
When inputs is a batch, this should not trigger an empty list indexing error
```

### Outlines/Python version information:

Version information
<details>
```
(command output here)
```
</details>

rlouf:
You are correct, this needs to be fixed.

randomcodelookup:
Thanks @rlouf . I'm looking to fix this too but just to get more context, it seems like this function generally does not allow parallelized operations across the batch is that true? I also wonder why not implement an equivalent of JSONLogitsProcessor, is this method faster?

rlouf:
Thank you for considering contributing. I am not sure why this choice was made at the time, you'd have to ask @saattrupdan

lapp0:
[...]

PR #926, opened 2024-05-29
Use `LogitsProcessor` for `transformers` integration

### LINKED ISSUE #806: Update the `transformers` integration
In the vein of #782 and #772 we should refactor the `transformers` integration to use logits processor.

We will keep the custom sampling loop, but via a `Outlines` model that wraps `transformers` models. We should be able to remove `torch` and `transformers` as default dependencies as a result.

lapp0:
Fixes #806

Related: https://github.com/outlines-dev/outlines/issues/789

Requesting comment on whether we should unify the logits processor in this PR as described in the last section of this message. @brandonwillard @rlouf 

### Problem

`transformers` has two issues in its `generate` integration:
- if `input_ids` is empty, it fails
- it has an inconsistent implementation with other models, it is the only model which doesn't use logits processors

### Solution

Implement logits processors for transformers.

#### Proposed Additional Work

We have four implementations of logits processors which do the same thing. Their implementations have drifted and now there are growing differences in their implementation details. We should simplify this and use a single logits processor for all models, ensuring a bug fix or enhancement for one logits processor applies to all models.

- Additionally, this minimizes the changes necessary for https://github.com/outlines-dev/outlines/issues/918  minimal
- Furthermore, this issue only would need to be fixed in one module with a singular implementation: https://github.com/outlines-dev/outlines/issues/855

This WIP PR shows generalized `logits_processors.py` which aims to be compatible with all inference engines, along with a modified `generate/regex.py` incorporating the new singular logits processor implementation.


lapp0:
closed in favor of https://github.com/outlines-dev/outlines/pull/966

Issue #795, opened 2024-04-10
Accelerate the index construction process


aeft:
### What behavior of the library made you think about the improvement?

My understanding of the index construction process is that for each state in the FSM, we need to iterate through all tokens in the vocabulary (e.g., 32000) and find all valid tokens (i.e., those that can reach valid states).

### How would you like it to behave?

First, let me give a example:

Considering the current state and its out edges, if its out edges are limited, for example, only Number edge, which means the state transistion happens only if the first character of the next token is a number. We can exploit this finding to avoid the huge vocabulary iteration for this state, i.e., we only iterate through tokens whose first character is a number. 

We can apply this idea into other characters. To do this, we need to preprocess the vocabulary, i.e., find the corresponding tokens for each first character. There are two ways:
1. We sort the vocabulary based on the first character. Then, for each character, we maintain an index pair (start_index, end_index), which indicates the token range.
2. We don't change the original vocabulary but create a new sorted vocabulary.

The first way can save a little memory, and it depends on whether the order of original vocabulary matters.

The cost introduced by this method (let's call it method1) is the sorting cost. If the number of states is large or we can save the sorted vocabulary. The cost can be almost ignored (amortized).

---

Can we extend this idea further?

We can build a trie tree for the vocabulary. Then we use BFS to traverse the trie tree and maintain the corresponding state from the FSM. Thus, for tokens with the same prefix, we don't need to access the FSM multiple times. For example, for "hel", "hell", and "hello" as tokens, the "h->e->l" transition path only needs to be traversed once.

The real performance gain from method2 depends on the implementation because the original implementation uses numba.jit, and it may compensate for the performance of the naive algorithm.

---

Note: For method1 and method2, we still need to consider all states. Are there some methods to skip some states? I'm still figuring that out. (Maybe impossible, because in the worst case, the LLM generates one character at a time.)

I might have gotten something wrong. I am looking forward to your advice if you are available. Thank you! If you don't mind, I would like to implement this when the method is mature.

rlouf:
Weakly related to #758.

aeft:
It seems like the trie implementation in #507 didn't get merged. Are there some reasons?

rlouf:
Said PR was making several changes at the same time, with different impacts on the compile time, which made it difficult to evaluate the performance impact. We asked for the PR to be split, but the author never got around to it. Happy to review a PR that implements a single change and with proper and extensive benchmarks.

[...]

PR #929, opened 2024-05-29
Fix llamacpp caching by making `LlamaCppTokenizer` an outlines `Tokenizer`

### LINKED ISSUE #922: `llamacpp` tests failing in `main`

Failures:

```
tests/generate/test_integration_llamacpp.py .FF.x......FFFFFFFFFFFFFF
```

All failures have a similar traceback. Example:

```
_________________ test_llamacpp_generation_api[regex-params1] __________________

model = <outlines.models.llamacpp.LlamaCpp object at 0x7ff76c1591b0>
generator_type = <function regex at 0x7ff81d254940>, params = ('[0-9]',)

    @pytest.mark.parametrize(
        "generator_type,params",
        (
            (generate.text, []),
            (generate.regex, ("[0-9]",)),
            (generate.cfg, (grammars.arithmetic,)),
        ),
    )
    def test_llamacpp_generation_api(model, generator_type, params):
>       generator = generator_type(model, *params)

tests/generate/test_integration_llamacpp.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/functools.py:889: in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/outlines/generate/regex.py:48: in regex_llamacpp
    logits_processor = RegexLogitsProcessor(regex_str, llm=model.model)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/outlines/integrations/llamacpp.py:154: in __init__
    fsm = RegexGuide(regex_string, tokenizer)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/outlines/fsm/guide.py:145: in __init__
    ) = create_states_mapping(regex_string, tokenizer)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/outlines/caching.py:118: in wrapper
    result = wrapper.__memory__.get(cache_key, default=ENOVAL, retry=True)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/diskcache/core.py:1149: in get
    db_key, raw = self._disk.put(key)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/outlines/caching.py:19: in put
    data = cloudpickle.dumps(key)
/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/cloudpickle/cloudpickle.py:1479: in dumps
    cp.dump(obj)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cloudpickle.cloudpickle.Pickler object at 0x7ff76add7b80>
obj = ('outlines.fsm.guide.create_states_mapping', '[0-9]', <outlines.integrations.llamacpp.LlamaCppTokenizer object at 0x7ff76ab05360>, None)

    def dump(self, obj):
        try:
>           return super().dump(obj)
E           ValueError: ctypes objects containing pointers cannot be pickled

/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/site-packages/cloudpickle/cloudpickle.py:1245: ValueError
```

Full run example:
https://github.com/lapp0/outlines/actions/runs/9252615735/job/25450574607

Failures started between ` May 25 06:24 GMT` -  `May 26 08:04 GMT`

```python
pytest -s tests/generate/test_integration_llamacpp.py
```

### Expected result:

```shell
Tests pass
```

lapp0:
Fixes #922

## Problem

`integrations/llamacpp.py` / `LlamaCppTokenizer` has `self.decode = model.tokenizer.decode`. This is not pickle-able. Attempting to pass a `LlamaCppTokenizer` to a `@cache`d function results in  `cloudpickle.dumps(tokenizer)` which causes `ValueError: ctypes objects containing pointers cannot be pickled` error.

Locally most `test_integration_llamacpp.py` fail.  (full test failure details in #922)

##  Solution

Make `LlamaCppTokenizer` subclass `outlines.models.tokenizer.Tokenizer` and implement its abstract methods. Specifically `__getstate__` is necessary to ensure `cloudpickle.dumps` works and hashing is possible.

### DIFF
--- a/outlines/integrations/llamacpp.py
+++ b/outlines/integrations/llamacpp.py
@@ -26,7 +26,7 @@ limitations under the License.
 """
 
 import math
-from typing import TYPE_CHECKING, Dict, Optional, Set, Type, Union
+from typing import TYPE_CHECKING, Optional, Type, Union
 
 import numpy as np
 import torch
@@ -36,45 +36,10 @@ from pydantic import BaseModel
 from outlines.fsm.guide import CFGGuide, Guide, RegexGuide
 from outlines.fsm.json_schema import build_regex_from_schema
 from outlines.integrations.utils import convert_json_schema_to_str
+from outlines.models.llamacpp import LlamaCppTokenizer
 
 if TYPE_CHECKING:
     from llama_cpp import Llama
-
-
-class LlamaCppTokenizer:
-    def __init__(self, model: "Llama"):
-        self.eos_token_id = model.token_eos()
-        self.eos_token = model.tokenizer().decode([self.eos_token_id])
-        self.pad_token_id = self.eos_token_id
-        self.special_tokens: Set[int] = set()
-
-        self.vocabulary: Dict[str, int] = dict()
-
-        tokenizer = model.tokenizer()
-
-        self.decode = tokenizer.decode
-
-        # TODO: Remove when https://github.com/ggerganov/llama.cpp/pull/5613 is resolved
-        try:
-            self.vocabulary = model.tokenizer_.hf_tokenizer.get_vocab()
-        except AttributeError:
-            # ###
-            for t in range(model.n_vocab()):
-                token_piece = model.tokenizer().decode([t])
-                self.vocabulary[token_piece] = t
-
-    def convert_token_to_string(self, token: str) -> str:
-        return token
-
-    def __getstate__(self):
-        """Allow tokenizer to be used as hash key by excluding self.decode"""
-        return (
-            self.vocabulary.items(),
-            self.eos_token_id,
-            self.eos_token,
-            self.pad_token_id,
-            sorted(self.special_tokens),
-        )
 
 
 class LogitsProcessor:
--- a/outlines/integrations/llamacpp.py
+++ b/outlines/integrations/llamacpp.py
@@ -66,6 +66,16 @@ class LlamaCppTokenizer:
     def convert_token_to_string(self, token: str) -> str:
         return token
 
+    def __getstate__(self):
+        """Allow tokenizer to be used as hash key by excluding self.decode"""
+        return (
+            self.vocabulary.items(),
+            self.eos_token_id,
+            self.eos_token,
+            self.pad_token_id,
+            sorted(self.special_tokens),
+        )
+
 
 class LogitsProcessor:
     """Bias LlamaCpp generation using a finite state machine.
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -1,13 +1,100 @@
 import dataclasses
+import pickle
 import warnings
-from typing import TYPE_CHECKING, Iterator, List, Optional, TypedDict, Union
+from typing import (
+    TYPE_CHECKING,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    TypedDict,
+    Union,
+)
 
 from typing_extensions import Unpack
 
 from outlines.generate.api import GenerationParameters, SamplingParameters
+from outlines.models.tokenizer import Tokenizer
 
 if TYPE_CHECKING:
     from llama_cpp import Llama, LogitsProcessorList
+
+
+class LlamaCppTokenizer(Tokenizer):
+    def __init__(self, model: "Llama"):
+        self.eos_token_id = model.token_eos()
+        self.eos_token = model.tokenizer().decode([self.eos_token_id])
+        self.pad_token_id = self.eos_token_id
+        self.special_tokens: Set[int] = set()
+
+        self.vocabulary: Dict[str, int] = dict()
+
+        self.tokenizer = model.tokenizer()
+
+        # TODO: Remove when https://github.com/ggerganov/llama.cpp/pull/5613 is resolved
+        try:
+            self.vocabulary = model.tokenizer_.hf_tokenizer.get_vocab()
+        except AttributeError:
+            # ###
+            for t in range(model.n_vocab()):
+                token_piece = model.tokenizer().decode([t])
+                self.vocabulary[token_piece] = t
+
+        # ensure stable ordering of vocabulary
+        self.vocabulary = {
+            tok: tok_id
+            for tok, tok_id in sorted(self.vocabulary.items(), key=lambda x: x[1])
+        }
+
+        self._hash = None
+
+    def decode(self, token_ids: List[int]) -> List[str]:
+        decoded_bytes = self.tokenizer.detokenize(token_ids)
+        return [decoded_bytes.decode("utf-8", errors="ignore")]
+
+    def encode(
+        self, prompt: Union[str, List[str]], add_bos: bool = True, special: bool = True
+    ) -> Tuple[List[int], List[int]]:
+        if isinstance(prompt, list):
+            raise NotImplementedError(
+                "llama-cpp-python tokenizer doesn't support batch tokenization"
+            )
+        token_ids = self.tokenizer.tokenize(
+            prompt.encode("utf-8", errors="ignore"), add_bos=add_bos, special=special
+        )
+        # generate attention mask, missing from llama-cpp-python
+        attention_mask = [
+            1 if token_id != self.pad_token_id else 0 for token_id in token_ids
+        ]
+        return token_ids, attention_mask
+
+    def convert_token_to_string(self, token: str) -> str:
+        return token
+
+    def __eq__(self, other):
+        if not isinstance(other, LlamaCppTokenizer):
+            return False
+        return self.__getstate__() == other.__getstate__()
+
+    def __hash__(self):
+        if self._hash is None:
+            self._hash = hash(pickle.dumps(self))
+        return self._hash
+
+    def __getstate__(self):
+        """Create a stable representation for outlines.caching"""
+        return (
+            self.vocabulary,
+            self.eos_token_id,
+            self.eos_token,
+            self.pad_token_id,
+            sorted(self.special_tokens),
+        )
+
+    def __setstate__(self, state):
+        raise NotImplementedError("Cannot load a pickled llamacpp tokenizer")
 
 
 class LlamaCppParams(TypedDict, total=False):
--- a/tests/generate/conftest.py
+++ b/tests/generate/conftest.py
@@ -1,0 +1,24 @@
+from importlib import reload
+
+import pytest
+
+
+@pytest.fixture
+def temp_cache_dir():
+    import os
+    import tempfile
+
+    import outlines.caching
+    import outlines.fsm.guide
+
+    with tempfile.TemporaryDirectory() as tempdir:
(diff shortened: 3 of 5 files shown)


outlines/integrations/llamacpp.py
```diff
@@ -66,6 +66,16 @@ def __init__(self, model: "Llama"):
     def convert_token_to_string(self, token: str) -> str:
         return token
 
+    def __getstate__(self):
+        """Allow tokenizer to be used as hash key by excluding self.decode"""
+        return (
+            sorted(self.vocabulary.items()),
```
brandonwillard, line 72: Is `sorted` necessary here?

brandonwillard (commented):
Also, the most important thing we need is a test that helps avoid issues like this in the future.

outlines/integrations/llamacpp.py
lapp0: I don't think so. If we use a `llama_cpp` tokenizer we decode sequentially by token ID. HF tokenizers `get_vocab()` has consistent ordering. I experimented with both, and they were fine without `sorted()`.

Fixed.

lapp0:
@brandonwillard should be ready for re-review when you have a chance by the way. A minimal reproducer test added.

brandonwillard (commented):
I just noticed this, but shouldn't `LlamaCppTokenizer` extend our `Tokenizer` class in the first place?  If that's the case, we can use that interface and simply copy the relevant parts (e.g. serialization methods) of our `transformers` `Tokenizer` subclass and be done.

lapp0:
> I just noticed this, but shouldn't `LlamaCppTokenizer` extend our `Tokenizer` class in the first place? If that's the case, we can use that interface and simply copy the relevant parts (e.g. serialization methods) of our `transformers` `Tokenizer` subclass and be done.

I agree that we should work towards that in general (relevant discussion https://github.com/outlines-dev/outlines/discussions/695)


vLLM also uses a separate tokenizer which doesn't guarantee the same interface as `outlines.models.tokenizer.Tokenizer`. I can open an issue for ensuring all tokenizers are subclasses of outlines `Tokenizer`.

brandonwillard:
> I agree that we should work towards that in general (relevant discussion #695)

We can start that work here.  It should only involve a change to `LlamaCppTokenizer(Tokenizer)` and adding proper `__eq__`, `__hash__`, `__getstate__`, and `__setstate__` implementations adapted from `TransformerTokenizer`.  Finally, the test `test_RegexGuide_caching` can be copied and used for `LlamaCppTokenizer`.

lapp0:
I've implemented `LlamaCppTokenizer` as a subclass of outlines `Tokenizer`.

It cannot be loaded from disk because `llama_cpp` tokenizers aren't pickleable. `__setstate__` and `__getstate__` are implemented solely for use with `outlines.caching`.

`decode()` and `encode()` have been adapted to support the type expectations of `Tokenizer`, allowing for uniformity between `models.llamacpp` and `models.transformers` use within sequence generators. However I left batch encoding (`tokenizer.encode(List[str])`)  as `NotImplementedError`, as `llama_cpp` doesn't track the tokenizers `padding_side`, therefore I cannot set the padding in batch encoded sequences safely.

outlines/models/llamacpp.py
```diff
@@ -1,15 +1,95 @@
...
+        attention_mask = [
+            1 if token_id != self.pad_token_id else 0 for token_id in token_ids
+        ]
+        return token_ids, attention_mask
+
+    def convert_token_to_string(self, token: str) -> str:
+        return token
+
+    def __eq__(self, other):
+        return hash(self) == hash(other)
```
brandonwillard, line 71: This won't be consistent due to hash conflicts.  The members/fields of this class&mdash;aside from `self.tokenizer`&mdash;should be sufficient to define this method, just as they are for `__getstate__` and `__hash__`.  (If they're not, then we need to add any missing members/fields to this class.)

N.B. See [`__eq__` and `__hash__`](https://docs.python.org/3/reference/datamodel.html#object.__eq__) for more info.

tests/models/test_llamacpp.py
```diff
@@ -0,0 +1,7 @@
+from outlines.fsm.guide import create_states_mapping
+from outlines.models.llamacpp import LlamaCppTokenizer
+
+
+def test_create_states_mapping_llamacpp(model):
+    """Minimal reproducer for #922, error passing llamacpp tokenizer to create_states_mapping"""
+    create_states_mapping("a", LlamaCppTokenizer(model.model))
```
brandonwillard, line 7: We shouldn't need this anymore.  Plus, our tests need to at least `assert` something (excluding rare cases in which that's not possible/practical, of course).

brandonwillard (changes requested):

tests/models/test_llamacpp.py
lapp0: To clarify, this test fails in `main` because `create_states_mapping` cannot be passed a `LlamaCppTokenizer`. Will remove though, as the test you recommended is excellent for verifying the `LlamaCppTokenizer` changes.

lapp0:
Thanks for your feedback @brandonwillard !

I've incorporated your requested changes.

outlines/models/llamacpp.py
```diff
@@ -1,15 +1,97 @@
...
+    def convert_token_to_string(self, token: str) -> str:
+        return token
+
+    def __eq__(self, other):
+        if not isinstance(other, LlamaCppTokenizer):
+            return False
+        return self.__getstate__() == other.__getstate__()
+
+    def __hash__(self):
+        # cache object hash
```
brandonwillard, line 76:
```suggestion
```

outlines/models/llamacpp.py
```diff
@@ -1,15 +1,97 @@
...
+        if self._hash is None:
+            self._hash = hash(pickle.dumps(self))
```
brandonwillard, line 78: I don't think `pickle.dumps` handles `dict` key and/or `set` value ordering, so this might not be a good approach.  

I think I see what you were trying to do previously with the sorting, but it doesn't matter for serialization.  It might only seem to matter because you're mixing the serialization interface with the equivalence check and hashing.  That's not necessary, though; you can compare the relevant member objects directly in `__eq__` and take special steps that are good for hashing in `__hash__` and only there (e.g. `json.dumps(..., sort_keys=True)` for `dict`s seems to be favored by many).

lapp0: >I think I see what you were trying to do previously with the sorting, but it doesn't matter for serialization. It might only seem to matter because you're mixing the serialization interface with the equivalence check and hashing. That's not necessary

It matters because the only reason `Tokenizer`s are serializable via `__setstate__` is because their serialized form is used to create a stable hash for the  `@cache`d `create_states_mapping(regex_string, tokenizer)`. Tokenizers are never deserialized.

> (e.g. json.dumps(..., sort_keys=True) for dicts seems to be favored by many).

`json.dumps` is much slower than pickle and pickling a `Tokenizer` is already 0.25 seconds, which matters because every time we create an FSM index we check the cache which has the pickled tokenizer as a key.

Dicts have stable order since 3.6, and while I successfully experimented with this, I don't know of a **guarantee** pickle maintains order. How about we revert to sorting to be safe? 

But this discussion brings up an important point. Once other index construction bottlenecks are taken care of by https://github.com/outlines-dev/outlines/issues/795 maybe we should address the performance issues I just described. We should only calculate the hash of the serialized tokenizer once.

brandonwillard: > It matters because the only reason `Tokenizer`s are serializable via `__setstate__` is because their serialized form is used to create a stable hash for the `@cache`d `create_states_mapping(regex_string, tokenizer)`. Tokenizers are never deserialized. They are serialized for hashing.

You're still mixing up serialization with equivalence and hashing.  The hashing we're talking about here (i.e. `__hash__`) is completely independent of any caching with `cache`.  

Also, if you want to address the potential cache misses due to `dict` and `set` ordering, that can be done in `CloudpickleDisk`.  That's where serialization is used for cache indexing purposes.

> `json.dumps` is much slower than pickle and pickling a `Tokenizer` is already 0.25 seconds, which matters because every time we create an FSM index we check the cache which has the pickled tokenizer as a key.

We can use whatever is sufficiently performant and accurate.

brandonwillard: > How about we revert to sorting then pickling to be safe?

If that's the best option, then we might as well make sure that `self.vocabulary` is sorted upon construction/creation.  Sometimes these `dict`s are already sorted by token ID, in which case that canonicalization step would be rather efficient.

outlines/models/llamacpp.py
```diff
@@ -1,15 +1,102 @@
...
+        return self._hash
+
+    def __getstate__(self):
+        """Create a stable representation for outlines.caching"""
+        return (
+            self.vocabulary,
+            self.eos_token_id,
+            self.eos_token,
+            self.pad_token_id,
+            sorted(self.special_tokens),
```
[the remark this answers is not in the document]
brandonwillard: Oh, it might be `adapt_tokenzier`.  That's problematic, since it means that this class isn't actually immutable, which&mdash;at the very least&mdash;invalidates the hashibility requirement.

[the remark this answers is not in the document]
lapp0: > I was going to ask why `special_tokens` isn't sorted in the constructor, but now I don't even see where/how it's being populated.

`cloudpickle` has non-deterministic behavior for sets, (even `FrozenSet`s) we need to convert to a sorted list when serializing to ensure a stable hash.

>Oh, it might be adapt_tokenzier. That's problematic, since it means that these class instances aren't actually immutable, which—at the very least—invalidates the hashability requirement.

`adapt_tokenizer` doesn't apply to `llamacpp` for now. It's only called in `integrations/transformers.py` and `integrations/vllm.py`.

>It looks like we need to change adapt_tokenizer so that it returns a completely new Tokenizer instance, or perhaps integrate the changes made by adapt_tokenizer into the Tokenizer subclasses directly.

IMHO we should create a separate issue to unify / correct the behavior and interfaces of tokenizers in general to prevent the scope of this PR from growing too large. This PR doesn't introduce any new problems with `LlamaCppTokenizer`, but it does fix the tests which have been failing in `main` all week.

brandonwillard (approved):
We can merge this now in order to prevent any more errors, but we need some follow-up issues for the mutability inconsistencies related to these types.

lapp0:
I agree 100% https://github.com/outlines-dev/outlines/issues/936

PR #904, opened 2024-05-19
Fix null byte `\x00` issue by switching to `numba.types.unicode_type`

M0gician:
Fixes #833 

## Changes
- Adopt `numba.types.unicode_type` in `outlines.fsm.regex` implementation to avoid [a bug in `numba.unicodecharseq` which its boxer has an invalid skip on null-byte causing the copy to end prematurely](https://github.com/numba/numba/issues/9542).

## Test
I've used this patch for about a month. So far everything is working well without any problem.

Issue #833, opened 2024-04-23
KeyError in `BetterFSM::FSMInfo` when input FSM `alphabet` contains UTF-8 characters that ends with `\xb8\x80`


m0g1cian:
#### Update 2

Can confirm there's something wrong with Numba's Typed Dict implementation. Check issue [here](https://github.com/numba/numba/issues/9542)

#### Update

> After some testing, it is clear that this KeyError occurs for UTF-8 characters that ends with `\xb8\x80` (e.g. "帀", "㸀", "渀", 縀"). 

-------

When `outlines` builds `BetterFSM` from a reference FSM (e.g. from `interegular`), if the reference FSM contains Chinese character "一", the corresponding `numba.typed.Dict` used by `BetterFSM::alphabet_symbol_map` somehow converts this character into an empty string, causing a KeyError whenever `__getitem__` is triggered .

### Steps/code to reproduce the bug:
`debug_keyerror.py`

```python
import interegular
from outlines.fsm.regex import FSMInfo, make_deterministic_fsm

if __name__ == "__main__":
    regex_string = r"(1|2|3|one|two|three|一|二|三)"
    regex_pattern = interegular.parse_pattern(regex_string)
    regex_fsm, _ = make_deterministic_fsm(regex_pattern.to_fsm().reduce())
    fsm_info: FSMInfo = regex_fsm.fsm_info
    print(fsm_info)
```

#### Some insight:
print (k, v) in `alphabet_symbol_mapping_items` before `create_fsm_info()` (right after `outlines.fsm.regex.py::96`)
```
e 3
w 9
2 1
h 4
t 8
o 6
r 7
1 0
3 2
n 5
一 10
二 12
三 11
```

print (k, v) in `alphabet_symbol_mapping_items` in `create_fsm_info()`  when building `alphabet_symbol_map` (right after `outlines.fsm.regex.py::139`)
```
e 3
w 9
2 1
h 4
t 8
o 6
r 7
1 0
3 2
n 5
 10
二 12
三 11
```

### Expected result:

I was able to get the expected result after tweaking two places:
1. `outlines.fsm.regex.py::112`: change `nb_unichar_2_type = numba.types.UnicodeCharSeq(2)` to `nb_unichar_2_type = numba.types.unicode_type`
[...]

PR #930, opened 2024-05-30
Fix null byte `\x00` issue in byte level fsm resulting in `KeyError` in `BetterFSM::FSMInfo` 

lapp0:
Follow-up to https://github.com/outlines-dev/outlines/pull/904 by @M0gician  -- Closes #904

Fixes https://github.com/outlines-dev/outlines/issues/833

# Problem 1

@m0gician identified [a bug in numba's handling of `UnicodeCharSeq`](https://github.com/numba/numba/issues/9542). if the elements first item is `\x00`, the entire element is null. This numba bug resulted in a `KeyError` while compiling a patterns index for certain characters.

More problem details available in #904

### Solution 1

@m0gician implemented the base solution - represent token bytes with `numba.unicode_type` to avoid this bug.

https://github.com/outlines-dev/outlines/commit/4a5ef55d683f72f18f50e5f7e49ede742cee91fa

# Problem 2

Use of `unicode_type` results in ambiguous representation of hex-bytes / chars.

E.g. in main a `UnicodeCharSeq` will keep the hex bytes and characters separate`["😇", "9", "F", "9F"]`   

Because we use unicode_type instead a List[UnicodeCharSeq(2)], the example sequence would be represented as

`["😇", "9", "F", "9F"]` -> `'😇9F9F'`

This demonstrates a problem, we have no idea whether consecutive hex characters represent a byte or two separate utf-8 characters.

### Solution 2

To resolve this, we prefix a hex-byte with a null byte.

`["😇", "9", "F", "9F"]` -> `'😇9F\x009F'`.

https://github.com/outlines-dev/outlines/commit/83c4d3a37f6d340e3ad21e5d93209f51c4bfe557

# Problem 3

Parsing a token to separate bytes and chars during `_walk_fsm` is inefficient and resulted in index compilation taking ~2.2x as long as `main`

### Solution 3

Instead of iterating over a string in `_walk_fsm`, we precompute a mapping of token -> transition key seq via `get_tokens_transition_keys()` and iterate over list of integer transition keys for each token in `_walk_fsm()`

https://github.com/outlines-dev/outlines/commit/56ea9572cd343778e2c566a66dddde2ec75489ab

Now instead of taking ~2.2x as long as `main`, index compilation takes ~0.65x as long as `main`. However Numba takes 16% longer to compile.

Benchmark Suite Results:

Benchmarks that have improved:

|    | Before [0b4d12b0]    | After [fe5cb3df]    |   Ratio | Benchmark (Parameter)                                                                                     |
|----------|----------------------|---------------------|---------|-----------------------------------------------------------------------------------------------------------|
| -        | 8.03±0.1s            | 5.29±0.02s          |    0.66 | JsonSchemaBenchmark.time_json_schema_to_fsm('complex_schema')                           |
| -        | 5.86±0.06s           | 3.72±0.07s          |    0.64 | JsonSchemaBenchmark.time_json_schema_to_fsm('simple_schema')                            |
| -        | 4.41±0.07s           | 2.77±0.01s          |    0.63 | RegexGuideBenchmark.time_regex_to_guide('complex_phone')                                |
| -        | 9.57±0.01s           | 6.37±0.02s          |    0.67 | RegexGuideBenchmark.time_regex_to_guide('complex_span_constrained_relation_extraction') |
| -        | 4.20±0.02s           | 2.63±0.01s          |    0.63 | RegexGuideBenchmark.time_regex_to_guide('date')                                         |
| -        | 4.12±0.04s           | 2.56±0.01s          |    0.62 | RegexGuideBenchmark.time_regex_to_guide('email')                                        |
| -        | 4.07±0.04s           | 2.54±0.01s          |    0.62 | RegexGuideBenchmark.time_regex_to_guide('ip')                                           |
| -        | 3.97±0.08s           | 2.48±0.03s          |    0.62 | RegexGuideBenchmark.time_regex_to_guide('simple_phone')                                 |
| -        | 3.89±0.02s           | 2.41±0.02s          |    0.62 | RegexGuideBenchmark.time_regex_to_guide('ssn')                                          |
| -        | 3.96±0.03s           | 2.49±0.01s          |    0.63 | RegexGuideBenchmark.time_regex_to_guide('time')                                         |
| -        | 4.22±0.04s           | 2.67±0.01s          |    0.63 | RegexGuideBenchmark.time_regex_to_guide('url')                                          |

Benchmarks that have stayed the same:

|    | Before [0b4d12b0]    | After [fe5cb3df]    |   Ratio | Benchmark (Parameter)                                                                                              |
|----------|----------------------|---------------------|---------|--------------------------------------------------------------------------------------------------------------------|
|          | 88.4±1μs             | 86.5±1μs            |    0.98 | bench_json_schema.JsonSchemaBenchmark.time_json_schema_to_regex('complex_schema')                                  |
|          | 49.2±0.4μs           | 48.6±0.3μs          |    0.99 | JsonSchemaBenchmark.time_json_schema_to_regex('simple_schema')                                   |
|          | 598M                 | 600M                |    1    | MemoryRegexGuideBenchmark.peakmem_regex_to_guide('complex_span_constrained_relation_extraction') |
|          | 496M                 | 499M                |    1.01 | MemoryRegexGuideBenchmark.peakmem_regex_to_guide('simple_phone')                                 |

Benchmarks that have got worse:

|    | Before [0b4d12b0]    | After [fe5cb3df]    |   Ratio | Benchmark (Parameter)                                        |
|----------|----------------------|---------------------|---------|--------------------------------------------------------------|
| +        | 4.98±0.05s           | 5.77±0.03s          |    1.16 | NumbaCompileBenchmark.time_compile_numba |

Performance degradation detected!

### DIFF
--- a/tests/fsm/test_parsing.py
+++ b/tests/fsm/test_parsing.py
@@ -9,7 +9,14 @@ from lark.lexer import UnexpectedCharacters, UnexpectedToken
 from outlines.fsm.parsing import PartialLark, PartialPythonIndenter
 
 
-def test_partial_parsing():
+@pytest.fixture
+def cleanup_lark_import():
+    yield
+    # Clean up lark.lark.LarkOptions._defaults
+    importlib.reload(lark.lark)
+
+
+def test_partial_parsing(cleanup_lark_import):
     lp = PartialLark.open_from_package(
         "tests",
         "partial_python.lark",
@@ -136,11 +143,8 @@ def test_partial_parsing():
     assert len(parser_state.state_stack) == 4
     assert parser_state.value_stack[-1].type == "LPAR"
 
-    # Clean up lark.lark.LarkOptions._defaults
-    importlib.reload(lark.lark)
 
-
-def test_sequential_parse_example():
+def test_sequential_parse_example(cleanup_lark_import):
     input_tokens = [
         "x ",
         "= ",
--- a/tests/fsm/test_parsing.py
+++ b/tests/fsm/test_parsing.py
@@ -204,6 +204,3 @@ def test_sequential_parse_example(cleanup_lark_import):
 
         if i + 1 == len(input_tokens):
             assert all(tk in next_vocab for tk in ["\n", "\nde", "  ", " + 1"])
-
-    # Clean up lark.lark.LarkOptions._defaults
-    importlib.reload(lark.lark)
--- a/outlines/fsm/parsing.py
+++ b/outlines/fsm/parsing.py
@@ -38,6 +38,7 @@ from lark.parsers.lalr_parser import LALR_Parser, ParseConf, ParserState, _Parse
 from outlines.fsm.regex import (
     fsm_union,
     get_sub_fsms_from_seq,
+    get_token_transitions,
     make_deterministic_fsm,
     walk_fsm,
 )
@@ -569,9 +570,15 @@ class PartialScanner(Scanner):
 
         text_part = text[start_pos:]
 
+        text_transitions = get_token_transitions(
+            self.fsm.fsm_info.alphabet_symbol_mapping,
+            self.fsm.fsm_info.alphabet_anything_value,
+            text_part,
+        )
+
         state_seq = walk_fsm(
             self.fsm,
-            text_part,
+            text_transitions,
             start_state,
             full_match=self.match_whole,
         )
--- a/outlines/fsm/parsing.py
+++ b/outlines/fsm/parsing.py
@@ -38,7 +38,7 @@ from lark.parsers.lalr_parser import LALR_Parser, ParseConf, ParserState, _Parse
 from outlines.fsm.regex import (
     fsm_union,
     get_sub_fsms_from_seq,
-    get_token_transitions,
+    get_token_transition_keys,
     make_deterministic_fsm,
     walk_fsm,
 )
@@ -570,7 +570,7 @@ class PartialScanner(Scanner):
 
         text_part = text[start_pos:]
 
-        text_transitions = get_token_transitions(
+        text_transitions = get_token_transition_keys(
             self.fsm.fsm_info.alphabet_symbol_mapping,
             self.fsm.fsm_info.alphabet_anything_value,
             text_part,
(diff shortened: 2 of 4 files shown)


M0gician:
Should we add some test cases like the one in #833 to verify if this `\x00` issue is solved in later updates as well? I am a little concerned whether numba's bug fix will break the current implementation in the future.

lapp0:
> Should we add some test cases like the one in #833 to verify if this `\x00` issue is solved in later updates as well? I am a little concerned whether numba's bug fix will break the current implementation in the future.

You mean a test case to ensure that a `unicode_type` starting with `\x00` isn't broken in numba in the same way it breaks `UnicodeCharSeq`?

M0gician:
Yes exactly

________________________________

lapp0:
I have introduced two new tests to verify the behavior:

# `def test_numba_leading_null_byte_UnicodeCharSeq_remains_broken()`

This test asserts that https://github.com/numba/numba/issues/9542 is still a problem. If this test **fails** we can **safely use `UnicodeCharSeq` to represent byte fsms.

`UnicodeCharSeq` allows for a cleaner representation of byte fsms than `unicode_type`, so its preferable that we use it, but this test failing doesn't necessarily indicate anything is broken.

# `def test_numba_leading_null_byte_unicode_type_sane()`

This test asserts that the `unicode_type` representation of null bytes (introduced in this PR) is legal.

If this fails, then we should consider reverting to use of `UnicodeCharSeq` (assuming the previous test also fails, indicating `UnicodeCharSeq` is viable).

tests/fsm/test_parsing.py
```diff
@@ -9,7 +9,14 @@
 from outlines.fsm.parsing import PartialLark, PartialPythonIndenter
 
 
-def test_partial_parsing():
+@pytest.fixture
+def cleanup_lark_import():
```
lapp0, line 13: This change isn't directly related to the PR, but fixes an issue - if the tests in `test_parsing.py` fail, it cascades and causes a large number of unrelated tests to fail.

PR #940, opened 2024-06-03
Add Documentation on Outlines Versioning and Releases

lapp0:
## Problem

We want to ensure downstream users are aware of Outlines versioning scheme and how to safely depend on and pin Outlines.

## Solution

Create a documentation page

Rendered: https://github.com/lapp0/outlines/blob/versioning-docs/docs/versioning.md

### DIFF
--- a/docs/community/versioning.md
+++ b/docs/community/versioning.md
@@ -1,3 +1,10 @@
+---
+title: Versioning Guide
+---
+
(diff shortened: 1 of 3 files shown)


rlouf:
Great! We need to link this from somewhere, for instance the `Community` menu.

PR #763, opened 2024-03-20
fixed parsing token vocabularies for gemma and gpt-sw3 models

ai-and-i:
This PR closes #762

Issue #762, opened 2024-03-20
Regex FSM fails with some tokenizers


saattrupdan:
### Describe the issue as clearly as possible:

When running a regex FSM with some tokenizers, the `reduced_vocabulary` function fails due to the existence of a token `▁�` in the vocabulary. This includes, but is not limited to, the following model types:

- All Gemma models
- All GPT-Sw3 models

The following models are _not_ affected by this:

- Mistral models
- LLaMA models
- OLMo models
- Qwen models
- Phi models
- Yi models
- GPT-2 models

This was implemented in [this commit](https://github.com/outlines-dev/outlines/commit/043117f47520c159028bb61bf11db0559d7b6deb) as part of [this PR](https://github.com/outlines-dev/outlines/pull/738). Tagging @ksvladimir here as he might be able to help resolve this 🙂 

**Note that this bug happens when installing from the main branch and not any release, but I include it here to be fixed before the next release (ideally)**

### Steps/code to reproduce the bug:

```python
from outlines.fsm.regex import reduced_vocabulary
from outlines.integrations.utils import adapt_tokenizer
from transformers import AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("AI-Sweden-Models/gpt-sw3-6.7b-v2")
tokenizer = adapt_tokenizer(tokenizer)
vocabulary = reduced_vocabulary(tokenizer)
```

### Expected result:

```shell
No error message.
```

### Error message:

```shell
RuntimeError                              Traceback (most recent call last)
Cell In [45], line 1
----> 1 vocab = reduced_vocabulary(tok)

File ~/.venv/lib/python3.10/site-packages/outlines/fsm/regex.py:793, in reduced_vocabulary(tokenizer)
    789         token_bytes = cast(
    790             List[int], [gpt2_unicode_to_bytes().get(c) for c in token]
    791         )
    792         if None in token_bytes:
--> 793             raise RuntimeError(
    794                 f"Cannot convert token `{token}` ({token_idx}) to bytes: {token_str}"
    795             )
    796     token_str = tuple(byte_symbol(b) for b in token_bytes)
    798 vocabulary.setdefault(token_str, []).append(token_idx)

RuntimeError: Cannot convert token `▁�` (6546) to bytes:  �
```

### Outlines/Python version information:

Version information
<details>

```
0.0.37.dev8+gf7cafe5
Python 3.10.13
annotated-types==0.6.0
asttokens==2.4.1
attrs==23.2.0
[...]

PR #948, opened 2024-06-10
Fix/extend re replacement seq

saattrupdan:
This PR is an extension of https://github.com/outlines-dev/outlines/pull/763, related to extending the `re_replacement_seq` regex.

The new [NorwAI models](https://huggingface.co/NorwAI) use a tokenizer that has the token `�.`, which leads to the same error as was described in the previous issue https://github.com/outlines-dev/outlines/issues/762.

This PR extends the fix from https://github.com/outlines-dev/outlines/pull/763 to deal with this case, as well as adding a unit test to test various tokenizers, and a comment describing why we need the prefix and suffix in the regex.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,8 @@ jobs:
     - name: Run tests
       run: |
         pytest --cov=outlines
+      env:
+        HUGGINGFACE_API_TOKEN: ${{ secrets.HUGGINGFACE_API_TOKEN }}
     - name: Upload coverage data
       uses: actions/upload-artifact@v3
       with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,8 +36,6 @@ jobs:
     - name: Run tests
       run: |
         pytest --cov=outlines
-      env:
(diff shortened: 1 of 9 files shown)


rlouf:
`HUGGINGFACE_API_TOKEN` should now be available in actions

saattrupdan:
> `HUGGINGFACE_API_TOKEN` should now be available in actions

That's great, thanks. Are you able to get access to the following models with that token?

1. `meta-llama/Meta-Llama-3-8B`
2. `mistralai/Mistral-7B-v0.3`
3. `google/gemma-2b`
4. `NorwAI/NorwAI-Mistral-7B-instruct`

Those are the culprits of the failing tests.

Also, just to be sure, the `HUGGINGFACE_API_TOKEN` is an [access token](https://huggingface.co/settings/tokens) and not a token for the Hugging Face inference API, right?

rlouf:
It's probably better to do this "manually" indeed.

saattrupdan:
@rlouf Changed the test to a "manual" one now, and all tests pass 🙂

rlouf:
Awesome, thank you!

PR #951, opened 2024-06-10
Update the documentation for OpenAI models

### LINKED ISSUE #947: Update the documentation for OpenAI-compatible models
`outlines.models.OpenAI` is now initialised by passing the OpenAI client, and the documentation is outdated.

We should add copy-pasteable examples for all major model providers that have an OpenAI compatible API.

rlouf:
Closes #947.

We also update the `outlines.models.openai` and `outlines.models.azure_openai` functions so they accept every parameter that the clients provided by the `openai` SDK accept.

- [x] Finish updating the documentation
- [x] Test the new interface for `openai` and `azure_openai`
- [ ] Add an example for alternative providers

### DIFF
--- a/docs/reference/models/openai.md
+++ b/docs/reference/models/openai.md
@@ -1,55 +1,162 @@
-# Generate text with the OpenAI and compatible APIs
+# OpenAI and compatible APIs
 
 !!! Installation
 
     You need to install the `openai` and `tiktoken` libraries to be able to use the OpenAI API in Outlines.
 
-Outlines supports models available via the OpenAI Chat API, e.g. ChatGPT and GPT-4. The following models can be used with Outlines:
+## OpenAI models
+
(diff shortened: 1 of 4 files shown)


docs/reference/models/openai.md
```diff
@@ -1,81 +1,169 @@
...
+## OpenAI models
+
+Outlines supports models available via the OpenAI Chat API, e.g. ChatGPT and GPT-4. You can initialize the model by passing the model name to `outlines.models.openai`:
 
 ```python
 from outlines import models
 
+
 model = models.openai("gpt-3.5-turbo")
 model = models.openai("gpt-4")
```
lapp0, line 16: nit, change to `gpt-4-turbo` and add `gpt-4o`

lapp0: For azure `gpt-4` is turbo, for openai it is `gpt-4` from june '23.

lapp0 (changes requested):
Minor change, we should include OpenAIs latest models. Looks great otherwise!

Issue #959, opened 2024-06-12
Using context-free grammars to guide generation does not work


border-b:
### Describe the issue as clearly as possible:

I was trying to run the example of using cfg to guide generation. But it seems there is some issue with `CFGGuide`. Just running the example without any change produces an error.

### Steps/code to reproduce the bug:

```python
import outlines
 
arithmetic_grammar = """
    ?start: expression
 
    ?expression: term (("+" | "-") term)*
 
    ?term: factor (("*" | "/") factor)*
 
    ?factor: NUMBER
           | "-" factor
           | "(" expression ")"
 
    %import common.NUMBER
"""
 
model = outlines.models.transformers("WizardLM/WizardMath-7B-V1.1")
generator = outlines.generate.cfg(model, arithmetic_grammar)
sequence = generator("Alice had 4 apples and Bob ate 2. Write an expression for Alice's apples:")
 
print(sequence)
# (8-2)
```

### Expected result:

```shell
(8-2)
```

### Error message:

```shell
Traceback (most recent call last)
Cell In[1], line 19
     17 model = outlines.models.transformers("WizardLM/WizardMath-7B-V1.1")
     18 generator = outlines.generate.cfg(model, arithmetic_grammar)
---> 19 sequence = generator("Alice had 4 apples and Bob ate 2. Write an expression for Alice's apples:")
     21 print(sequence)
     22 # (8-2)
 
File /usr/local/lib/python3.10/site-packages/outlines/generate/api.py:207, in SequenceGenerator.__call__(self, prompts, max_tokens, stop_at, rng)
    205 while True:
    206 try:
--> 207 last_state = next(states)
    208 if max_tokens or stop_sequences:
    209 token_ids = last_state.token_ids
 
File /usr/local/lib/python3.10/site-packages/outlines/generate/generator.py:92, in sequence_generator(model, sampler, fsms, token_ids, sequence_weights, attention_masks, fsm_states, rng)
     89 fsms = reorder_fsms(fsms, ancestors)
     90 fsm_states = reorder_fsm_states(fsm_states, ancestors)
---> 92 fsm_states = get_next_fsm_states(fsms, fsm_states, next_token_ids)
     93 is_finished = is_generation_finished(fsms, fsm_states)
     95 if is_finished:
 
File /usr/local/lib/python3.10/site-packages/outlines/generate/generator.py:131, in get_next_fsm_states(fsms, fsm_states, next_token_ids)
    114 def get_next_fsm_states(
[...]

PR #966, opened 2024-06-13
Update `models.transformers` to use `SequenceGeneratorAdapter` and `OutlinesLogitsProcessors`

### LINKED ISSUE #789: RegexPrefixAllowedTokens does not work for batch

### LINKED ISSUE #806: Update the `transformers` integration

lapp0:
~~In draft until https://github.com/huggingface/transformers/pull/31448 makes it into a new `transformers` release~~

Fixes https://github.com/outlines-dev/outlines/issues/1021

Fixes #789

Fixes #806 (does everything except remove the issues requirement of "remove `torch`")

Closes https://github.com/outlines-dev/outlines/pull/910 (device inconsistency issue handled through other means)

# Problem

- For #789 in `SequenceGenerator` if `input_ids` is empty, it fails.
- Generally, `models.transformers` has an inconsistent implementation with other modules making the codebase harder to manage.

# Solution

- Use `SequenceGeneratorAdapter` for transformers instead of `SequenceGenerator`
- Implement `Transformers.generate` and `Transformers.stream` which use `model.generate(logits_processor=...)` argument with `outlines.processors.OutlinesLogitsProcessor`

### Additional Changes

- ~~Temporarily disables `stop_at` argument for transformers and implements test to determine whether upstream is fixed (https://github.com/huggingface/transformers/issues/31435)~~
  - fixed upstream, change reverted
- Update docs and examples to use `seed` instead of `rng`
- Disables `outlines.generate.cfg` for now (https://github.com/outlines-dev/outlines/issues/959)

# TODO:

- [x] Determine cause of regression when generating json
  - was caused by `GenerationConfig` default `max_tokens` being 20
- [x] replace `rng` with `seed` in all tests and all documentation.
- [x] Awaiting `transformers` fix for https://github.com/huggingface/transformers/issues/31435 so `stop_at` works.

### DIFF
--- a/docs/reference/models/transformers.md
+++ b/docs/reference/models/transformers.md
@@ -15,7 +15,7 @@ Outlines provides an integration with the `torch` implementation of causal model
 ```python
 from outlines import models
 
-model = models.transformers("mistralai/Mistral-7B-v0.1", device="cuda")
+model = models.transformers("mistralai/Mistral-7B-v0.3", device="cuda")
 ```
 
 If you need more fine-grained control you can also initialize the model and tokenizer separately:
@@ -28,6 +28,57 @@ from outlines import models
 llm = AutoModelForCausalLM.from_pretrained("gpt2", output_attentions=True)
 tokenizer = AutoTokenizer.from_pretrained("gpt2")
 model = models.Transformers(llm, tokenizer)
+```
+
+# Using Logits Processors
+
+There are two ways to use Outlines Structured Generation with HuggingFace Transformers:
+- 1) Use Outlines generation wrapper, `outlines.models.transformers`
+- 2) Use `OutlinesLogitsProcessor` with `transformers.AutoModelForCausalLM`
+
+Outlines supports a myriad of logits processors for structured generation. In these example, we will use the `RegexLogitsProcessor` which guarantees generated text matches the specified pattern.
+
+## Example: `outlines.models.transformers`
+
+```
+import outlines
+
+time_regex_pattern = r"(0?[1-9]|1[0-2]):[0-5]\d\s?(am|pm)?"
+
+model = outlines.models.transformers("microsoft/Phi-3-mini-4k-instruct", device="cuda")
+generator = outlines.generate.regex(model, time_regex_pattern)
+
+output = generator("The the best time to visit a dentist is at ")
+print(output)
+# 2:30 pm
+```
+
+## Example: Direct `transformers` library use
+
+```
+import outlines
+import transformers
+
+
+model_uri = "microsoft/Phi-3-mini-4k-instruct"
+
+outlines_tokenizer = outlines.models.TransformerTokenizer(
+    transformers.AutoTokenizer.from_pretrained(model_uri)
+)
+phone_number_logits_processor = outlines.processors.RegexLogitsProcessor(
+    "\\+?[1-9][0-9]{7,14}",  # phone number pattern
+    outlines_tokenizer,
+)
+
+generator = transformers.pipeline('text-generation', model=model_uri)
+
+output = generator(
+    "Jenny gave me her number it's ",
+	logits_processor=transformers.LogitsProcessorList([phone_number_logits_processor])
+)
+print(output)
+# [{'generated_text': "Jenny gave me her number it's 2125550182"}]
+# not quite 8675309 what we expected, but it is a valid phone number
 ```
 
 [transformers]: https://github.com/huggingface/transformers
--- a/examples/llamacpp_example.py
+++ b/examples/llamacpp_example.py
@@ -1,6 +1,5 @@
 from enum import Enum
 
-import torch
 from pydantic import BaseModel, constr
 
 import outlines
@@ -37,10 +36,9 @@ if __name__ == "__main__":
     generator = outlines.generate.json(model, Character)
 
     # Draw a sample
-    rng = torch.Generator(device="cpu")
-    rng.manual_seed(789005)
+    seed = 789005
 
     prompt = "Instruct: You are a leading role play gamer. You have seen thousands of different characters and their attributes.\nPlease return a JSON object with common attributes of an RPG character. Give me a character description\nOutput:"
 
-    sequence = generator(prompt, rng=rng, max_tokens=512)
+    sequence = generator(prompt, seed=seed, max_tokens=512)
     print(sequence)
--- a/outlines/generate/cfg.py
+++ b/outlines/generate/cfg.py
@@ -1,16 +1,14 @@
 from functools import singledispatch
 
-from outlines.fsm.guide import CFGGuide
-from outlines.generate.api import SequenceGenerator, SequenceGeneratorAdapter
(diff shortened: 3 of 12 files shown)


outlines/generate/cfg.py
```diff
@@ -24,40 +22,16 @@ def cfg(model, cfg_str: str, sampler: Sampler = multinomial()) -> SequenceGenera
...
-
-
-@cfg.register(MLXLM)
-@cfg.register(VLLM)
-def cfg_unimplemented(
-    model,
-    cfg_str: str,
-    sampler: Sampler = multinomial(),
-):
     raise NotImplementedError(
```
lapp0, line 28: Not directly related to the PR, but because `cfg` is broken for all models, we should reference the issue to enable it.

outlines/generate/regex.py
```diff
@@ -39,8 +40,9 @@ def regex(model, regex_str: str, sampler: Sampler = multinomial()):
 
 
 @regex.register(MLXLM)
-def regex_mlxlm(
-    model: MLXLM,
+@regex.register(Transformers)
```
lapp0, line 43: the `_unified` dispatchers will become the default dispatcher as a next step. In this PR it's just used by `MLXLM` and `Transformers`

outlines/models/transformers.py
```diff
@@ -190,6 +194,169 @@ def __call__(
...
+        from transformers import GenerationConfig, set_seed
+
+        max_tokens, stop_at, seed = dataclasses.astuple(generation_parameters)
+        sampler, num_samples, top_p, top_k, temperature = dataclasses.astuple(
+            sampling_parameters
+        )
+
+        # global seed, not desirable
+        if seed is not None:
+            set_seed(seed)
```
lapp0, line 214: We might consider getting rid of `seed` all-together in a separate issue.

tests/generate/conftest.py
```diff
@@ -1,24 +1,40 @@
...
+        skipped_models.append("model_vllm")
+
+    skip_marker = pytest.mark.skip(
+        reason="Skipping test due to incompatible hardware / system."
+    )
+    for item in items:
+        if "model_fixture" in item.fixturenames:
+            model_param = item.callspec.params.get("model_fixture", None)
+            if model_param in skipped_models:
+                item.add_marker(skip_marker)
```
lapp0, line 37: These changes aren't directly related to the PR, but a minor change: I added vLLM integration tests to `test_generate.py`. This disables vLLM (and mlxlm) tests if the hardware is wrong.

examples/llamacpp_example.py
```diff
@@ -37,10 +36,9 @@ class Character(BaseModel):
     generator = outlines.generate.json(model, Character)
 
     # Draw a sample
-    rng = torch.Generator(device="cpu")
-    rng.manual_seed(789005)
+    seed = 789005
 
     prompt = "Instruct: You are a leading role play gamer. You have seen thousands of different characters and their attributes.\nPlease return a JSON object with common attributes of an RPG character. Give me a character description\nOutput:"
 
-    sequence = generator(prompt, rng=rng, max_tokens=512)
```
lapp0, line 45: This documentation change reflects behavior in `main`.

docs/reference/models/transformers.md
```diff
@@ -30,4 +30,55 @@ tokenizer = AutoTokenizer.from_pretrained("gpt2")
 model = models.Transformers(llm, tokenizer)
 ```
 
+# Using Logits Processors
```
rlouf, line 33: We'll need to improve the documentation to reach something similar to the lamacpp integration's.

outlines/models/transformers.py
```diff
@@ -179,6 +182,10 @@ def forward(
 
         return output.logits, output.past_key_values
 
+    @property
+    def device(self):
+        return self.model.device
```
lapp0, line 187: test should be fixed rather than adding this unnecessary property

outlines/models/transformers.py
```diff
@@ -190,6 +197,182 @@ def __call__(
...
+    def stream(
+        self,
+        prompts: Union[str, List[str]],
+        generation_parameters: GenerationParameters,
+        logits_processor: Optional["OutlinesLogitsProcessor"],
+        sampling_parameters: SamplingParameters,
+    ) -> Iterator[Union[str, List[str]]]:
+        """
+        Temporary stream stand-in which implements stream() signature
+        and equivalent behaviour but isn't iterable.
```
lapp0, line 261: Incorrect doc string

docs/reference/models/transformers.md
lapp0: We should plan a restructuring and cleaning up of documentation in a separate issue. I could share some ideas in call on how we might approach this.

In this case, a lot of information documented for `llamacpp` applies to all other models including `transformers`. We shouldn't repeat ourselves. We should explain the behavior of all models generally, highlight the models differences with a feature table, and document only `transformers` specific information on its documentation page.

rlouf: What I meant was listing the main arguments to you can pass when initialising and calling the model, cf https://outlines-dev.github.io/outlines/reference/models/llamacpp/

rlouf (approved):

rlouf:
Very good work, thank you!

PR #955, opened 2024-06-11
Remove unused dependencies, add `deptry` to make sure the dependencies stay up-to-date

fpgmaas:
Running [`deptry`](https://deptry.com/) on the codebase with this configuration

```

[tool.deptry]
extend_exclude = ["examples", "benchmarks", "outlines/integrations", "outlines/models"]
pep621_dev_dependency_groups = ["test"]
```

gives:

```
outlines/fsm/types.py:5:1: DEP003 'typing_extensions' imported but it is a transitive dependency
outlines/generate/api.py:161:16: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/api.py:279:16: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/generator.py:65:12: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/generator.py:205:12: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/generator.py:228:12: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/generator.py:269:12: DEP004 'torch' imported but declared as a dev dependency
outlines/generate/generator.py:304:12: DEP004 'torch' imported but declared as a dev dependency
outlines/samplers.py:67:16: DEP004 'torch' imported but declared as a dev dependency
outlines/samplers.py:147:16: DEP004 'torch' imported but declared as a dev dependency
outlines/samplers.py:177:12: DEP004 'torch' imported but declared as a dev dependency
outlines/samplers.py:202:12: DEP004 'torch' imported but declared as a dev dependency
outlines/samplers.py:287:16: DEP004 'torch' imported but declared as a dev dependency
outlines/types/email.py:3:1: DEP003 'typing_extensions' imported but it is a transitive dependency
outlines/types/isbn.py:3:1: DEP003 'typing_extensions' imported but it is a transitive dependency
outlines/types/phone_numbers.py:8:1: DEP003 'typing_extensions' imported but it is a transitive dependency
outlines/types/zip_codes.py:7:1: DEP003 'typing_extensions' imported but it is a transitive dependency
pyproject.toml: DEP002 'datasets' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'ray' defined as a dependency but not used in the codebase
Found 19 dependency issues.

For more information, see the documentation: https://deptry.com/
```

This PR proposes to fix these issues. In addition, it proposes to remove `pyairports` and `pycountry`, since their imports are wrapped in a try-except block:

```py
try:
[...]

PR #967, opened 2024-06-14
Fix some dependency issues and remove obsolete try-except block

fpgmaas:
Since https://github.com/outlines-dev/outlines/pull/955 was closed with the reason that there was no interest in adding another step to the CI pipeline, it might still be worthwhile to resolve some issues with the project's dependencies without adding `deptry` to the project or the CI pipeline. This PR:

- Removes the obsolete try-except blocks for importing `pyairports` and `pycountry`, since they are listed in the project's dependencies.
- Add the transitive dependency `typing_extensions` to the dependencies
- Remove `ray` as a development dependency

Let me know your thoughts :) Feel free to close if you think it does not bring improvement to the project

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ dependencies = [
    "requests",
    "tqdm",
    "datasets",
+   "typing_extensions",
    "pycountry",
    "pyairports",
 ]
@@ -64,7 +65,6 @@ test = [
 ]
 serve = [
     "vllm>=0.3.0",
-    "ray==2.9.0",
     "uvicorn",
     "fastapi",
     "pydantic>=2.0",
--- a/outlines/types/airports.py
+++ b/outlines/types/airports.py
@@ -1,13 +1,7 @@
 """Generate valid airport codes."""
 from enum import Enum
 
-try:
-    from pyairports.airports import AIRPORT_LIST
-except ImportError:
-    raise ImportError(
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -36,7 +36,8 @@ dependencies = [
    "jsonschema",
    "requests",
    "tqdm",
-   "datasets",
+   "typing_extensions",
+   "torch",
```
rlouf, line 40: We rewrote the code so we don't have to force users who use models that don't require `torch` to run outlines without installing `torch`, which is a big dependency.

lapp0:
@remic33 do many users use Outlines in a way which doesn't require torch? The only models which don't require torch AFAIK are `models.llamacpp`, `models.mlxlm` and `models.openai`.

@fpgmaas Could you rebase and squash this to one or two commits?

remic33:
@remic33

fpgmaas:
> @remic33 do many users use Outlines in a way which doesn't require torch? The only models which don't require torch AFAIK are `models.llamacpp`, `models.mlxlm` and `models.openai`.
> 
> @fpgmaas Could you rebase and squash this to one or two commits?

done!

rlouf:
`torch` is a really big dependency and I would rather not force it on people who don’t need it; especially libraries that import outlines just for the FSM logic (vLLM, etc)

rlouf:
Thanks for contributing @fpgmaas ! Merging.

Issue #976, opened 2024-06-16
ModuleNotFoundError for numpy >= 2.0.0


scott2b:
### Describe the issue as clearly as possible:

function_base appears now to be private and no longer directly accessible as done here: https://github.com/outlines-dev/outlines/blob/main/outlines/base.py#L8

As of numpy 2.0.0, this will raise:
ModuleNotFoundError: No module named 'numpy.lib.function_base'

### Steps/code to reproduce the bug:

```python
$ pip install 'numpy>=2.0.0
$ python

>>> import outlines
```
```


### Expected result:

```shell
the module imports
```

### Error message:

```shell
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/__init__.py", line 2, in <module>
    import outlines.generate
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/generate/__init__.py", line 2, in <module>
    from .cfg import cfg
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/generate/cfg.py", line 5, in <module>
    from outlines.models import OpenAI
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/models/__init__.py", line 14, in <module>
    from .openai import OpenAI, azure_openai, openai
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/models/openai.py", line 9, in <module>
    from outlines.base import vectorize
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/base.py", line 8, in <module>
    from numpy.lib.function_base import (
ModuleNotFoundError: No module named 'numpy.lib.function_base'
```

### Outlines/Python version information:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/__init__.py", line 2, in <module>
    import outlines.generate
  File "/Users/scott/github/scott2b/work/outlines/.venv/lib/python3.12/site-packages/outlines/generate/__init__.py", line 2, in <module>
    from .cfg import cfg
[...]

PR #977, opened 2024-06-16
Pin Numpy: `numpy<2.0.0`, Prevent `ModuleNotFoundError`

lapp0:
Addresses: https://github.com/outlines-dev/outlines/issues/976

Fixes https://github.com/vllm-project/vllm/pull/5582

We should close 976 **after** we fix Outlines to be compatible with Numpy >= 2.0.0, which this PR doesn't do.

### Problem

Cannot `import outlines` due to `outlines/base.py` importing `from numpy.lib.function_base import (` which no longer exists in numpy 2.0.0.

### Temporary Solution

Pin `numpy<2.0.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
    "jinja2",
    "lark",
    "nest_asyncio",
-   "numpy",
+   "numpy<2.0.0",
    "cloudpickle",
    "diskcache",
    "pydantic>=2.0",


glibg10b:
Major revision bumps (1.x.x -> 2.x.x) always have breaking changes. Should the other requirements not be limited in this way as well?

That would have to wait, though. This issue affects all scripts and modules that import outlines, and should be merged ASAP.

lapp0:
> Major revision bumps (1.x.x -> 2.x.x) always have breaking changes. Should the other requirements not be limited in this way as well?
> 
> That would have to wait, though. This issue affects all scripts and modules that import outlines, and should be merged ASAP.

Good idea, I created an issue for that https://github.com/outlines-dev/outlines/issues/981

WoosukKwon:
Hi @rlouf, could you please make a new release? vLLM imports outlines for the OpenAI server and is currently broken because of this Numpy issue.

rlouf:
Will do right now

Issue #978, opened 2024-06-16
Bug: Samplers error from the Outline docs classification example


jrinder42:
### Describe the issue as clearly as possible:

The error occurs when trying to replicate the classification example [here](https://outlines-dev.github.io/outlines/cookbook/classification/). This error seems to be due to the `isinstance` always returning False. This might be more of a Python oddity.

### Steps/code to reproduce the bug:

```python
class FizzBuzz:
    pass

hi = FizzBuzz
print(isinstance(hi, FizzBuzz))  # --> False
```

### Expected result:

```shell
True
```

### Outlines/Python version information:

Outlines: 0.0.44
Python: 3.10.14

PR #980, opened 2024-06-16
Fix samplers documentation

jrinder42:
This fixes issue #978. Specifically for the classification example in the outlines cookbook section [here](https://outlines-dev.github.io/outlines/cookbook/classification/),  it fixes the piece of code in outlines.generate.api.SequenceGeneratorAdapter that keeps returning the wrong result (i.e. returning False even if a specific sampler is passed (greedy, multinomial, or beam search).

### DIFF
--- a/docs/cookbook/classification.md
+++ b/docs/cookbook/classification.md
@@ -45,7 +45,7 @@ Outlines provides a shortcut to do multi-label classification, using the `outlin
 ```python
 from outlines.samplers import greedy
 
-generator = outlines.generate.choice(model, ["URGENT", "STANDARD"], sampler=greedy)
+generator = outlines.generate.choice(model, ["URGENT", "STANDARD"], sampler=greedy())
 ```
 Outlines supports batched requests, so we will pass two requests to the model:
 
@@ -96,7 +96,7 @@ class Classification(BaseModel):
 and we can use `generate.json` by passing this Pydantic model we just defined, and call the generator:
 
 ```python
-generator = outlines.generate.json(model, Classification, sampler=greedy)
+generator = outlines.generate.json(model, Classification, sampler=greedy())
 labels = generator(prompts)
 print(labels)
 # [Classification(label=<Label.urgent: 'URGENT'>), Classification(label=<Label.standard: 'STANDARD'>)]
--- a/outlines/samplers.py
+++ b/outlines/samplers.py
@@ -77,7 +77,7 @@ class GreedySampler:
         return next_token_ids, ancestors, weights
 
 
-greedy = GreedySampler()
+greedy = GreedySampler
 
 
 class MultinomialSampler:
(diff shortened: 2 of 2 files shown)


lapp0:
Thanks for identifying and working towards fixing this issue!

Could we fix the docs instead of changing the code though? Some lines in the documentation and tests rely on `samplers.greedy` being callable.

```
outlines/docs/reference/samplers.md:sampler = samplers.beam_search(beams=5)
tests/generate/test_integration_llamacpp.py:    sampler = samplers.beam_search(1)
tests/generate/test_integration_transformers.py:    sampler = beam_search(5)
tests/test_samplers.py:    beam_search,
tests/test_samplers.py:    assert beam_search == BeamSearchSampler
```

jrinder42:
I understand that there are tests written regarding this, but something doesn't seem to be working correctly. The first line of the classification docs is as follows:
```python
from outlines.samplers import greedy
```

This is where the issue begins. A potential solution would be to do the following in the docs:
```python
from outlines.samplers import BeamSearchSampler, GreedySampler, MultinomialSampler

generator = outlines.generate.choices(model, ["URGENT", "STANDARD"], sampler=GreedySampler())
```

However, then that begs the question of whether `greedy`, `multinomial`, and `beam_search` from `outlines.samplers` would be useless (I honestly have not looked to see if these are used in the rest of the codebase). I still think that some code should be changed, I suppose the question is what code / behavior should be changed...

jrinder42:
And one thing about the about a test like:
```python
tests/test_samplers.py:    assert beam_search == BeamSearchSampler
```
In this context, `==` and `isinstance` behave differently. There is an example demonstrating this in the description for issue #978. So the test above is technically not checking for the correct thing (unless I am misunderstanding something).

lapp0:
> I understand that there are tests written regarding this, but something doesn't seem to be working correctly. The first line of the classification docs is as follows:
> 
> ```python
> from outlines.samplers import greedy
> ```
> 
> This is where the issue begins. A potential solution would be to do the following in the docs:
> 
> ```python
> from outlines.samplers import BeamSearchSampler, GreedySampler, MultinomialSampler
> 
> generator = outlines.generate.choices(model, ["URGENT", "STANDARD"], sampler=GreedySampler())
> ```
> 
> However, then that begs the question of whether `greedy`, `multinomial`, and `beam_search` from `outlines.samplers` would be useless (I honestly have not looked to see if these are used in the rest of the codebase). I still think that some code should be changed, I suppose the question is what code / behavior should be changed...

`samplers.greedy` is a helper. It's equivalent to `samplers.GreedySampler`, but less verbose. I think we only need a docs change here.

> tests/test_samplers.py:    assert beam_search == BeamSearchSampler
> 
> In this context, == and isinstance behave differently. There is an example demonstrating this in the description for issue https://github.com/outlines-dev/outlines/issues/978. So the test above is technically not checking for the correct thing (unless I am misunderstanding something).

The test is correct, it's asserting that `beam_search` is a convenience variable equivalent to the class itself.

jrinder42:
Ok, sounds good. I will make the necessary changes.

jrinder42:
It should be good to go

jrinder42:
It should be good to go

lapp0 (approved):

rlouf:
Thank you!

PR #990, opened 2024-06-19
Fix os.environ use in docs

rshah713:


lapp0 (changes requested):
Good catch!

Could we have the `load_dotenv` portion in a separate code block labelled as optional?

rlouf:
> Good catch!

+1

> Could we have the `load_dotenv` portion in a separate code block labelled as optional?

Could we remove it instead? This adds confusion more than anything.

PR #996, opened 2024-06-21
Ensure `models.llamacpp` Doesn't Have Implicit `max_tokens`

### LINKED ISSUE #973: Invalid `generate.json` Output with `models.llamacpp`
Related: 
- https://github.com/outlines-dev/outlines/issues/952
- https://github.com/outlines-dev/outlines/issues/985

`models.llamacpp` used with `generate.json` can result in an output which isn't valid with the pydantic schema.

```python
# Reproducer from "Robert Roland Roger" on Discord:

import outlines
from pydantic import BaseModel, StringConstraints
from typing import Annotated
import llama_cpp

model = outlines.models.llamacpp(
    "TheBloke/zephyr-7B-beta-GGUF", 
    "zephyr-7b-beta.Q6_K.gguf",
    tokenizer=llama_cpp.llama_tokenizer.LlamaHFTokenizer.from_pretrained("HuggingFaceH4/zephyr-7b-beta")
)

class Character(BaseModel):
    name: Annotated[str, StringConstraints(max_length=10)]
    #age: int = Field(..., ge=18, lt=99)
    age: int
    #strength: int = Field(..., ge=1, lt=100)
    strength: int

prompt = """
<|system|>
You always output valid JSON.
<|user|>
Generate a new character in valid json. Use the following fields:
name, age (between 18 and 99), armor beween 1 and 100) and strength.
<|assistant|>
"""
character = generator(prompt)
```

### Expected result:

An output which pydantic recognizes as valid with the schema.

### Error message:

```shell
Expecting ',' delimiter: line 1 column 28 (char 27) [type=value_error.jsondecode, input_value='{ "name": "Elena", "age": 2', input_type=str]
```

Version information
<details>
```
outlines==0.0.44 
llama-cpp-python==0.2.78 
```
</details>

# Another Reproducer using `models.transformers_multimodal`

```
import outlines

model = outlines.models.transformers_multimodal(
    "llava-hf/llava-v1.6-mistral-7b-hf",
	device="cuda",
	model_kwargs=dict(torch_dtype=torch.bfloat16)
)

from typing import List, Optional

class ImageData(BaseModel):
    caption: str
    tags: List[str]
    location: Optional[str]
    objects: List[str]

image_data_generator = outlines.generate.json(model, ImageData)
image_data_generator(
    "<image> detailed JSON metadata:",
    "https://upload.wikimedia.org/wikipedia/commons/e/ea/FCAB_EMD_GT22CU-3_San_Pedro_-_Ascotan.jpg"
)
```

lapp0:
Fixes #973 

## Problem

In #973 an invalid pattern was being generated, however it was only invalid because it was incomplete. in the `LlamaCpp.generate()` call `llama_cpp_params["max_tokens"]` was unset, however the model was still finishing for reason "length":

```
[{'text': '{ "name": "Grace Lee", "age":32, "', 'index': 0, 'logprobs': None, 'finish_reason': 'length'}]                                                                                                         
```

In turns out the max tokens was being implicitly set to `30`.

## Solution

Allow 2**30 tokens if `max_length` is unset /`None`, because `llama_cpp_params` requires an `int` for the `max_tokens` field.

### DIFF
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -166,7 +166,9 @@ class LlamaCpp:
 
         # Somehow `llama-cpp-python` generates `max_tokens + 1`  tokens
         if "max_tokens" not in llama_cpp_params:
-            if max_tokens is not None:
+            if max_tokens is None:
+                llama_cpp_params["max_tokens"] = -1  # indicates unlimited tokens
+            else:
                 llama_cpp_params["max_tokens"] = max_tokens - 1
         else:
             llama_cpp_params["max_tokens"] = llama_cpp_params["max_tokens"] - 1
--- a/tests/generate/test_integration_llamacpp.py
+++ b/tests/generate/test_integration_llamacpp.py
@@ -356,3 +356,22 @@ def test_tokenizer_vocabulary_decode_sanity():
         ]
     )
     assert decoded_nl_token == vocab_nl_token
+
+
+def test_no_length_constraint_when_unset():
+    """Assert that models.llamacpp doesn't have an implicit max_tokens preventing full sequence generation"""
+    import llama_cpp
+
+    model = models.llamacpp(
+        repo_id="M4-ai/TinyMistral-248M-v2-Instruct-GGUF",
+        filename="TinyMistral-248M-v2-Instruct.Q4_K_M.gguf",
+        tokenizer=llama_cpp.llama_tokenizer.LlamaHFTokenizer.from_pretrained(
+            "Locutusque/TinyMistral-248M-Instruct"
+        ),
+    )
+
+    long_pattern = "abcdefg" * 10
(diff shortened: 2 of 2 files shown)


outlines/models/llamacpp.py
```diff
@@ -166,7 +166,9 @@ def prepare_generation_parameters(
 
         # Somehow `llama-cpp-python` generates `max_tokens + 1`  tokens
         if "max_tokens" not in llama_cpp_params:
-            if max_tokens is not None:
+            if max_tokens is None:
```
rlouf, line 169: Did you check `llama.cpp`’s code? They could be using `-1` to indicate there’s no limit.

lapp0: Great catch, `llama.cpp`'s behavior.

Issue #965, opened 2024-06-13
LlamaCpp doesnt work with generate.fsm for custom FSMs


Radu1999:
### Describe the issue as clearly as possible:

The example with custom fsm from documentation doesnt work for LlamaCpp as 
```
logits, kv_cache = model(token_ids, attention_masks, kv_cache)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'LlamaCpp' object is not callable
```

### Steps/code to reproduce the bug:

```python
from transformers import AutoTokenizer
from outlines import models, generate
from outlines.models.transformers import TransformerTokenizer
from llama_cpp import Llama
import interegular
import torch

if __name__ == "__main__":
    # Create model
    llm = Llama("./models/Mistral-7B-Instruct-v0.2/mistral-7b-instruct-v0.2.Q5_K_M.gguf")
    model = models.LlamaCpp(llm)
    model.tokenizer = TransformerTokenizer(AutoTokenizer.from_pretrained('mistralai/Mistral-7B-Instruct-v0.2', use_fast=True))
    model.device = 'cpu'

    # Create fsm
    list_of_strings_pattern = """\["[^"\s]*"(?:,"[^"\s]*")*\]"""
    pink_elephant_pattern = """.*(pink|elephant).*"""

    list_of_strings_fsm = interegular.parse_pattern(list_of_strings_pattern).to_fsm()
    pink_elephant_fsm = interegular.parse_pattern(pink_elephant_pattern).to_fsm()

    difference_fsm = list_of_strings_fsm - pink_elephant_fsm

    generator = generate.fsm(model, difference_fsm)
    rng = torch.Generator(device="cpu")
    rng.manual_seed(789005)

    response = generator("[INST] Don't talk about pink elephants [/INST]")
    print(response)
```

### Expected result:

```shell
I d expect it to work :)
```

### Outlines/Python version information:

Version information
latest

lapp0:
It will be a bit before [this](https://github.com/outlines-dev/outlines/pull/997) is merged into `main`, but you can try it early with

```
pip install --upgrade git+https://github.com/lapp0/outlines@fix-llamacpp-fsm
```

Works on my end, please let me know if you run into any issues!

PR #998, opened 2024-06-21
Improve `outlines.processors`, add integration tests to test_generate.py

lapp0:
A lot of these fixes were intended for https://github.com/outlines-dev/outlines/pull/966 however that's blocked until there's a new `transformers` release. 

These improvements are general to all models and will enable PRs resolving https://github.com/outlines-dev/outlines/issues/806 and https://github.com/outlines-dev/outlines/issues/965

# Structure of `OutlinesLogitsProcessor`

The goal is to create a base class which allows a logits processors to be implemented once and used for any `outlines.models` inference library.

To accomplish this we must normalize the input array. It must have a consistent type (`torch.Tensor`) and consistent dimensionality (2). We can normalize both of these simply, and without any copy operations.

`mlx.core.array`, `numpy.array`, and `torch.Tensor` all support [pythons array standard `__dlpack__`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html). This standard allows for casting between array types without copying.

`torch.Tensor` is the only input type which cannot always be cast to any other type because torch tensors may live in GPU memory. Therefore, we cast all arrays to `torch.Tensor`, implement logits processors using torch methods, and convert back to the original array type in `OutlinesLogitsProcessor`. See docstring of `OutlinesLogitsProcessor.__call__()` for more details.

# Detailed Changes
- Rename `BaseLogitsProcessor` to `OutlinesLogitsProcessor`
- Ensure `OutlinesLogitsProcessor.process_logits()` is always passed a 2D batch request with `torch.Tensor` logits and `List` input_ids. Also clean up code to be more readable in `OutlinesLogitsProcessor__call__()`
- Ensure `FSMLogitsProcessor` allows unstable sequence ordering (beam search in transformers and vLLM change the order of sequences)
- Update `tests/generate/test_generate.py` to cover more permutations of 
  - regex / text 
  - batch / single
  - greedy / multinomial / beam search
  - `stream()` / `generate()`
- Ensure performance stability with difference array libraries through `benchmark_processors.py`

### DIFF
--- a/outlines/processors/base_logits_processor.py
+++ b/outlines/processors/base_logits_processor.py
@@ -1,23 +1,30 @@
 from abc import abstractmethod
-from typing import List, Protocol, Union
+from typing import TYPE_CHECKING, List, Protocol, Type, Union
 
 import numpy as np
 import torch
 from numpy.typing import NDArray
 
+if TYPE_CHECKING:
+    import mlx.core as mx
 
-def is_mlx_array(logits):
+
+Array = Union[NDArray, torch.Tensor, List, "mx.array"]
+
+
+def is_mlx_array_type(array_type):
     try:
         import mlx.core as mx
     except ImportError:
         return False
-    return isinstance(logits, mx.array)
+    return issubclass(array_type, mx.array)
 
 
-class BaseLogitsProcessor(Protocol):
+class OutlinesLogitsProcessor(Protocol):
     """
     Base class for logits processors which normalizes types of logits:
     - ndarray (used by llama-cpp-python), converted to torch.Tensor
+    - mlx.core.array (used by mlx-lm), converted to torch.Tensor
     - torch.Tensor (used by everything else)
 
     Normalization of types and conversion to torch.Tensor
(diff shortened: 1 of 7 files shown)


outlines/processors/base_logits_processor.py
```diff
@@ -14,10 +14,11 @@ def is_mlx_array(logits):
     return isinstance(logits, mx.array)
 
 
-class BaseLogitsProcessor(Protocol):
+class OutlinesLogitsProcessor(Protocol):
```
rlouf, line 17: Why not call it `LogitsProcessor`?

outlines/processors/base_logits_processor.py
```diff
@@ -29,50 +30,100 @@ class BaseLogitsProcessor(Protocol):
...
-                return mx.array(logits_float32_numpy)
-
-            else:
-                raise TypeError(
-                    "LogitsProcessor must be called with either np.NDArray"
-                    ", torch.Tensor, or mlx.core.array typed logits"
-                )
+
+        # ensure logits are torch Tensors
+        torch_logits = self._to_torch(logits)
```
rlouf, line 65: What is the cost of this conversion? We need to profile this.

outlines/processors/base_logits_processor.py
```diff
@@ -29,50 +30,100 @@ class BaseLogitsProcessor(Protocol):
...
+        - logits processors are only used once and never re-applied for a new sequence generator
+        - Some models only pass output_ids, some models such as llamacpp and transformers prefix with input_ids
+        - Some sampling methods, such as beam search, result in unstable sequence ordering in models like vLLM
+        """
+        pass
 
+    @torch.no_grad()
     def __call__(
         self,
         input_ids: Union[NDArray[np.int64], List[int], torch.Tensor],
```
rlouf, line 51: It's probably time to create a type alias that include the MLX tensor type.

outlines/processors/structured.py
```diff
@@ -63,12 +63,13 @@ def __init__(self, tokenizer: "Tokenizer", fsm: Guide):
             The finite state machine which is used to bias the logits.
         """
         self.tokenizer = tokenizer
-        self._fsm_state = 0
+        self._fsm_states: Dict[int, int] = {}
         self.fsm: Guide = fsm
         self._is_first_token = True
+        self._seq_start_idx: Optional[int] = None
 
     def process_logits(
```
rlouf, line 71: cf above remark on dispatching the function depending on the type of the tensors.

outlines/processors/base_logits_processor.py
rlouf: Actually, do we really need that class if all logit processors are ever going to be `FSMLogitsProcessors`

rlouf (changes requested):
I would not convert back-and-forth to `torch` arrays and instead dispatch the logic depending on the type of `input_ids` and `logits` so as to not add a performance penalty during inference.

outlines/processors/base_logits_processor.py
lapp0: The idea is that we can cast `numpy` and `mlx.core` arrays to `torch.Tensor` using shared memory, **not** a copy operation :)

Then we can implement a single `OutlinesLogitsProcessor` class with a single `process_logits()` method and it works out of the box for any library. While not in `main` yet, I've tested `FSMLogitsProcessor`, and `RegexLogitsProcessor` with every nearly `outlines.models` option and they work with no additional changes.

This makes the processors easy to maintain and implement.

outlines/processors/base_logits_processor.py
lapp0: My thinking here was if we want a new logits processor which isn't for structured generation, e.g. `RepetitionPenaltyLogitsProcessor`. We can easily implement this processor once and use it in every `outlines.models` without any changes specific to the model.

This provides us the flexiiblity to have the inference library handle the decoder pass and sampling, and `outlines` can handle all logits augmentation.

Another example: `mlxlm` doesn't support `stop_strings` and `stop_strings` is broken in the latest `transformers` version (`4.41.2`). We could implement a `StopStringsLogitsProcessor` and make it available to all inference libraries. Not arguing that this should be a priority, but I am arguing that this provides us flexibility and opportunities.

lapp0:
> I would not convert back-and-forth to `torch` arrays and instead dispatch the logic depending on the type of `input_ids` and `logits` so as to not add a performance penalty during inference.

Here are the benchmark results:

```
[50.00%] ··· Running (bench_processors.LogitsProcessorBenchmark.time_logits_processor--).
[100.00%] ··· bench_processors.LogitsProcessorBenchmark.time_logits_processor                                                                                                                                   ok
[100.00%] ··· ======== ==========
               param1            
              -------- ----------
               torch    150±20μs 
               numpy    161±7μs  
                mlx     195±20μs 
              ======== ==========
```

The performance penalty for casting `mlx.core.array` -> `torch.Tensor` -> `mlx.core.array` is 45 nanoseconds.

PR #1004, opened 2024-06-24
Adding the option to avoid displaying tqdm bars at inference with `vllm`

BIMM99:
Very minor change that adds the possibility to avoid displaying progress bars within `outlines`. It is not changing the default behaviour and is just about adding an argument (`use_tqdm`) to the `LLM.generate` call.

It was useful for a personal project and I did not find another way to do so, so I decided to PR this.


rlouf:
Thank you for contributing!

BIMM99:
Pleasure wasn't too much work tbh ! :)

PR #1008, opened 2024-06-28
Add asknews example

robcaulk:
AskNews is processing 300k articles per day on-premise with vLLM and outlines. Outlines enables much deeper analysis of each article, since the LLM doesnt need to worry about structure. Instead, it can focus attention on important output like extracting "Reporting Voice" typed as a `Literal[str]`.

Thanks to the maintainers of this project!


rlouf:
Thank you for contributing!

Issue #1009, opened 2024-06-29
Exllamav2 integration


isamu-isozaki:
### Presentation of the new feature

Currently, for exllamav2, we have an outlines model that handles it. However, in the recent update to exllamav2, a dynamic generator was added which is not compatible with current outline's generate method as it does paged attention/radix cache etc. However, exllamav2 has a way to do constrained generation with this which is called filters. A demonstration of this with lm format enforcer is [here](https://github.com/turboderp/exllamav2/blob/master/examples/inference_json.py). The idea of this feature is to do the same thing by moving outlines logic to filters like in the integration scripts for other codebases.

### Where does it fit in Outlines?

For the current lm format enforcer+exllamav2, the constrained generation is not reliable/a bit buggy and I noticed using outlines does solve this issue.

### Are you willing to open a PR?
I already made one.

Thanks for contributing!

Issue #807, opened 2024-04-12
Update the `exllamav2` integration


rlouf:
Similarly to #772 and #782 use logits processors instead of the custom sampling loop.

Issue #655, opened 2024-02-13
Serve Any Model


lapp0:
### What behavior of the library made you think about the improvement?

Only vLLM can be served.

### How would you like it to behave?

Our serve endpoint should be made generic through the following steps
- Implement `outlines.models.vllm`
- Refactor `outlines.serve` so it takes a model and model kwargs as command arguments
- Refactor `outlines.serve` so it accepts any `outlines.generate` calls (`choice`, `json`, `cfg`, `regex`, `text`), additionally the sampler should be allowed to be specified

We also should get commentary from those who are currently using the vllm serve, as we don't want to disable any vllm specific features they are currently using without ensuring outlines has those capabilities available.

Additionally we should keep in mind the functionality requested in https://github.com/outlines-dev/outlines/issues/597

lapp0:
A default implementation should be built which operates asynchronously through a task queue. This setup can then overridden for inference engines, such as vLLM, which have built-in support for async processing.

kelsey-sorrels:
In a research setting, it's really helpful to have model and model kwargs as request parameters. Yes, there are many implications to such a design, but the ability to compare models without having to redeploy is a huge QoL improvement in the model-selection/contrast/ablation phase.

rlouf:
This is a very good point @kelsey-sorrels. What parameters do you usually have to pass to the model?

kelsey-sorrels:
I usually set
- trust_remote_code
- pipeline_parallel_size
- tensor_parallel_size
- quantization

But I want to be clear, those aren't settings that I'm usually varying with model_ids, they're simply incidental.
  
I will also be doing some LoRA-related work in the future that may require additional settings to be set, but I couldn't say what they will be until I do the task.

SuperMasterBlasterLaser:
Currently SGLang has ability to host models on remote servers while your own local code works as thin client. However, it has it's own problems.

PR #1010, opened 2024-06-29
Exllamav2 Integration

isamu-isozaki:
This fixes https://github.com/outlines-dev/outlines/issues/1009
Also fixes https://github.com/outlines-dev/outlines/issues/807

The tests I did were:

For loading:
```python
from outlines.integrations.exllamav2 import RegexFilter, TextFilter, JSONFilter, ChoiceFilter
import json
import torch
from exllamav2.generator.filters import ExLlamaV2PrefixFilter
from typing import Literal
from exllamav2 import(
    ExLlamaV2,
    ExLlamaV2Config,
    ExLlamaV2Cache,
    ExLlamaV2Cache_8bit,
    ExLlamaV2Cache_Q4,
    ExLlamaV2Tokenizer,
)

from exllamav2.generator import ExLlamaV2DynamicGenerator, ExLlamaV2Sampler, ExLlamaV2DynamicJob
import uuid

repo_id = "../Phi-3-mini-128k-instruct-exl2"
paged = False
model_dir = repo_id
total_context = 8192
max_context = 1024 
max_batch_size = 4 if paged else 1
max_chunk_size = 1024
max_new_tokens = 1024
healing = True
draft_model = None
draft_cache = None
use_ngram_draft = None
use_ngram = None

config = ExLlamaV2Config(model_dir)
config.max_input_len = max_chunk_size
config.max_attention_size = max_chunk_size ** 2

config.max_seq_len = max_context
model = ExLlamaV2(config)

cache = ExLlamaV2Cache_Q4(
    model,
    max_seq_len = total_context,
    lazy = True
)
tokenizer = ExLlamaV2Tokenizer(config)
hf_tokenizer_kwargs = {}
hf_tokenizer_kwargs.setdefault("padding_side", "left")
hf_tokenizer = AutoTokenizer.from_pretrained(model_dir, **hf_tokenizer_kwargs)
model.load_autosplit(cache, progress = True)
generator = ExLlamaV2DynamicGenerator(
    model = model,
    cache = cache,
    draft_model = draft_model,
    draft_cache = draft_cache,
    tokenizer = tokenizer,
    max_batch_size = max_batch_size,
    use_ngram_draft = use_ngram,
    max_chunk_size = max_chunk_size,
    paged = paged,
)
```

Choices test:
```python
filters = [
    ChoiceFilter(["bob", "fred"], hf_tokenizer)
]
context_ids = torch.empty((1, 0), dtype = torch.long)


instruction = "Who is better bob or fred?"
print()
print("Assistant:", end = "")

instruction_ids = tokenizer.encode(f"[INST] {instruction} [/INST]", add_bos = True)
context_ids = torch.cat([context_ids, instruction_ids], dim = -1)

generator.enqueue(
    ExLlamaV2DynamicJob(
        input_ids = context_ids,
        max_new_tokens = 1024,
        stop_conditions = [],
        filters=filters
    )
)

eos = False
while not eos:
    results = generator.iterate()
    for result in results:
        if result["stage"] == "streaming":
            eos = result["eos"]
            if "text" in result:
                print(result["text"], end="")
                sys.stdout.flush()
            if "token_ids" in result:
                context_ids = torch.cat([context_ids, result["token_ids"]], dim = -1)

print()
```
Returns 
```
Assistant:bob
```

Json test
```python
class JSONResponse(BaseModel):
    response: str
    confidence: Literal["low", "medium", "high"]
    is_subjective: Literal["no", "yes", "possibly"]
filters = [
    JSONFilter(JSONResponse, hf_tokenizer)
]
context_ids = torch.empty((1, 0), dtype = torch.long)


instruction = f"Give a sample response in the format of {JSONResponse.schema()} on a movie review of love actually"
print()
print("Assistant: ", end = "")

instruction_ids = tokenizer.encode(f"[INST] {instruction} [/INST]", add_bos = True)
context_ids = torch.cat([context_ids, instruction_ids], dim = -1)

generator.enqueue(
    ExLlamaV2DynamicJob(
        input_ids = context_ids,
        max_new_tokens = 1024,
        stop_conditions = [tokenizer.eos_token_id],
        filters=filters
    )
)

eos = False
while not eos:
    results = generator.iterate()
    for result in results:
        if result["stage"] == "streaming":
            eos = result["eos"]
            if "text" in result:
                print(result["text"], end="")
                sys.stdout.flush()
            if "token_ids" in result:
                context_ids = torch.cat([context_ids, result["token_ids"]], dim = -1)

print()

```
Returns
```
Assistant: {"response": "Love Actually is a charming and heartwarming romantic comedy that delivers a delightful experience. The performances by the lead actors, especially Drew Barrymore and Gael García Bernal, are genuinely commendable. The film beautifully blends humor with heart-tugging moments, making it an ideal watch for those in search of a feel-good cinematic experience. Despite some predictable plot trends, the overall impact of the film remains largely positive. Rating: 7/10", "confidence": "medium", "is_subjective": "no"}
```


isamu-isozaki:
Some questions I had for maintainers were
1. Should we do the prefix logic here? I noticed that in some exllamav2 filters in their repo the prefix is ignored but for one they are used.
2. Do we want to return the stop tokens? It requires us to check through all the allowed tokens to see which one gives a final state. This may be a bit slower

lapp0 (changes requested):
Sorry for the late delay and request for refactor.

We've been moving towards using `SequenceGeneratorAdapter` and `outlines.processors` in `outlines.generate`. Currently the only local `outlines.model` which doesn't have a `SequenceGeneratorAdapter` based implementation is `exllamav2`.

Would you be able to refactor this to use `SequenceGeneratorAdapter` instead?

This would involve
- Adding `ExLlamaV2` to the `*_unified` dispatcher https://github.com/outlines-dev/outlines/blob/main/outlines/generate/regex.py#L42-L53
- Ensuring the passed `OutlinesLogitsProcessor` is applied when exllamav2 `generator.generate(prompt)` is called
- Adding an `ExLlamaV2` fixture to `tests/generate/test_generate.py` which will automatically test all generation methods (structured, batch, stream, etc) against the model [here](https://github.com/outlines-dev/outlines/blob/main/tests/generate/test_generate.py#L73)

isamu-isozaki:
@lapp0 make sense! Let me try doing this tomorrow

lapp0:
Thanks so much, please let me know if you have any questions!

isamu-isozaki:
@lapp0 sry for delay! Two questions
Background:
The current exllamav2 model in outlines(with ExllamaV2 class) as can be seen [here](https://github.com/turboderp/exllamav2/blob/master/exllamav2/model.py), doesn't support filters(which is exllamav2's logitsprocessor). The filters are mainly used in exllamav2's custom generators like ExLlamaV2DynamicGenerator, ExLlamaV2DynamicGeneratorAsync and I think more of them.
So my questions are
1. Do we want to use the logits_processor in SequenceGeneratorAdapter to be converted to an exllamav2 filter like in their library(like logits_processor in llamacpp)? In this case this might involve changing the logic [here](https://github.com/outlines-dev/outlines/blob/main/outlines/models/exllamav2.py) to use one of the generators. Another option is using [ExLlamaV2Sampler](https://github.com/turboderp/exllamav2/blob/master/exllamav2/generator/sampler.py) but we will be redoing the generator logic in this case.
2. This depends on the previous question but in this case, do you have a recommended generator? I have mainly used the ExLlamaV2DynamicGenerator which is mainly used for handling multiple asynchronous requests and responses which is not necessarily what I think outlines is going for ex (one request, one response at a time). But it seems like the most well-supported generator in exllamav2.

Sry for the delayed response and let me know if it makes sense/what you think!

lapp0:
Great questions!

Converting it to a filter is a bit hacky IMO, but may be the simplest solution and doesn't require an upstream change.

Alternatively we could apply logits processing directly. The way `exllamav2`s library is structured makes this a tricky. `ExLlamaV2Sampler.sample()` is a staticmethod, and `gen_settings: ExLlamaV2Sampler.Settings` is a `generate(...)` argument, however the sampler itself is not. I think the only clean way to handle this is an upstream PR:

- Option 1: update `ExLlamaV2Sampler.Settings` to accept a `logits_processor` argument, and update `def sample()` to apply the `settings.logits_processor` if it exists.
- Option 2: update `ExLlamaV2DynamicGenerator.generate()` to accept a `sampler` argument, defaulting to `ExLlamaV2Sampler`, allowing us to inject our own sampler class.

The first option makes more sense to me, it is generator-class agnostic. 

>This depends on the previous question but in this case, do you have a recommended generator? I have mainly used the ExLlamaV2DynamicGenerator which is mainly used for handling multiple asynchronous requests and responses which is not necessarily what I think outlines is going for ex (one request, one response at a time). But it seems like the most well-supported generator in exllamav2.

Tbh, I'm not sure how well `outlines.models` works with asyncronous structured generation.  It is a reasonable use case though, and necessary for https://github.com/outlines-dev/outlines/issues/655

isamu-isozaki:
@lapp0 sry one more q(we can move conv to discord if it is more convenient!), for exllamav2's sampler design, they, before computing the logits,
1. Get the next possible tokens to use in the form of the list given ex the current fsm
2. Apply it using their cuda code like

isamu-isozaki:
@lapp0 I'm going with option 1 as you said but sry one more q(we can move conv to discord if it is more convenient!), for exllamav2's sampler design, they, before computing the logits,
1. Get the next possible tokens to use in the form of the list given ex the current fsm
2. Apply it using their cuda code like 
```python
if pass_tokens is not None:
    assert pass_tokens, "Filter excluded all tokens"
    if filter_prefer_eos and tokenizer.eos_token_id in pass_tokens:
        pass_tokens = { tokenizer.eos_token_id }
    ext_c.logit_filter_exclusive(logit_filter, [sorted(list(pass_tokens))])
```
3. Then finally the logits are computed.
The question I have is the current OutlinesLogitsProcessor expects 
1. logits to be there
2. To create/output the torch mask which may be unnecessary processing in terms of exllamav2
In this case, I think I'll have to make a new LogitsProcessor for exllama2 which only shows the allowed tokens, but am I correct? Of course, we can do something like putting in torch.ones as logits and then recover allowed tokens from the mask but I believe that'll be a bit too hacky.

isamu-isozaki:
@lapp0 Sounds good! I think I'll go with option 1. For this, I think the steps needed are

- [ ] Make a new OutlinesLogitsProcessor inherited class like structured but do not return mask and only return the next available tokens. I might even override the __call__ since conversion to pytorch might not be necessary(similar to filters)
- [ ] To models/exllamav2 add in class ExllamaV2SamplerOutlines(ExLlamaV2Sampler) with optional logits processor
- [ ] See if I can convert current exllamav2 base model forward logic etc to ExllamaV2DynamicGenerator
- [ ] Add to the unified dispatcher
- [ ] Test code
Let me know if it looks good. I'll try finishing this within the week

lapp0:
Rather than implementing a new logits processor, I'm awaiting correspondence with the ExLlamaV2 maintainer, turboderp, regarding whether a `logits_processor` argument would be acceptable within their sampler.

isamu-isozaki:
@lapp0 interesting! The main reason I was thinking of a new logits processor is because we do some redundant steps in terms of exllamav2's code base I thought. In that for them, they first
1. Get the passed tokens(the next allowed tokens) and then apply that filter using cuda code

lapp0:
>So I thought some of the steps here overlap with our current logits processor. But yeah very much happy to get advice here since this is just making the exllamav2 filter. And also happy to hear what turboderp thinks.

Yes, they will have multiple methods of filtering, but given Outlines singular logits processor implementation, which is tested against all inference engines, it's likely better to follow the same pattern with ExLlamaV2. This will ensure bug fixes and optimizations present in one integration are available to all integrations!

I spoke with turboderp on their discord server, he is open to having a `logits_processor` argument in `ExLlamaV2Sampler.Settings`. 

Here's the steps I think we should take, let me know what you think:
- 1) Update `outlines.generate.*` generators so they use the default dispatcher for `ExLlamaV2` (simply delete [the ExLlamaV2 dispatcher](https://github.com/outlines-dev/outlines/blob/main/outlines/generate/regex.py#L52-L63) in each `outlines.generate` module) This will ensure [the default](https://github.com/outlines-dev/outlines/blob/main/outlines/generate/regex.py#L13-L37) method of `SequenceGeneratorAdapter` and `outlines.processors` is used.
- 2) Implement a `turboderp/exllamav2` fork with a `logits_processor` argument in `ExLlamaV2Sampler.Settings` which is applied in `ExLlamaV2Sampler.sample()` (let me know if you'd like to take this over, or if you'd like me to take a shot at it)
- 3) Implement a new model `outlines.models.exllamav2` which is compatible with the fork 
- 4) Test it against `outlines.models.exllamav2` by adding an exllamav2 fixture to `tests/generate/test_generate.py` and running `pytest -s tests/generate/test_generate.py -k exllamav2`

Let me know if you think this is the right path.

Thanks so much for your great work on this PR. The users in the ExLlamaV2 discord were excited to hear about this PR!

isamu-isozaki:
@lapp0 wow, didn't know exllamav2 had a discord server! And makes perfect sense. 
If you can do ii that'll be awesome since I was thinking of this and I couldn't think of a clean way to do it atm.
For iii. sounds good. I'll try converting it to the dynamic generator

lapp0:
@isamu-isozaki can you please take a look at this changeset and the provided example `json_schema_outlines.py`?

https://github.com/lapp0/exllamav2/pull/1

I believe it should provide a sufficient basis for implementing `outlines.models.exllamav2`. 

Let me know if you see anything that should be changed in my implementation. If you have any questions, please do not hesitate! Good luck!

isamu-isozaki:
@lapp0 sounds good. And sorry got a bit side tracked by some work. I'll try get to this at least by the weekend. Sorry for delay!

isamu-isozaki:
Sorry for the delay, I finally got the exllamav2 fork built and I was able to run the current pr's code with below which worked!
```python
import sys
sys.path.append("../outlines-dev")
import outlines

from enum import Enum
from pydantic import BaseModel, constr

model = outlines.models.exl2(
    model_path="turboderp/TinyLlama-1B-32k-exl2",
    cache_q4=True,
    paged=False
)

prompt = """You are a sentiment-labelling assistant.
Is the following review positive or negative?

Review: This restaurant is just awesome!
"""

generator = outlines.generate.choice(model, ["Positive", "Negative"])
answer = generator(prompt)
print(answer)

prompt = "<s>result of 9 + 9 = 18</s><s>result of 1 + 2 = "
answer = outlines.generate.format(model, int)(prompt, max_tokens=1)
print(answer)

generator = outlines.generate.format(model, float)
answer = generator(prompt, max_tokens=10)
print(answer)

unstructured = generator(prompt, max_tokens=30)

    model,
)

print(unstructured)
# What is the IP address of the Google DNS servers?
#
# Passive DNS servers are at DNS servers that are private.
# In other words, both IP servers are private. The database
# does not contain Chelsea Manning

print(structured)

class Weapon(str, Enum):
    sword = "sword"
    axe = "axe"
    mace = "mace"
    spear = "spear"
    bow = "bow"
    crossbow = "crossbow"


class Armor(str, Enum):
    leather = "leather"
    chainmail = "chainmail"
    plate = "plate"


class Character(BaseModel):
    name: constr(max_length=10)
    age: int
    armor: Armor
    weapon: Weapon
    strength: int


# Construct structured sequence generator

# Draw a sample
seed = 789001

character = generator("Give me a character description", seed=seed)

print(repr(character))
# Character(name='Anderson', age=28, armor=<Armor.chainmail: 'chainmail'>, weapon=<Weapon.sword: 'sword'>, strength=8)

character = generator("Give me an interesting character description", seed=seed)

print(repr(character))
# Character(name='Vivian Thr', age=44, armor=<Armor.plate: 'plate'>, weapon=<Weapon.crossbow: 'crossbow'>, strength=125)
```

isamu-isozaki:
The current main issue is that I can't seem to run the tests due to some error with the pyairports. @lapp0 do you have some advice on how to fix this?
```bash
 pytest -s tests/generate/test_generate.py -k exllamav2
======================== test session starts =========================
platform linux -- Python 3.10.12, pytest-8.3.2, pluggy-1.5.0
rootdir: /mnt/d/personal_projects/whiterabbitneo-pentestgpt/outlines-dev
configfile: pyproject.toml
plugins: anyio-3.6.2
collected 0 items / 1 error

=============================== ERRORS ===============================
__________ ERROR collecting tests/generate/test_generate.py __________
tests/generate/test_generate.py:6: in <module>
    import outlines.generate as generate
outlines/__init__.py:6: in <module>
    import outlines.types
outlines/types/__init__.py:1: in <module>
    from . import airports, countries
outlines/types/airports.py:4: in <module>
    from pyairports.airports import AIRPORT_LIST
/home/isamu/miniconda3/lib/python3.10/site-packages/pyairports/airports.py:1: in <module>
    from pkg_resources import resource_string
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:3663: in <module>
    def _initialize_master_working_set():
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:3646: in _call_aside
    f(*args, **kwargs)
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:3687: in _initialize_master_working_set
    tuple(dist.activate(replace=False) for dist in working_set)
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:3687: in <genexpr>
    tuple(dist.activate(replace=False) for dist in working_set)
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:3144: in activate
    declare_namespace(pkg)
/home/isamu/miniconda3/lib/python3.10/site-packages/pkg_resources/__init__.py:2542: in declare_namespace
    warnings.warn(msg, DeprecationWarning, stacklevel=2)
E   DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
E   Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
====================== short test summary info =======================
ERROR tests/generate/test_generate.py - DeprecationWarning: Deprecated call to `pkg_resources.declare_nam...
!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!
========================= 1 error in 17.91s ==========================
```

lapp0:
@isamu-isozaki sorry for the delayed response.

`pyairports` is an annoying library which has caused a lot of issues for me as well. And the only thing we use the library for is loading the 3 letter airport code list from https://github.com/ozeliger/pyairports/blob/f611ee5a5a82b4e98b22641bb99693d862c802e4/pyairports/data/airport_list.json

A quick and easy hack is to remove the import and run tests again.

isamu-isozaki:
Thanks! For tests
```
outlines-dev# pytest -s tests/generate/test_generate.py -k exllamav2 -x
============================================ test session starts =============================================
platform linux -- Python 3.10.12, pytest-8.3.2, pluggy-1.5.0
rootdir: /mnt/d/personal_projects/whiterabbitneo-pentestgpt/outlines-dev
configfile: pyproject.toml
collected 320 items / 288 deselected / 32 selected
collecting 0 items
Loading: blockblockblock/TinyLlama-1.1B-Chat-v1.0-bpw4.6-exl2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:01:00 0:00:00
Loading tokenizer...
Compiling FSM index for all state transitions: 100%|████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 19.81it/s]
Compiling FSM index for all state transitions: 100%|███████████████████████████████████████████████████████████| 25/25 [00:00<00:00, 113.21it/s]
Compiling FSM index for all state transitions: 100%|████████████████████████████████████████████████████████████| 21/21 [00:00<00:00, 88.63it/s]
.s......................

========================================== 31 passed, 1 skipped, 288 deselected in 117.40s (0:01:57) ===========================================
```
pre-commit result
```
outlines-dev> pre-commit run --all-files
check for merge conflicts................................................Passed
debug statements (python)................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
isort....................................................................Passed
pyupgrade................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

outlines/generate/regex.py:3:1: F401 'outlines.fsm.guide.RegexGuide' imported but unused
outlines/generate/regex.py:4:1: F401 'outlines.generate.api.SequenceGenerator' imported but unused
outlines/generate/regex.py:9:1: F401 'outlines.models.ExLlamaV2Model' imported but unused
outlines/generate/text.py:4:1: F401 'outlines.generate.api.SequenceGenerator' imported but unused
outlines/generate/text.py:9:1: F401 'outlines.models.ExLlamaV2Model' imported but unused

black....................................................................Passed
mypy.....................................................................Passed
```

isamu-isozaki:
@lapp0 Thanks! I think I'm pretty much done:
tests
```
outlines-dev# pytest -s tests/generate/test_generate.py -k exllamav2 -x
============================================================= test session starts ==============================================================
platform linux -- Python 3.10.12, pytest-8.3.2, pluggy-1.5.0
rootdir: /mnt/d/personal_projects/whiterabbitneo-pentestgpt/outlines-dev
configfile: pyproject.toml
plugins: anyio-3.6.2
collected 320 items / 288 deselected / 32 selected

Loading: blockblockblock/TinyLlama-1.1B-Chat-v1.0-bpw4.6-exl2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:48 0:00:00
Loading tokenizer...
Compiling FSM index for all state transitions: 100%|████████████████████████████████████████████████████████████| 10/10 [00:00<00:00, 41.80it/s]
Compiling FSM index for all state transitions: 100%|███████████████████████████████████████████████████████████| 25/25 [00:00<00:00, 116.97it/s]
Compiling FSM index for all state transitions: 100%|████████████████████████████████████████████████████████████| 21/21 [00:00<00:00, 83.39it/s]
.s......................

========================================== 31 passed, 1 skipped, 288 deselected in 102.91s (0:01:42) ===========================================
```
and pre-commit:
```
outlines-dev> pre-commit run --all-files
check for merge conflicts................................................Passed
debug statements (python)................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
isort....................................................................Passed
pyupgrade................................................................Passed
flake8...................................................................Passed
black....................................................................Passed
mypy.....................................................................Passed
```

isamu-isozaki:
@lapp0 sry 2 questions
1. Should I modify the pyproject.toml to install exllamav2 or do you recommend skipping the tests(like diffusers)?
2. How do I add models/is that done by the org side maybe?

lapp0:
>Should I modify the pyproject.toml to install exllamav2 or do you recommend skipping the tests(like diffusers)?

We should allow it in the test dependencies for any platform which supports `exllamav2`

https://github.com/outlines-dev/outlines/blob/main/pyproject.toml#L59-L63

>How do I add models/is that done by the org side maybe?

Sorry, could you please clarify?

isamu-isozaki:
@lapp0 Thanks for your reply! I think I was a bit confused on how to add models but I'm guessing blockblockblock/TinyLlama-1.1B-Chat-v1.0-bpw4-exl2 is already in the test environment?
Also, sorry currently figuring out how to run exllamav2 on the CPU for the tests as currently the tests seem to be failing because of that

lapp0:
> I think I was a bit confused on how to add models but I'm guessing blockblockblock/TinyLlama-1.1B-Chat-v1.0-bpw4-exl2 is already in the test environment?

Yes it's already in the test environment. 

If you're interested, after this change, a good upstream change would be allowing model loading via hub uri with `hf_hub_download`.

> Also, sorry currently figuring out how to run exllamav2 on the CPU for the tests as currently the tests seem to be failing because of that

Don't bother. The hardware isn't supported. Please just add exllamav2 to https://github.com/outlines-dev/outlines/blob/main/tests/generate/conftest.py#L23-L33

This will skip the tests on any machine without CUDA.

isamu-isozaki:
@lapp0 Got it and thanks! I think I'm only missing coverage which I'll try making tests for once I get time

remichu-ai:
Hi, just want to pop by and see how it is going. Will this feature be released soon? If there is some dev branch i can try it as well.

lapp0:
>@lapp0 Got it and thanks! I think I'm only missing coverage which I'll try making tests for once I get time

Great, please let me know when you're ready for review!

>Hi, just want to pop by and see how it is going. Will this feature be released soon? If there is some dev branch i can try it as well.

You might be able to get it working with the installation commands below. Please report back with any issues or feedback, it will help with this PR!

```
pip install git+https://github.com/isamu-isozaki/outlines@exllamav2_filter
pip install git+https://github.com/lapp0/exllamav2@sampler-logits-processor
````

**Comment by isamu-isozaki** [comment]
@remichu-ai Hi! If you had an issue building exllamav2 like me you can just install outlines with my initial commit to this pr and you can use the code examples and it should work.
However, I did hear some issues with the speed of inference if you have a bad CPU in this case. I'm not sure how much more performant the current latest commit is.
You can def use this branch to test it out since the main thing left is just writing tests etc and not much for functionality

**Comment by isamu-isozaki** [comment]
@lapp0 hi! Sorry for more qs. I did write some tests to attempt to fill up the exllamav2.py coverage to 100% locally but it seems like if the tests are skipped they don't count towards coverage. Do you happen to know a simple way to fix this by any chance?
Other than this I think I'm ready for review!

File: outlines/models/exllamav2.py [hunk 1]
```diff
@@ -1,121 +1,217 @@
...
+
+        if len(prompts) == 1:
+            prompts = prompts[0]
 
-        self.past_seq = seq_tensor
+        return exllamav2_params, prompts
 
-        return self.model.forward(
-            seq_tensor[-1:].view(1, -1), self.cache, loras=[self.lora]
+    def reformat_output(
```
lapp0, line 96: why is `reformat_output` necessary? Could we get some docstrings and comments here and throughout. Especially for details which might not be understood by those making changes in the future?

outlines/models/exllamav2.py
```diff
@@ -1,121 +1,217 @@
...
+            generation_parameters,
+            sampling_parameters,
+            structure_logits_processor,
+        )
+        output = self.generator.generate(
+            prompt=prompts,
+            gen_settings=exllamav2_params["gen_settings"],
+            max_new_tokens=min(exllamav2_params["max_new_tokens"]),
+            completion_only=True,
+            encode_special_tokens=False,
```
lapp0, line 136: Why don't we encode special tokens?

outlines/models/exllamav2.py
```diff
@@ -1,121 +1,217 @@
...
+        exllamav2_params, prompts = self.prepare_generation_parameters(
+            prompts,
+            generation_parameters,
+            sampling_parameters,
+            structure_logits_processor,
+        )
+        output = self.generator.generate(
+            prompt=prompts,
+            gen_settings=exllamav2_params["gen_settings"],
+            max_new_tokens=min(exllamav2_params["max_new_tokens"]),
```
lapp0, line 134: was this intended to be `min(max new tokens, model max tokens)`?

outlines/models/exllamav2.py
```diff
@@ -162,71 +256,66 @@ def exl2(
...
             ExLlamaV2Config,
+            ExLlamaV2Tokenizer,
         )
+        from exllamav2.generator import ExLlamaV2DynamicGenerator
         from transformers import AutoTokenizer
+
     except ImportError:
         raise ImportError(
-            "The `exllamav2`, `transformers` and `torch` libraries needs to be installed in order to use `exllamav2` models."
+            "The `exllamav2` and `torch` libraries needs to be installed in order to use `exllamav2` models."
```
lapp0, line 268: Transformers needs to be installed as well.

outlines/models/exllamav2.py
```diff
@@ -162,71 +256,66 @@ def exl2(
...
-        print(" -- Loading model...")
-        model.load_autosplit(cache)
-
-    return ExLlamaV2Model(model, tokenizer, device, cache)
+        cache = ExLlamaV2Cache(model, max_seq_len=max_seq_len, lazy=True)
+    model.load_autosplit(cache, progress=True)
+
+    print("Loading tokenizer...")
+    tokenizer = ExLlamaV2Tokenizer(config)
+    tokenizer.vocabulary = tokenizer.extended_piece_to_id
```
lapp0, line 287: Could you implement a function which converts an EXL2 tokenizer to a `TransformerTokenizer` rather than patching here?

It would ensure downstream users don't need to define their own function which applies the same logic.

lapp0 (commented):
My example script works with the code. Minor change requests, great work!

isamu-isozaki:
@lapp0 Thanks for review! Let me check it out tomorrow

outlines/models/exllamav2.py
isamu-isozaki: @lapp0 Ah so it's necessary mainly because for exllamav2, it mainly accepts only a list or a string then
1. If the prompt is a string, return a string
2. If a prompt is a list, return a list. This is not the same as outlines output in that if the list is only one element, the string is expected to be outputted.
4. No such thing as num_samples, so I had to duplicate the prompt by num_samples times and have this function output a list of lists
So in summary, it's to fix the output from exllamav2 to be in outlines/llamacpp form. I can add this as a comment if that works!

outlines/models/exllamav2.py
isamu-isozaki: @lapp0 Ah, sorry for the confusion. For exllamav2, it needs the max amount of new tokens generated. The reason exllamav2_params["max_new_tokens"] is a list is because I calculated on line 53 the max amount of tokens that can be generated by the model for each prompt(by encoding with tokenizer).  I picked the minimum as otherwise it might be possible for one of the prompts to exceed the max sequence length.

outlines/models/exllamav2.py
isamu-isozaki: Good point fixing now.

outlines/models/exllamav2.py
isamu-isozaki: @lapp0 oh do you mean instead of having the hf_tokenizer like below create the hf_tokenizer from the exllamav2's tokenizer using a function?

outlines/models/exllamav2.py
isamu-isozaki: I mainly did this because in exllamav2 the default value for encode_special_tokens was False. But I can't think of much reason for it needing to be False. Let me try changing to True and then running the tests

outlines/models/exllamav2.py
lapp0: Exactly. This way downstream users won't have to define their own `patch_tokenizer()` to use our logits processors as I do currently in the example here:

https://github.com/lapp0/exllamav2/pull/1/files#diff-26f303de07c10aad998e33d3df52581643673a598162cc4b35ef051f52d7c60bR68-R81

outlines/models/exllamav2.py
lapp0: Thanks for the explanation!

isamu-isozaki: @lapp0 Added comment! Let me know if it looks good

outlines/models/exllamav2.py
isamu-isozaki: @lapp0 Added comment! Let me know if it looks good

outlines/models/exllamav2.py
isamu-isozaki: Done!

isamu-isozaki:
@lapp0 Thanks for the review. I did all the changes I think and all my tests pass locally(including pre-commit)

lapp0:
Great job @isamu-isozaki !

I've opened the EXL2 PR for logits processors

https://github.com/turboderp/exllamav2/pull/634

isamu-isozaki:
@lapp0 awesome!

lapp0:
@isamu-isozaki I'm not sure whether the ExLlamaV2 PR will be merged soon, it's been a week without comment. Perhaps to get this out the door would you be able to update Outlines ExLlamaV2 documentation to make clear:
- ExLlamaV2 doesn't have logits processor support yet.
- There is a **third party** fork which supports logits processors and is compatible with outlines
  - The install command is `pip install git+https://github.com/lapp0/exllamav2@sampler-logits-processor`

Could you also let me know what build issues you experienced? I didn't run into any but I'd like to ensure the install-from-git command doesn't result in additional confusion.

isamu-isozaki:
@lapp0 Sorry was away for a bit and sounds good! And also thanks for making that pr! One question, I was following the discussion in exllamav2 but even after this will exllamav2's constrained gen be the slowest(because of double creation of the logits mask+it relies on this in the cpu)?

lapp0:
@isamu-isozaki No problem, thanks for your great work!

I expect constrained generation with `models.exllamav2` to have similar overhead to other models. 

The mask is currently applied on CPU, but it can be applied on GPU, prior to CPU offloading. I've profiled this mode and it increased token throughput in "normal" mode to 145 tok/s. This isn't pushed yet though.

I'm not sure what you mean by double creation of logits mask. The mask should only be applied once per token generated.

Please let me know if you have any other questions.

isamu-isozaki:
@lapp0 Ah, I misread and that sounds good! To do mask creation in gpu, then can this be done fully on the outlines side(after your pr)?

lapp0:
Two changes are necessary:
- Update Outlines so Guides optionally keep cached masks on GPU
- Update my ExLlamaV2 branch so it applies logits processors before its offloaded to cpu (offloading to CPU is immediate right now: https://github.com/lapp0/exllamav2/blob/10a8842b251685b4fbd3d263070b08ef07203903/exllamav2/generator/base.py#L265)

isamu-isozaki:
@lapp0 Thanks! Yeah, I just have a person hosting an outlines server with this code and he noticed that his production server with very weak CPUs was very slow which was our observation haha. I think what's best may depend on the type of hardware in production

lapp0:
@isamu-isozaki could you please open a discussion with further details about their performance regression? I'll take a look and see what the root cause is.

Issue #1011, opened 2024-06-30
Performance loss with guided_json activated in vLLM


robcaulk:
### Describe the issue as clearly as possible:

We've noticed that vLLM is showing a lower throughput when guided_json is used, versus when it is not used. Is this expected? In the following script, you need a vLLM server running, but then you can comment out the `model_kwargs` line of the ChatOpenAI object to see the difference in throughput. 

We measure somewhere around 50% slower when guided_json is activated.

Latest vLLM (v0.5.0-post1)
Python3.11

### Steps/code to reproduce the bug:

```python
from langchain.prompts import (
    HumanMessagePromptTemplate,
    ChatPromptTemplate,
)
from langchain_openai import ChatOpenAI
import time
from pydantic import BaseModel

class GRAPHMODEL(BaseModel):
    nodes: list[dict[str, str]]
    edges: list[dict[str, str]]


GRAPHKWARGS = {
    "extra_body": {
        "guided_json": GRAPHMODEL.model_json_schema(),
    }
}

BUILD_RELATIONSHIPS = ChatPromptTemplate.from_messages([

    HumanMessagePromptTemplate.from_template("""A chat between a curious user and an artificial intelligence Assistant. The Assistant gives helpful, detailed, and polite answers to the user's questions.

The User provides a news story in the format:

-------News story begin-------
Title: [Title of the article]
Story: [Summary of the article]    
-------News story end-------

The Assistant follows the following steps before replying to the User:          

1. **identify the most important entities** The Assistant identifies the most important entities in the news story. These entities are listed in the JSON output under the key "nodes", they follow the structure of a list of dictionaries where each dict is:

"nodes":[
    {{"id": <entity N>, "type": <type>, "detailed_type": <detailed type>}},
]

where "type": <type> is a broad categorization of the entity. "detailed type": <detailed_type>  is a very descriptive categorization of the entity.

2. **determine relationships** The Assistant uses the news story between -------News story begin------- and -------News story end------- to determine the relationships between the entities identified in the "nodes" list. These relationships are called "edges" and they follow the structure of:

"edges":[
    {{"from": <entity 1>, "to": <entity 2>, "label": <relationship>}},
]

The <entity N> must correspond to the "id" of an entity in the "nodes" list.

The Assistant responds to the User in JSON only, according to the following JSON schema:
                  
[...]

Issue #920, opened 2024-05-26
Performance updates for outlines and vllm integration


lynkz-matt-psaltis:
### What behavior of the library made you think about the improvement?

I'm working on ways to increase the throughput for batched requests for vllm when using outlines. Currently, batched operations throughput is highly constrained due to the recreation of tensors from the allowed_tokens list on every iteration. This is found to be due to the following factors:

1. Conversion of the larger python lists to tensors is slow and can lead to cpu contention, using numpy arrays as an intermediary step does alleviate some of this slowdown.
2. FSM states_to_token_maps are extracted as a collection of keys and then converted into a list. See: https://github.com/outlines-dev/outlines/blob/main/outlines/fsm/guide.py#L176 This in turn is converted to a tensor on every pass.

I'm looking for more efficient ways to be able to cache the generated tensors to avoid the intermediate steps on every iteration.
Currently the calculated state value for the state machine does not equate to a unique set of tokens that can be cached.
Additionally, because the keys are further wrapped in a list, its impossible to use reference equality on the returned data to determine if the returned allowed_token_ids in fact represents a previously cached result.

I'm currently resorting to using hash over these collections to determine this and hence this improvement consideration.

### How would you like it to behave?

Can we make it easier to determine when the FSM is returning a previously generated result maybe as part of the new Instruction concept?

Should we store this information directly to a provided tensor or provide ways to override behaviours to stream tokens to the desired destination in a more memory, cpu & gpu efficient fashion?

lynkz-matt-psaltis:
Some further context here: https://github.com/vllm-project/vllm/issues/3567

PR #1013, opened 2024-07-01
`RegexFSM`: Cache Legal-Token Mask as `torch.Tensor` to Improve Performance

lapp0:
Fixes https://github.com/outlines-dev/outlines/issues/1011

Fixes https://github.com/outlines-dev/outlines/issues/920

## Problem

Runtime performance for structured generation was weak because the legal next token mask is `List[int]` in main. This is inefficient because we have to convert a list of potentially tens of thousands of integers to `torch.tensor`.

## Solution

Cache legal next token ids as a `torch.tensor`. Don't construct the mask from `List[int]`, construct it from `torch.Tensor`.

With a vocabulary of ~50,000 tokens, worst case performance decreases 50-fold.

```
| Change   | Before [833f68f8]    | After [d9baa1ba]    |   Ratio | Benchmark (Parameter)                                                                            |
|----------|----------------------|---------------------|---------|--------------------------------------------------------------------------------------------------|
| -        | 56.0±0.3ms           | 1.05±0.02ms         |    0.02 | bench_processors.LogitsProcessorStructuredBenchmark.time_structured_generation('numpy', '[^Z]*') |
| -        | 58.1±0.6ms           | 1.05±0.01ms         |    0.02 | bench_processors.LogitsProcessorStructuredBenchmark.time_structured_generation('torch', '[^Z]*') |
```

### DIFF
--- a/benchmarks/bench_processors.py
+++ b/benchmarks/bench_processors.py
@@ -1,8 +1,13 @@
-import mlx.core as mx
 import numpy as np
 import torch
 
-from outlines.processors import OutlinesLogitsProcessor
+import outlines.models as models
+from outlines.processors import OutlinesLogitsProcessor, RegexLogitsProcessor
+
+try:
+    import mlx.core as mx
+except ImportError:
+    pass
 
 
 def is_mlx_lm_allowed():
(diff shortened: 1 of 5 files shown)


Qubitium:
@lapp0 Do you expect even more gains for models with >200k tokenizers as it has become more common?

lapp0:
@Qubitium yes, the gains are proportional to the size of the tokenizers vocabulary.

rlouf:
Great catch!

Issue #987, opened 2024-06-19
Format `prompts` Using Chat Templates in `SequenceGeneratorAdapter`


lapp0:
Related: https://github.com/outlines-dev/outlines/issues/756

### What behavior of the library made you think about the improvement?

Currently when using `outlines.generate`, chat templates aren't applied by default. It's awkward and unintuitive to structure your prompts as chat templates. For example, a well structured input for a llama-3 model might look like

```
generator = outlines.generate.json(...)

my_prompt = """<|begin_of_text|><|start_header_id|>user<|end_header_id|>\n\nProvide me JSON Data<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"""
generator(my_prompt)
```

I'd prefer 
```
generator("Provide me JSON Data")
```

### Why We Should Apply Chat Templates by Default

Without the application of chat templates, the model emulates the continuation of a monologue. Where-as chat template format generally follows a query-response structure.

##### No Chat Template

```
>>> output = model.generate(**tokenizer("What is 1 + 1?", return_tensors="pt"), max_length=32)                                                                                                                    
>>> tokenizer.decode(output[0])
"<s> What is 1 + 1?\n\nThis question has been asked by many people, but I don't understand the answer.\n\nCould"
```

```
>>> output = model.generate(**tokenizer("Give me a random color:", return_tensors="pt"), max_length=32)
>>> tokenizer.decode(output[0])
'<s> Give me a random color:\n\n- Response: A random color can be represented in hexadecimal format as #RRGGBB,'
```

##### With Chat Template

```
output = model.generate(**tokenizer('<s><|user|> What is 1 + 1?<|end|><|assistant|>', return_tensors="pt"), max_length=32)
tokenizer.decode(output[0])
'<s><s><|user|> What is 1 + 1?<|end|><|assistant|> 1 + 1 equals 2. This is a basic arithmetic addition problem. When you'
```

```
>>> output = model.generate(**tokenizer('<s><|user|> Give me a random color:<|end|><|assistant|>', return_tensors="pt"), max_length=32)
>>> tokenizer.decode(output[0])
"<s><s><|user|> Give me a random color:<|end|><|assistant|> The random color I'll describe for you is a vibrant shade of teal, with"
```

### How would you like it to behave?

By default `generator(prompt)` applies the chat template.

Current behavior should remain available via `generator(prompt, raw=True)`

[...]

PR #1019, opened 2024-07-05
Auto-apply chat template in `SequenceGenerator` and `SequenceGeneratorAdapter`, if available

leloykun:
This PR auto-applies chat templates by default when using instruct/chat models. Doesn't support LlamaCPP for now tho.

---

### Why?

Instruct/Chat models tend to be annoyingly template dependent (i.e. they perform worse if the prompts don't follow the chat template used in the finetuning step). And the more and longer they are finetuned, the worse the problem gets. Hence this PR.

Also see this issue https://github.com/outlines-dev/outlines/issues/987 raised by @lapp0 

---

### Interface changes

This PR has minimal impact on the interface. It just changes the default behavior of the generation step.

However, this feature can be disabled either on the creation of the generator:
```python
generator = outlines.generate.choice(model, ["Positive", "Negative"], apply_chat_template=False)
```
Or when calling the generator
```python
answer = generator(prompt, apply_chat_template=False)
```

### DIFF
--- a/outlines/generate/api.py
+++ b/outlines/generate/api.py
@@ -1,8 +1,10 @@
 import datetime
+import warnings
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
 
 from outlines.generate.generator import sequence_generator
+from outlines.models.transformers import TransformerTokenizer
 from outlines.samplers import BeamSearchSampler, GreedySampler, MultinomialSampler
 
 if TYPE_CHECKING:
@@ -132,6 +134,7 @@ class SequenceGenerator:
         max_tokens: Optional[int] = None,
         stop_at: Optional[Union[str, List[str]]] = None,
         rng: Optional["torch.Generator"] = None,
+        apply_chat_template: bool = True,
     ) -> Union[FormattedOutput, List[FormattedOutput], List[List[FormattedOutput]]]:
         """Generate the full text sequence.
 
@@ -153,6 +156,9 @@ class SequenceGenerator:
         rng
             The random number generator. Defaults to a non-seeded `torch.Generator`
             instance.
+        apply_chat_template
+            Whether to apply the chat template to the prompts. Defaults to `True`.
+            Only applies to `TransformerTokenizer` for now.
 
         Returns
         -------
@@ -162,6 +168,20 @@ class SequenceGenerator:
 
         if isinstance(prompts, str):
             prompts = [prompts]
+
+        if apply_chat_template:
+            if not isinstance(self.tokenizer, TransformerTokenizer):
+                warnings.warn(
+                    "Chat template is only supported for `Transformer` models for now. The raw prompts will be used instead."
+                )
+            elif getattr(self.tokenizer.tokenizer, "chat_template", None) is None:
+                warnings.warn(
+                    "The model does not have chat template support. The raw prompts will be used instead. To turn this warning off, either explicitly set the `apply_chat_template` argument to 'False' or assign a value to `tokenizer.tokenizer.chat_template`."
+                )
+            else:
+                prompts = [
+                    self.tokenizer.apply_chat_template(prompt) for prompt in prompts
+                ]
 
         if isinstance(stop_at, str):
             stop_at = [stop_at]
--- a/outlines/generate/api.py
+++ b/outlines/generate/api.py
@@ -4,7 +4,6 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Iterator, List, Optional, Union
 
 from outlines.generate.generator import sequence_generator
-from outlines.models.transformers import TransformerTokenizer
 from outlines.samplers import BeamSearchSampler, GreedySampler, MultinomialSampler
 
 if TYPE_CHECKING:
@@ -571,6 +570,8 @@ class SequenceGeneratorAdapter:
 
 
 def apply_chat_template_util(model, prompts: Union[str, List[str]]) -> List[str]:
+    from outlines.models.transformers import TransformerTokenizer
+
     if isinstance(prompts, str):
         prompts = [prompts]
     if not isinstance(model.tokenizer, TransformerTokenizer):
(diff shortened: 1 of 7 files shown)


outlines/generate/api.py
```diff
@@ -529,3 +567,22 @@ def stream(
             self.sampling_params,
             **model_specific_params,
         )
+
+
+def apply_chat_template_util(model, prompts: Union[str, List[str]]) -> List[str]:
+    from outlines.models.transformers import TransformerTokenizer
+
+    if isinstance(prompts, str):
+        prompts = [prompts]
```
lapp0, line 576: I think the signature should be `List[str]` -> `List[str]` and raise an error if a list isn't passed. In `transformers.py`, this function is called after the prompts are normalized to 2D anyways.

outlines/generate/api.py
```diff
@@ -529,3 +567,22 @@ def stream(
...
+    if not isinstance(model.tokenizer, TransformerTokenizer):
+        warnings.warn(
+            "Chat template is only supported for `Transformer` models for now. The raw prompts will be used instead."
+        )
+        return prompts
+    tokenizer: "TransformerTokenizer" = model.tokenizer
+    if getattr(tokenizer.tokenizer, "chat_template", None) is None:
```
lapp0, line 583: Should we use `default_chat_template` instead?

https://huggingface.co/docs/transformers/v4.34.0/chat_templating#how-do-chat-templates-work

outlines/generate/api.py
```diff
@@ -516,9 +549,14 @@ def stream(
         max_tokens: Optional[int] = None,
         stop_at: Optional[Union[str, List[str]]] = None,
         seed: Optional[int] = None,
+        apply_chat_template: Optional[bool] = None,
         **model_specific_params,
     ):
         """Return a text generator from a prompt or a list of prompts."""
+        if apply_chat_template is None:
```
lapp0, line 556: To make this pythonic, we should have one obvious way of applying a chat template. IMO the argument should only be accepted in the constructor.

outlines/models/transformers.py
```diff
@@ -100,6 +116,24 @@ def convert_token_to_string(self, token: str) -> str:
...
+        Template for one-turn chat.
+        """
+        return (
+            [self.SYSTEM_MESSAGE.copy(), self.USER_MESSAGE_TEMPLATE(prompt)]
+            if self.SYSTEM_MESSAGE is not None
+            else [self.USER_MESSAGE_TEMPLATE(prompt)]
+        )
+
+    def apply_chat_template(self, prompt: str) -> str:
+        messages = self.get_messages(prompt)
```
lapp0, line 130: We should probably have this be part of `class Tokenizer` and raise `NotImplementedError` by default.

lapp0 (changes requested):
Thanks so much for implementing this!

There's some things I'd like to see before this is ready for merge, could you please address the following
- Test cases which ensure application of the chat template works properly for a standard case, for a missing chat template
- Consider that some users of Outlines are relying on the current behavior and make `apply_chat_template=False` the default. Perhaps we can warn the user as well if they're not setting the parameter.
- Add `apply_chat_template` explanation to the docs
- Use `apply_chat_template=True` in all `models.transformers` generator examples

rlouf:
I don't think this is the correct design; we want to avoid kwargs as much as we possibly can. Here I would simply add a `apply_chat_template` to the model instance.

leloykun:
@lapp0 @rlouf

How do you guys think should the interface look like?

---

Here, I mirrored Huggingface's Pipeline interface where we can specify configs/args in the constructor and (optionally) override them in the model call. I like it cuz it's more flexible. But yah, it does make things a bit more complicated and less pythonic.

I did a quick look at other libraries, and it seems that they either (1) don't auto-apply the chat templates at all or (2) have a separate code path for base & instruct/chat models (e.g. TGI & MixEval). I think there are two reasons why:

1) *It's hard to know which models are base models and which are instruct/chat models*. I thought checking whether `chat_template` is `None` or not would suffice. But some chat models apparently just leave them out (especially the older ones & third-party finetunes). Additionally, `transformers`' `PreTrainedTokenizerBase` base class has a `default_chat_template` property--so, if I'm not mistaken, we can run `tokenizer.apply_chat_template` on _all_ tokenizers without erroring out. And

2) *Some models don't support system prompts and it's hard to know which ones do and which ones don't*.

So yah, for now, we need a way to somehow let the library know that whether we're dealing with a base model or an instruct/chat model. Worst case is we might also need to ask the user to specify the system prompt. But if we're gonna force them to go all that trouble anyway, we might as well not do this by default.

I think a good compromise is to (1) not apply the chat templates by default but (2) warn the user if the chat template is specified in the tokenizer config but is not being used.

rlouf:
```python
answer = generator(
   model.apply_chat_template(prompt)
)
```

alonsosilvaallende:
I like to give this pull request that adds the possibility to add the `apply_chat_template`, but I wonder if it's a good idea to make it the default behavior. I have had very bad experiences with `apply_chat_template` where it add spaces or remove spaces when it shouldn't or even worse in function calling cases where it completely ignores the functions given without even raising errors (see [this example](https://gist.github.com/alonsosilvaallende/b252f6db38039158101b719bacc1bf25)). Many people might complain about something which is outside the control of Outlines.

rlouf:
Indeed, we are not going to make it the default behavior. Users should be able to inspect the modified prompt before they pass it to the generator.

rlouf:
@leloykun are you still working on this?

leloykun:
@rlouf yup! I just deprioritized this in favor of other stuff; I'll get back to this soon

btw, thanks for the feedback, @alonsosilvaallende!

Bit0r:
So can this feature still be added? A chat template is really needed in many applications.

lapp0:
> So can this feature still be added? A chat template is really needed in many applications.

Yes, if you'd like to introduce it in a PR, I'll review and support its inclusion provided it follows the behavior discussed in this thread! :)

RobinPicard:
Closing the PR as Outlines now has a `Chat` model input object that serves this purpose.

Issue #923, opened 2024-05-27
Translate Pydantic models into a regular expression that accept the corresponding YAML


rlouf:

lapp0:
I love the idea, yaml uses fewer syntactic tokens and allows language models to generate without needing to keep track of as much "nesting" / context.

Here's what I'm thinking for a strategy, would love to hear your thoughts:

We should refactor `fsm/json_schema.py` so it uses a class-based approach with handler methods for each `type`. Then we can subclass to implement the different behavior in yaml.

```
class JSONSchemaRegexGenerator:
    def __init__(self):
        self.handlers = {
            "string": self.handle_string,
            "object": self.handle_integer,
            ...
        }

    @classmethod
    def get_pattern(cls, schema):
        return cls().handle_node(schema)

    def get_pattern(self, node):
        handler = self.handlers.get(node["type"], self.handle_default)
        return handler(node)

    def handle_string(self, node):
        return STRING

    def handle_array(self, node):
        ...
        return rf"\[{whitespace_pattern}({'|'.join(regexes)})(,{whitespace_pattern}({'|'.join(regexes)})){num_repeats}){allow_empty}{whitespace_pattern}\]"


class YAMLSchemaRegexGenerator(JSONSchemaRegexGenerator):
    def handle_array(self, node):
        """handle format for yaml arrays:
            - string1
            - string2
        """
        ...     
``

This would make the code more readable, extensible, reduce technical debt, and make it so we don't have to have conditional handling for a passed `is_yaml` for many rules within `to_regex()`

rlouf:
I can get on board with this. To follow `ast.NodeVisitor`'s naming scheme we could name the handlers `visit_X`. I think we should first implement a first version of the converter to YAML with only a few primitives before refactoring.

PR #1022, opened 2024-07-06
Translate JSON Schema to YAML regex

patricebechard:
This is a tentative implementation of a regex generator for arbitrary YAML given a JSON Schema. This PR relates to #923 .

There are still some issues:
* the NUMBER regex can be revisited. I have taken it from the official pyyaml repo [link](https://github.com/yaml/pyyaml/blob/main/lib/yaml/resolver.py). The INTEGER regex can also be changed.
* I don't address the refactoring suggested in #923. This results in some code deduplication between both files.
* The regex generator is not very strict regarding the indentation levels, which might make it impractical in some use cases.

@rlouf

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,4 +30,4 @@ repos:
     - id: mypy
       args: [--allow-redefinition]
       exclude: ^examples/
-      additional_dependencies: [types-tqdm]
+      additional_dependencies: [types-tqdm, types-PyYAML]
--- a/outlines/fsm/yaml_schema.py
+++ b/outlines/fsm/yaml_schema.py
@@ -1,0 +1,574 @@
+import inspect
+import json
+import re
+import warnings
+from typing import Callable, Optional, Tuple
+
+import yaml
+from jsonschema.protocols import Validator
+from pydantic import create_model
+from referencing import Registry, Resource
+from referencing._core import Resolver
+from referencing.jsonschema import DRAFT202012
+
+# taken from https://github.com/yaml/pyyaml/blob/main/lib/yaml/resolver.py
+TRUE = r"(?:yes|Yes|YES|true|True|TRUE|on|On|ON)"
+FALSE = r"(?:no|No|NO|false|False|FALSE|off|Off|OFF)"
+BOOLEAN = (
+    r"(?:yes|Yes|YES|no|No|NO|true|True|TRUE|false|False|FALSE|on|On|ON|off|Off|OFF)"
+)
+INTEGER = r"(?:[-+]?0b[0-1_]+|[-+]?0[0-7_]+|[-+]?(?:0|[1-9][0-9_]*)|[-+]?0x[0-9a-fA-F_]+|[-+]?[1-9][0-9_]*(?::[0-5]?[0-9])+)"
+NUMBER = (
+    r"(?:[-+]?(?:[0-9][0-9_]*)\.[0-9_]*(?:[eE][-+][0-9]+)?"
+    r"|\.[0-9][0-9_]*(?:[eE][-+][0-9]+)?"
+    r"|[-+]?[0-9][0-9_]*(?::[0-5]?[0-9])+\.[0-9_]*"
+    r"|[-+]?\.(?:inf|Inf|INF)"
+    r"|\.(?:nan|NaN|NAN))"
+)
+NULL = r"(?: ~|null|Null|NULL| )"
+TIMESTAMP = (
+    r"^(?:[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]"
+    r"|[0-9][0-9][0-9][0-9] -[0-9][0-9]? -[0-9][0-9]?"
+    r"(?:[Tt]|[ \t]+)[0-9][0-9]?"
+    r":[0-9][0-9] :[0-9][0-9] (?:\.[0-9]*)?"
+    r"(?:[ \t]*(?:Z|[-+][0-9][0-9]?(?::[0-9][0-9])?))?)$"
+)
+
+# allow `\"`, `\\`, or any character which isn't a control sequence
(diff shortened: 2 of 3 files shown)


rlouf:
Thank you for the PR! It looks like many tests related to these changes are failing…

lapp0 (changes requested):
Great work on this! Thanks so much for implementing.

Could we harden this a bit with some test cases? I've written some test cases that could help
- [Modified `test_json_schema.py` which uses the existing test cases and to validate an equivalent yaml file](https://github.com/lapp0/outlines/blob/yaml-tests/tests/fsm/test_json_schema.py)

patricebechard:
I have already copied some tests from `test_json_schema.py` and updated them for the YAML use case, but I am open to reusing the code as you have done it (or simply add more tests from the test_json file and edit them).

Given that you're refactoring some of the code on your branch, what is the easiest way to go forward to minimize the amount of duplication / deprecation of code?

lapp0:
> I have already copied some tests from `test_json_schema.py` and updated them for the YAML use case, but I am open to reusing the code as you have done it (or simply add more tests from the test_json file and edit them).
> 
> Given that you're refactoring some of the code on your branch, what is the easiest way to go forward to minimize the amount of duplication / deprecation of code?
 
I've refactored `json_schema.py`, but unless you're interested in incorporating those changes, we can just focus on `test_json_schema.py` for now. Simply replacing `test_json_schema.py` on your branch with my branches version and ensuring it works is sufficient. The new module simply ensures the tested behavior in `json_schema.py` is matched by `yaml_schema.py`.


Please let me know if you have any other questions.

lapp0:
@patricebechard is there anything I can do to help with this?

patricebechard:
sorry, was quite busy lately, but I can work on it this week, will let you know if I need help with anything

lapp0:
No worries at all, thanks for your continued work!

lapp0:
>I am also making sure that we support both quoted and unquoted strings for YAML. Since one of the main advantages of using YAML for guided generation is that there are less tokens, if we add a double quote every time there is a string, we end up with a generation which is almost as big as the one obtained in JSON, so transitioning to YAML would not make sense.

We may want to smoke test qualitative generation performance when using YAML. Out of scope for this PR, but disallowing quotes may, in some cases, confuse the model. 

>the implementation differs from what @lapp0 has on his branch, which would mean we would have to do some refactoring at some point.

This is fine for now.

Thanks for getting this working! Please let me know when this PR is ready for review.

patricebechard:
Not exactly sure how the coverage is computed here. It says the coverage for `yaml_schema.py` is ~1% although it should be higher. Any idea how to remedy this? This did not happen previously from what I understand.

lapp0:
Change in coverage may be related to https://github.com/outlines-dev/outlines/pull/1089

I've created a separate issue to address this problem https://github.com/outlines-dev/outlines/issues/1105

outlines/fsm/json_schema.py
```diff
@@ -28,10 +28,16 @@
...
+DATE_TIME = (
+    r'("(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]{3})?(Z)?"'
+    r"|"
+    r"'(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]{3})?(Z)?'"
+    r"|"
+    r"(-?(?:[1-9][0-9]*)?[0-9]{4})-(1[0-2]|0[1-9])-(3[01]|0[1-9]|[12][0-9])T(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\.[0-9]{3})?(Z)?)"
+)
+DATE = r'("(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1])"|\'(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1])\'|(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1]))'
+TIME = r'("(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?"|\'(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?\'|(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?)'
+UUID = r'("[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"|\'[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\'|[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})'
```
lapp0, line 40: Shouldn't this only be in `yaml_schema.py`? It allows for single-quote and no-quote datetime, date, time, and UUID.

lapp0 (changes requested):
Requesting `json_schema.py` types fix.

Good work integrating the additional yaml tests. I'd like to smoke test a bit before we merge as well to ensure there aren't some edge cases we're missing.

RobinPicard:
Hi @patricebechard! We recently released the v1 of Outlines. Outlines has changed a bit since you last worked on this PR. There are in particular more operations that are now handled by [outlines-core](https://github.com/dottxt-ai/outlines-core). What you want to do in this PR is very similar to the function `build_regex_from_schema` in `outlines_core`. If you're still interested by this topic, it could be interesting for you to go see how it works.

PR #1030, opened 2024-07-11
Fix broken link in README.md regarding Serving with vLLM

shinohara-rin:
updated the link to https://outlines-dev.github.io/outlines/reference/serve/vllm/


rlouf:
Thank you!

PR #1034, opened 2024-07-12
Update file contributors style for docs

gtsiolis:
Minor style change for file contributors for accessibility.

| BEFORE | AFTER |
|--------|--------|
| Frame 1235 | Frame 1234 |
| Frame 1236 | Frame 1237 |


rlouf:
Thank you! Are the contributors supposed to be centered? I still see them on the left (phone).

PR #1039, opened 2024-07-14
Adds support for custom regex parsers (for multimodal structured generation)

leloykun:
As [discussed in our Discord server](https://discord.com/channels/1182316225284554793/1182317446225481788/1261998326077984802)

This PR adds support for custom regex parsers. This doesn't change the behavior of Outlines by default. But this allows us to write custom `Guide` classes that uses custom regex parsers for e.g. multimodal generation.

Also improves documentation


rlouf:
Thank you!

PR #1042, opened 2024-07-15
Add QA with Citations example to the Cookbook

alonsosilvaallende:
As requested by @rlouf this PR adds a question answering with citations example to the Cookbook using llama-cpp-python.


rlouf:
Awesome, thank you! Two things:

1- Can you add a link to this file here: https://github.com/outlines-dev/outlines/blob/main/mkdocs.yml so it shows in the docs
2- Could you run `pre-commit` locally?

alonsosilvaallende:
Thank you @rlouf 
I have done both requests.

rlouf:
Thank you for adding the example!

PR #1051, opened 2024-07-18
Correct link and add llama-cpp-python installation instructions

alonsosilvaallende:
- Correct link for llama-cpp-python
- Add installation instructions for llama-cpp-python
- Correct first question-answer


rlouf (approved):

rlouf:
Thank you for following up!

PR #1057, opened 2024-07-20
Modal documentation: fix deprecated memory= parameter

Perdu:
memory= parameter is deprecated in favor of size=
See https://modal.com/docs/reference/changelog#062174-2024-05-17

Current doc example produces the following error:
```
/path/test_modal.py:56: DeprecationError: 2024-05-16: The `memory` parameter is deprecated. Use the `size='80GB'` parameter instead.
  @app.function(image=outlines_image, gpu=gpu.A100(memory=80))
```


rlouf:
Thank you for updating the example!

PR #1061, opened 2024-07-22
Remove deprecated `outlines.integrations`

### LINKED ISSUE #1054: Clean Up Dead `outlines.integrations` Code
### What behavior of the library made you think about the improvement?

With all models other than `exllamav2` having a single integration implementation via `outlines.generate.api.SequenceGeneratorAdapter` and `outlines.processors`, we don't need `outlines.integrations`.

- `outlines.integrations.transformers`  has `RegexPrefixAllowedTokens` and `JSONPrefixAllowedTokens` which are now handled by `outlines.processors`
- `outlines.integrations.llamacpp` has logits processors which are redundant with `outlines.processors`
- `outlines.integrations.vllm` has logits processors which are redundant with `outlines.processors`
- `outlines.integrations.utils` has `adapt_tokenizer`, useful for FSM construction, and `convert_json_schema_to_str`, used for the three model integrations listed above

After the [vLLM integration](https://github.com/outlines-dev/outlines/pull/1053), we no longer need `outlines.integrations`

### How would you like it to behave?

Get rid of `outlines/integrations/` and `tests/generate/test_integration_*.py`

- Remove `outlines.integrations.transformers`, `outlines.integrations.llamacpp`, and `outlines.integrations.vllm`
- Remove `outlines.integrations.utils`, but move `def adapt_tokenizer(...)` to `outlines.models.tokenizer`

rlouf:
Fixes #1054


outlines/models/vllm.py
```diff
@@ -185,3 +186,41 @@ def vllm(model_name: str, **vllm_model_params):
     model = LLM(model_name, **vllm_model_params)
 
     return VLLM(model)
+
+
+def adapt_tokenizer(tokenizer: PreTrainedTokenizerBase) -> PreTrainedTokenizerBase:
```
rlouf, line 191: Since vLLM uses `transformers` tokenizer this logic is repeated over from `transformers.py`. We need to re-use what is implemented in the latter.

outlines/serve/serve.py
```diff
@@ -35,12 +35,14 @@
 from vllm.sampling_params import SamplingParams
 from vllm.utils import random_uuid
 
-from outlines.integrations.vllm import JSONLogitsProcessor, RegexLogitsProcessor
+from outlines.models.vllm import adapt_tokenizer
+from outlines.processors.structured import JSONLogitsProcessor, RegexLogitsProcessor
```
lapp0, line 39: nit: `from outlines.processors import JSONLogitsProcessor, RegexLogitsProcessor`

outlines/serve/serve.py
```diff
@@ -69,9 +71,9 @@ async def generate(request: Request) -> Response:
     json_schema = request_dict.pop("schema", None)
     regex_string = request_dict.pop("regex", None)
     if json_schema is not None:
-        logits_processors = [JSONLogitsProcessor(json_schema, engine.engine)]
+        logits_processors = [JSONLogitsProcessor(json_schema, tokenizer)]
     elif regex_string is not None:
-        logits_processors = [RegexLogitsProcessor(regex_string, engine.engine)]
+        logits_processors = [RegexLogitsProcessor(regex_string, tokenizer)]
```
lapp0, line 76: Has this been smoke tested?

Also, we should notify downstream of the interface change. Ensure they're pinned to our prior release or https://github.com/vllm-project/vllm/pull/6715 is merged before we publish a new release.

tests/fsm/test_regex.py
```diff
@@ -18,8 +18,8 @@
     reduced_vocabulary,
     walk_fsm,
 )
-from outlines.integrations.utils import adapt_tokenizer
 from outlines.models.transformers import TransformerTokenizer
+from outlines.models.vllm import adapt_tokenizer
```
lapp0, line 22: `test_reduced_vocabulary_with_rare_tokens` should just use `TransformerTokenizer` rather than calling `adapt_tokenizer`

rlouf:
@lapp0 I addressed your comments. Please review one last time before we merge.

PR #1068, opened 2024-07-27
Fix TransformerTokenizer pad_token_id reset

ispobock:
For some LLM models (such as [chatglm3-6b](https://huggingface.co/THUDM/chatglm3-6b)), the `pad_token_id` is defined as 0 and immutable. We only need to reset for undefined cases.


ispobock:
@shawnz Could you help review?

lapp0 (approved):
Great find, thanks for fixing!

PR #1073, opened 2024-07-29
Add `apply_chat_template` to Tokenizer

leloykun:
This PR adds `apply_chat_template` to Outline's `Tokenizer`. This replaces https://github.com/outlines-dev/outlines/pull/1019

###

Unlike Huggingface's tokenizers, we can directly pass the text prompt to `tokenizer.apply_chat_template`. This will first convert the prompt into conversation format with a predefined system message, then applies the chat template to the result.

```python
answer = generator(
    tokenizer.apply_chat_template(prompt)
)
```

To use a different system message, either (1) modify `tokenizer.SYSTEM_MESSAGE["content"]` or (2) build the conversation yourself then pass it to the `apply_chat_template` here:

```
conversation = [
    {
        "role": "system",
        "content": "You are a helpful assistant. Please do not say swear words.",
    },
    {
        "role": "user",
        "content": [insert prompt here],
    },
]
```

```python
answer = generator(
    tokenizer.apply_chat_template(conversation)
)
```

## Interface changes:

Default behavior is kept as is.

-----------

@rlouf @lapp0 imo adding it to the tokenizer instead of the model makes more sense.

* in chatml format

### DIFF
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -52,14 +52,6 @@ class LlamaCppTokenizer(Tokenizer):
 
         self._hash = None
 
-        # set to None if no system message
-        self.SYSTEM_MESSAGE = {
-            "role": "system",
-            "content": "You are a helpful assistant.",
-        }
-        self.USER_MESSAGE_TEMPLATE = lambda x: {"role": "user", "content": x}
-        self.ASSISTANT_MESSAGE_TEMPLATE = lambda x: {"role": "assistant", "content": x}
-
     def decode(self, token_ids: List[int]) -> List[str]:
         decoded_bytes = self.tokenizer.detokenize(token_ids)
         return [decoded_bytes.decode("utf-8", errors="ignore")]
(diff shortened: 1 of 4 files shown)


outlines/models/llamacpp.py
```diff
@@ -83,6 +83,12 @@ def convert_token_to_string(self, token: str) -> str:
         else:
             return token
 
+    def apply_chat_template(
+        self,
+        prompt_or_conversation: Union[str, List[Dict[str, str]]],
+    ) -> str:
+        raise NotImplementedError("Chat templates are not yet supported by llama-cpp")
```
lapp0, line 90: Could you either say "not yet supported by Outlines llama-cpp integration" or implement https://github.com/abetlen/llama-cpp-python/blob/master/README.md#chat-completion

outlines/models/transformers.py
```diff
@@ -81,6 +89,14 @@ def __init__(self, tokenizer: "PreTrainedTokenizer", **kwargs):
         self.vocabulary = self.tokenizer.get_vocab()
         self.is_llama = isinstance(self.tokenizer, get_llama_tokenizer_types())
 
+        # set to None if no system message
+        self.SYSTEM_MESSAGE = {
+            "role": "system",
+            "content": "You are a helpful assistant.",
+        }
+        self.USER_MESSAGE_TEMPLATE = lambda x: {"role": "user", "content": x}
+        self.ASSISTANT_MESSAGE_TEMPLATE = lambda x: {"role": "assistant", "content": x}
```
lapp0, line 98: this is never used

outlines/models/transformers.py
```diff
@@ -105,6 +121,28 @@ def convert_token_to_string(self, token: str) -> str:
...
 
+    def format_prompt_into_conversation(self, prompt: str) -> List[Dict[str, str]]:
+        """
+        Template for one-turn chat.
+        """
+        return (
+            [self.SYSTEM_MESSAGE.copy(), self.USER_MESSAGE_TEMPLATE(prompt)]
+            if self.SYSTEM_MESSAGE is not None
+            else [self.USER_MESSAGE_TEMPLATE(prompt)]
+        )
```
lapp0, line 132: Not all models have a system message (e.g. phi-3)

Is there a way built into `transformers` to handle this? (I'm not aware of one)

lapp0 (changes requested):
Good progress!

Could you please address my comments and
- add unit tests
- add integration tests which simply verify the method runs successfully and introduces a `bos` token in the result with different popular models tokenizers?

rlouf:
I think these changes should be applied to the [new `prompts` library](https://github.com/outlines-dev/prompts).

rlouf:
Closing in favor of `outlines-dev/prompts`

Issue #999, opened 2024-06-22
Version the documentation


rlouf:
The documentation points to the code in `main`, and not the latest release. Material Mkdocs [allows to version the docs](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/) and we should set this so as to not confuse our users.

PR #1111, opened 2024-08-22
Update modal example

cpfiffer:
I updated some of the text of the modal example to add more information on what's happening where, since Modal's general structure may be unfamiliar to new users.

List of changes:

- Removed version constraints on the image. I think this is a questionable choice so I'd love input -- in general, this should probably better be addressed by versioning the docs #999. There's an open PR #1059 here that can help with this too -- I'm happy to revert to pinned versions, or at least update to the current version of outlines (0.0.46).
- Added a brief note that the Mistral model is gated and requires you to request access on HuggingFace.
- Moved the environment variable setting from `import_model` to a `.env` call when the image is created. This is more idiomatic Modal code. It was previously addressed in #1058, but this should be more current.
- More comments and prep.

### DIFF
--- a/docs/cookbook/deploy-using-modal.md
+++ b/docs/cookbook/deploy-using-modal.md
@@ -4,36 +4,77 @@
 
 In this guide we will show you how you can use Modal to run programs written with Outlines on GPU in the cloud.
 
+## Requirements
+
+We recommend installing `modal` and `outlines` in a virtual environment. You can create one with:
+
+```bash
+python -m venv venv
+source venv/bin/activate
+```
+
+Then install the required packages:
+
+```bash
+pip install modal outlines
+```
+
 ## Build the image
 
-First we need to define our container image. We download the Mistral-7B-v0.1 model from HuggingFace as part of the definition of the image so it only needs to be done once (you need to provide an [access token](https://huggingface.co/settings/tokens))
+First we need to define our container image. We download the Mistral-7B-v0.1 model from HuggingFace as part of the definition of the image so it only needs to be done once (you need to provide an [access token](https://huggingface.co/settings/tokens)).
+
+Setting a token is best done by setting an environment variable `HF_TOKEN` with your token. If you do not wish to do this, we provide a commented-out line in the code to set the token directly in the code.
 
 ```python
 from modal import Image, App, gpu
+import os
 
+# This creates a modal App object. Here we set the name to "outlines-app".
+# There are other optional parameters like modal secrets, schedules, etc.
+# See the documentation here: https://modal.com/docs/reference/modal.App
 app = App(name="outlines-app")
 
+# Please set an environment variable HF_TOKEN with your Hugging Face API token.
+# The code below (the .env({...}) part) will copy the token from your local
+# environment to the container.
+# More info on Image here: https://modal.com/docs/reference/modal.Image
 outlines_image = Image.debian_slim(python_version="3.11").pip_install(
-    "outlines==0.0.37",
-    "transformers==4.38.2",
-    "datasets==2.18.0",
-    "accelerate==0.27.2",
-)
+    "outlines",
+    "transformers",
+    "datasets",
+    "accelerate",
+).env({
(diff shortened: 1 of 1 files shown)


docs/cookbook/deploy-using-modal.md
```diff
@@ -4,36 +4,79 @@
...
+
+    # To set the token directly in the code, uncomment the line below and replace
+    # 'YOUR_TOKEN' with the HuggingFace access token.
+    # 'HF_TOKEN':'YOUR_TOKEN'
+})
+```
+
+## Setting the container up
+
+When running longer Modal apps, it's recommended to download your language model when the container starts, rather than when the function is called. This is to reduce latency on the first time you send a request to the container.
```
rlouf, line 57: It's because whatever is done when initializing the container is cached for future runs?

rlouf (commented):
Great! Just one minor comment.

docs/cookbook/deploy-using-modal.md
cpfiffer: Updated!

cpfiffer:
Addressed the comments here -- any other comments?

charlesfrye:
> Removed version constraints on the image. I think this is a questionable choice so I'd love input -- in general, this should probably better be addressed by versioning the docs https://github.com/outlines-dev/outlines/issues/999. There's an open PR
https://github.com/outlines-dev/outlines/pull/1059 here that can help with this too -- I'm happy to revert to pinned versions, or at least update to the current version of outlines (0.0.46).

We generally pin to improve the stability of our examples. But you'll notice we don't pin Modal versions, because we expect to handle any issues that our own code changes cause. Makes sense for you to do the same!

> Added a brief note that the Mistral model is gated and requires you to request access on HuggingFace.

I have mostly moved over to using forks or finetunes of gated models that are not themselves gated, e.g. from [NousResearch](https://huggingface.co/NousResearch/Hermes-2-Pro-Mistral-7B), to avoid this bit of friction.

charlesfrye:
Re: non-gated versions of models.

I made that change to our version in [this commit](https://github.com/modal-labs/modal-examples/commit/1ff2dd432fdd27d50bb3dfe637a0721f5f2facb2). Note that the `mistral-community` organization on Hugging Face includes their CEO, `clem`, as a member, so while it's "unofficial" it's not jank.

cpfiffer:
Great, good idea! I added the ungated Mistral model, which also required a `sentencepiece` dependency. Added that too.

docs/cookbook/deploy-using-modal.md
```diff
@@ -4,36 +4,82 @@
...
+Then install the required packages:
+
+```bash
+pip install modal outlines
+```
+
 ## Build the image
 
-First we need to define our container image. We download the Mistral-7B-v0.1 model from HuggingFace as part of the definition of the image so it only needs to be done once (you need to provide an [access token](https://huggingface.co/settings/tokens))
+First we need to define our container image. We download the Mistral-7B-v0.1 model from HuggingFace as part of the definition of the image so it only needs to be done once (you need to provide an [access token](https://huggingface.co/settings/tokens)).
```
cpfiffer, line 24:
```suggestion
First we need to define our container image. If you need to access a gated model, you will need to provide an [access token](https://huggingface.co/settings/tokens). See the `.env` call below for how to provide a HuggingFace token.
```

docs/cookbook/deploy-using-modal.md
```diff
@@ -4,36 +4,82 @@
...
-    "datasets==2.18.0",
-    "accelerate==0.27.2",
-)
+    "outlines",
+    "transformers",
+    "datasets",
+    "accelerate",
+    "sentencepiece",
+).env({
+    'HF_TOKEN':os.environ['HF_TOKEN']
```
cpfiffer, line 52:
```suggestion
    # This will pull in your HF_TOKEN environment variable if you have one.
    'HF_TOKEN':os.environ['HF_TOKEN']
```

PR #1122, opened 2024-08-28
>> Add Support for Including the Prompt in the Pattern in SequenceGenerator

JulesGM:
Let's say you have a Json schema and you want to generate just a subset of keys from other keys. With the current SequenceGenerator, it's not possible to do that in an easy way.

We add the "include_prompt" argument, where the prompt is included in the pattern. 

We just unroll the FSMs with the prompt tokens. We also keep the prompt tokens inside of the output, as they are needed for the output `format_sequence` call in things like json.

### DIFF
--- a/outlines/generate/api.py
+++ b/outlines/generate/api.py
@@ -179,7 +179,6 @@ class SequenceGenerator:
             rng.seed()
 
         prompt_token_ids, attention_masks = self.tokenizer.encode(prompts)
-        rich.print(f"Input: [green]{self.tokenizer.decode(prompt_token_ids)[0]}")
 
         prompt_token_ids = prompt_token_ids.to(self.device)
         attention_masks = attention_masks.to(self.device)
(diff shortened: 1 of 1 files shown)


outlines/generate/api.py
```diff
@@ -191,6 +197,22 @@ def __call__(
...
+            final_fsm_states = []
+            for fsm, state, prompt_token_id_seq, attention_mask_seq in mit.zip_equal(
+                fsms, fsm_states, prompt_token_ids, attention_masks):
+                assert state == 0, f"Expected state to be 0, got {state}"
+                assert isinstance(state, int), f"Expected state to be an int, got {state}"
+                for token_id, attention_mask in zip(prompt_token_id_seq, attention_mask_seq):
+                    state = fsm.get_next_state(state=state, token_id=token_id.item())
+                    if state < 0:
+                        raise ValueError(f"Invalid state {state}")
+                final_fsm_states.append(state)
```
lapp0, line 210: Would it be better to implement this as a Guide? I've started work on https://github.com/dottxt-ai/outlines/pull/531and I'd like to see Guides become more composable.  

Also `interegular` has functionality built in to make this easy.

```
>>> import interegular
>>> fsm = interegular.parse_pattern("abcd").to_fsm()
>>> fsm.accepts("abcd")
True
>>> partial_fsm = fsm.derive("ab")
>>> partial_fsm.accepts("abcd")
False
>>> partial_fsm.accepts("cd")
True
```

PR #1137, opened 2024-09-08
Update vllm.md

americanthinker:
Minor type that prevents model from being downloaded from Huggingface hub.  Corrected the typo and verified that the download now works.

### DIFF
--- a/docs/reference/models/vllm.md
+++ b/docs/reference/models/vllm.md
@@ -73,7 +73,7 @@ vLLM supports AWQ, GPTQ and SqueezeLLM quantized models:
 ```python
 from outlines import models
 
-model = models.vllm("TheBloke/Llama2-7b-Chat-AWQ", quantization="awq")
+model = models.vllm("TheBloke/Llama-2-7b-Chat-AWQ", quantization="awq")
 model = models.vllm("TheBloke/Mistral-7B-Instruct-v0.2-GPTQ", quantization="gptq")
 model = models.vllm("https://huggingface.co/squeeze-ai-lab/sq-llama-30b-w4-s5", quantization="squeezellm")
 ```


docs/reference/models/vllm.md
```diff
@@ -73,7 +73,7 @@ vLLM supports AWQ, GPTQ and SqueezeLLM quantized models:
 ```python
 from outlines import models
 
-model = models.vllm("TheBloke/Llama2-7b-Chat-AWQ", quantization="awq")
+model = models.vllm("TheBloke/Llama-2-7b-Chat-AWQ", quantization="awq")
```
cpfiffer, line 76:
```suggestion
model = models.vllm("TheBloke/Llama-2-7B-Chat-AWQ", quantization="awq")
```

We should probably capitalize the B there, it seems like TheBloke has changed the name again. 

https://huggingface.co/TheBloke/Llama-2-7B-Chat-AWQ

cpfiffer: It redirects to the same place but worth doing anyway.

cpfiffer: Could you accept the suggestion here? I'd like to avoid future naming issues, since I'm not sure how long huggingface supports redirects.

americanthinker: @cpfiffer are you asking me?  I proposed the change because what's in the current docs doesn't work but I don't have merge authorization.

cpfiffer: Oh, you have to make the change on your branch, I cannot edit yours for you. You should be able to click "accept suggestion" in my first comment.

americanthinker: Done

cpfiffer: Great! Appreciate it 😄

cpfiffer (approved):

PR #1143, opened 2024-09-11
Change Outlines' logo

rlouf:


brandonwillard (approved):

docs/assets/images/logo.svg
cpfiffer, line 1: The svg file seems to be in sans

rlouf: It's because you don't have the font on your system I think

PR #1150, opened 2024-09-14
Fix MultinomialSampler hyperparameter bug

aidansan:
Currently seems like the MultinomialSampler does not utilize all logit_processors, only the last one in self.logit_processors. Thus, some of the hyperparameters (i.e. top_k) are not utilized. This fixes this issue with a small variable change.


lapp0 (commented):
Looks good to me. However I'm not sure we will be using `Sampler.__call__` in v0.1.0. 

Did you see this behavior cause an issue somewhere?

aidansan:
I initialized a sampler and passed it into a regex generator.

```python
sampler = outlines.samplers.MultinomialSampler(temperature=1.5, top_k=100, samples=5)
    model,
    r"I really love (\S*) cake with ice cream.",
    sampler=sampler
)
```

First `generate.api.SequenceGenerator.__call__` is called

which calls [ `generate.generator.sequence_generator`](https://github.com/dottxt-ai/outlines/blob/main/outlines/generate/api.py#L194
)

which calls [`sampler.__call__`](https://github.com/dottxt-ai/outlines/blob/main/outlines/generate/generator.py#L82 )

lapp0:
Was this with ExLllamaV2?

aidansan:
I'm not sure about `ExLlamaV2`, but maybe? This was my model initialization code `model = outlines.models.transformers("meta-llama/Meta-Llama-3-8B")`

Issue #1110, opened 2024-08-21
llama_cpp - JSON fails to generate when using Pydantic model with models.llama_cpp


willkurt:
### Describe the issue as clearly as possible:

When using `models.llamacpp` and creating JSON using a Pydantic model I get an error when generating the first result (see code to reproduce below). I have runt his code using `models.transformers` with no issue.

The model I'm using in this example is taken directly from the Chain of Thought cook book example, but I have also tried others and had the same issue.

### Steps/code to reproduce the bug:

```python
import llama_cpp
from outlines import generate, models
from textwrap import dedent

llama_tokenizer = llama_cpp.llama_tokenizer.LlamaHFTokenizer.from_pretrained(
            "NousResearch/Hermes-2-Pro-Llama-3-8B"
            )
tokenizer = llama_tokenizer.hf_tokenizer

model = models.llamacpp("NousResearch/Hermes-2-Pro-Llama-3-8B-GGUF",
            "Hermes-2-Pro-Llama-3-8B-Q4_K_M.gguf",
            tokenizer=llama_tokenizer,
            n_gpu_layers=-1,
            flash_attn=True,
            n_ctx=8192,
            verbose=False)

complaint_data = [{'message': 'Hi, my name is Olivia Brown.I recently ordered a knife set from your wellness range, and it arrived earlier this week. Unfortunately, my satisfaction with the product has been less than ideal.My order was A123456',
  'order_number': 'A12-3456',
  'department': 'kitchen'},
 {'message': 'Hi, my name is John Smith.I recently ordered a dress for an upcoming event, which was alleged to meet my expectations both in fit and style. However, upon arrival, it became apparent that the fabric was of subpar quality, leading to a less than satisfactory appearance.The order number is A12-3456',
  'order_number': 'A12-3456',
  'department': 'clothing'},
 {'message': 'Hi, my name is Sarah Johnson.I recently ordered the ultimate ChefMaster 8 Drawer Cooktop. However, upon delivery, I discovered that one of the burners is malfunctioning.My order was A458739',
  'order_number': 'A45-8739',
  'department': 'kitchen'}]

from pydantic import BaseModel, Field, constr
from enum import Enum


class Department(str, Enum):
    clothing = "clothing"
    electronics = "electronics"
    kitchen = "kitchen"
    automotive = "automotive"

class ComplaintData(BaseModel):
    first_name: str
    last_name: str
[...]

Issue #1106, opened 2024-08-20
Accelerate not installed in Docker Image


aowen14:
### Describe the issue as clearly as possible:

Not sure if this is a bug or a feature request. But `accelerate` apparently isn't installed in the docker image. Which means that one can either use transformers with no GPU acceleration, or vLLM. vLLM currently doesn't have feature parity with transformers from what I can tell (like `generate.json()`).

Running the code outside of the image with the library + `accelerate` works.`pip install accelerate` in the container works to solve the issue as well, and it seems like the marginal download was very small.

### Steps/code to reproduce the bug:

```python
#within an outlines docker container

from outlines import models

model = models.transformers("microsoft/Phi-3-mini-128k-instruct", device="cuda:0")
```

### Expected result:

```shell
The model should load and then be usable with the outlines sdk, with transformers and sending the model to a gpu, like `"cuda:0"`
```

### Error message:

```shell
Traceback (most recent call last):
  File "/outlines/Performance-Benchmarking/outlines_local_examples.py", line 81, in <module>
    model = models.transformers("microsoft/Phi-3-mini-128k-instruct", device="cuda:0")
  File "/usr/local/lib/python3.10/site-packages/outlines/models/transformers.py", line 430, in transformers
    model = model_class.from_pretrained(model_name, **model_kwargs)
  File "/usr/local/lib/python3.10/site-packages/transformers/models/auto/auto_factory.py", line 564, in from_pretrained
    return model_class.from_pretrained(
  File "/usr/local/lib/python3.10/site-packages/transformers/modeling_utils.py", line 3296, in from_pretrained
    raise ImportError(
ImportError: Using `low_cpu_mem_usage=True` or a `device_map` requires Accelerate: `pip install accelerate`
```

### Outlines/Python version information:

Docker Image Version Hash: `98c8512bd46f`

Version information
<details>
```
0.1.dev1+g8e94488.d20240816
Python 3.10.14 (main, Aug 13 2024, 02:10:16) [GCC 12.2.0]
aiohappyeyeballs==2.3.6
aiohttp==3.10.3
aiosignal==1.3.1
annotated-types==0.7.0
anyio==4.4.0
async-timeout==4.0.3
attrs==24.2.0
certifi==2024.7.4
charset-normalizer==3.3.2
click==8.1.7
cloudpickle==3.0.0
cmake==3.30.2
datasets==2.21.0
dill==0.3.8
diskcache==5.6.3
distro==1.9.0
[...]

Issue #985, opened 2024-06-19
Add a non_strict mode to deal with JSONDecodeError errors


davanstrien:
### What behavior of the library made you think about the improvement?

It might be nice to have the option to have a `non_strict` mode when doing generations for large batches of data. This could be particularly helpful for creating synthetic data where you usually don't care too much about skipping some prompts but do care about having failures to a computationally expensive pipeline. 

Currently, when using a generator constructed using a JSON Schema/Pydantic Class, i.e.

```python
generator = generate.json(model, AbstractDescriptions)
```

and calling the generator 

```python
results = generator(prompts, sampling_params=params)
``` 

I am running into `JSONDecodeError` errors with the source being a `ValidationError` for an Unterminated string. 

```python
ValidationError: 1 validation error for AbstractDescriptions
__root__
  Unterminated string starting at: line 1 column 631 (char 630) [type=value_error.jsondecode, input_value='{ "good":[ "]bad1_1_1_1_...1_1_1_1_1_1_1_1_1_1_1_1', input_type=str]
```

It is currently possible to attempt to fix this using different whitespace patterns, increasing `best_of` values or using a different LLM. 

This approach to addressing the issue works okay if the error occurs regularly, but it's a bit annoying if you process large amounts of prompts and only get the error late. One approach you can take now is to manage the generation in smaller batches with a try/except block and either reattempt the failed batch or skip it. 

### How would you like it to behave?

It might be nice instead to have an option flag to manage some exceptions when calling the generator and returning `None` for the failed generations, i.e. 

```python
results = generator(prompts, sampling_params=params, non_strict=True)
```

Which would return something like:

```python
[ JSON, JSON, None]
```

I am not very familiar with Outlines' internals, and I do not know how feasible it is to add this across all the LLM engines currently supported. This option would obviously make sense to be off by default, but IMO, it could be useful for some workloads where you care about the generations being correct but don't mind too much if one or two prompts result in `None` being generated. 

I couldn't find anything proposing this before (I might have missed an issue, but some other related issues:
https://github.com/outlines-dev/outlines/issues/759 
https://github.com/outlines-dev/outlines/issues/612

hugolytics:
I can second the need for this feature! Maybe it would be possible to fix by in the backend setting a pydantic [type_adapter](https://docs.pydantic.dev/latest/api/type_adapter/) that can choose between Union[YourModel,None], so basically if the input data doesn't validate, it will return None (or str, or whatever).

[...]

PR #1154, opened 2024-09-16
Fix Infinite Repetition in JSON Schemas Using Integer and String

lapp0:
# Overview

The [repetition problem](https://arxiv.org/abs/2012.14660) of language models combined with patterns allowing for infinite-length fields results in broken JSON Schema outputs.

This was addressed previously for infinite whitespaces issues by setting a [safe whitespace pattern](https://github.com/outlines-dev/outlines/pull/916) as the default. In this PR, the safety of whitespaces is extended to Integer and String patterns.

## Behavior

`json_schema.to_regex` now includes an kwarg `safe_subset=True`. 

`safe_subset=False` 
- Whitespace: `r"[\n\t ]*"`
- Integer: any number
- String: any string

`safe_subset=True` (default)
- Whitespace: `r"[ ]?"`
- Integer: (-1e19, 1e19)
- String: Any string of length (0, 256)

## Fixes

Safe Integer
- Fixes https://github.com/outlines-dev/outlines/issues/1110
- Fixes https://github.com/dottxt-ai/outlines/issues/1099

Safe String
- Fixes https://github.com/dottxt-ai/outlines/issues/1075
- Addresses https://github.com/dottxt-ai/outlines/issues/985 (doesn't fix, they requested a `non_strict` mode)
- Fixes https://github.com/outlines-dev/outlines/issues/1106
  - While this technically fixes their **error**, their **output** is still bad because they didn't apply a chat template https://github.com/dottxt-ai/outlines/issues/987

## Further Work

- **Important:** Low quality outputs are **often** caused by not applying a chat template. All examples should include chat templates, or user response quality will suffer. https://github.com/outlines-dev/outlines/issues/987

- Make code more failsafe: https://github.com/outlines-dev/outlines/issues/985

- `number` has no `safe_subset` implementation. Its' likely the only unsafe primitive remaining without one. There aren't any open issues for an error caused by number, however.


outlines/fsm/json_schema.py
```diff
@@ -41,7 +50,211 @@
...
+        raise ValueError("minLength must be greater than or equal to 0")
+
+    min_length = min_length or ""
+    if max_length is None:
+        max_length = str(SAFE_STR_MAX_LEN) if safe_subset else ""
+
+    return f'"{STRING_INNER}{{{min_length},{max_length}}}"'
+
+
+def build_regex_from_schema(
```
lapp0, line 255: Probably should be at the top. We might also consider separating into multiple modules so it's easier to read

rlouf: Yes, I generally like to move helper functions to the end of the file.

lapp0:
We might want to hold off on this one actually. I did some profiling on `get_str_pattern`. Constrained strings have a large state and take a long time to compile.

```
>>> len(interegular.parse_pattern(STRING_INNER + "{,256}").to_fsm().reduce().states)
513
>>> len(interegular.parse_pattern(STRING_INNER + "*").to_fsm().reduce().states)
2
```

The better alternative is to
- now: Update this PR, enable `safe_subset` as an optional parameter non-default parameter
- soon: Reduce failure rate via
  - https://github.com/dottxt-ai/outlines/issues/987
  - https://github.com/dottxt-ai/outlines/issues/985
- later: integrate this functionality into an automata which has a state stack, CFG or otherwise.

Let me know if this makes sense

PR #1156, opened 2024-09-16
Remove Broken pyairports Package, Replace with airportsdata

lapp0:
Fixes https://github.com/dottxt-ai/outlines/issues/1093

`pyairports` is unmaintained and often breaks `pip install outlines`. It's also out of date.

This PR replaces it with the more recently maintained and working package [airportsdata](https://github.com/mborsetti/airportsdata/)

### DIFF
--- a/outlines/types/airports.py
+++ b/outlines/types/airports.py
@@ -1,10 +1,11 @@
 """Generate valid airport codes."""
 from enum import Enum
 
-from pyairports.airports import AIRPORT_LIST
+import airportsdata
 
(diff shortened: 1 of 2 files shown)


lapp0:
Why does this test fail though? https://en.wikipedia.org/wiki/Charles_de_Gaulle_Airport

```
FAILED tests/test_types.py::test_type_enum[Airport-CDG-True] - assert False is True
```

cpfiffer (approved):
I'm all for this one!

PR #1159, opened 2024-09-17
Create optional generate.json(..., strict=False)

lapp0:
Fixes https://github.com/dottxt-ai/outlines/issues/985

# Problem
Sometimes models exhaust their available token length and generate a sequence which isn't valid JSON. This results in an error which is especially problematic with batch, as the entire batch is lost.

# Solution
Implement `strict` mode, which defaults to `True` (same behavior as `main`)

If `generate.json(..., strict=False)` is called, an error dict will be returned rather than an error being raised.

```
{
    'error': 'Unterminated string starting at: line 1 column 13 (char 12)', 
    'error_type': 'JSONDecodeError', 
    'output': '{"message"                                                        \n\t\t\t\t\t'
}


lapp0:
> Pydantic does offer [partial JSON parsing](https://docs.pydantic.dev/latest/concepts/json/#partial-json-parsing), is that something we'd want to use here?

IMO, users should expect their generation to comply with their schema or an error. Perhaps an `allow_partial` kwarg should be implemented separately if it fits with someones use case?

rlouf:
I meant in the "output" field of the error, sorry for not being more specific.

PR #1162, opened 2024-09-19
Reuse jinja environment for a prompt

jantrienes:
I really like the `@outlines.prompt` decorator. However, I noticed it is slow when formatting a large number of examples. The main inefficiency is creating the Jinja environment for every call of `render`. This PR makes a minor change to re-use the same Jinja environment per prompt. 

Note: This makes the `render` function obsolete. However, I kept it for backwards compatibility. Please advise if this should be further refactored.

## Benchmark

Small benchmark to test it:

```py
import timeit
import outlines

@outlines.prompt
def test_tpl(variable):
    """{{variable}} test"""

print('Formatted template:', test_tpl(variable='this is a'))

def test_render():
    test_tpl(variable='this is a')

execution_time = timeit.timeit(test_render, number=1000)
print(f"Execution time for 1000 runs: {execution_time} seconds")
```

With `main`:

```
Formatted template: this is a test
Execution time for 1000 runs: 0.11741420900216326 seconds
```

With this PR:

```
Formatted template: this is a test
Execution time for 1000 runs: 0.004638582991901785 seconds
```

### DIFF
--- a/outlines/prompts.py
+++ b/outlines/prompts.py
@@ -24,6 +24,7 @@ class Prompt:
 
     def __post_init__(self):
         self.parameters: List[str] = list(self.signature.parameters.keys())
+        self.jinja_environment = create_jinja_template(self.template)
 
     def __call__(self, *args, **kwargs) -> str:
         """Render and return the template.
@@ -35,7 +36,7 @@ class Prompt:
         """
         bound_arguments = self.signature.bind(*args, **kwargs)
         bound_arguments.apply_defaults()
-        return render(self.template, **bound_arguments.arguments)
+        return self.jinja_environment.render(**bound_arguments.arguments)
 
     def __str__(self):
         return self.template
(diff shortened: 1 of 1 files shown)


cpfiffer:
Big fan of this PR, free speed = dope. 

> Note: This makes the `render` function obsolete. However, I kept it for backwards compatibility. Please advise if this should be further refactored.

Could you add a comment or something to the `render` method so people are aware it's a no-op? 

I'm also curious if anyone's actually using it at the lowered level. I suspect this group of people is relatively small, in which case just removing the method is probably fine with an appropriate deprecation warning.

We might also like to have tests using 1-2 different variables as well to test that this still works in both cases.

rlouf:
Great PR! We are moving this logic to our new [prompts](https://github.com/dottxt-ai/prompts) library, and it would be awesome if you could submit a PR there as well!

PR #1165, opened 2024-09-21
Add support for URI format in JSON schema validation

Muhtasham:
This PR adds support for the URI format in JSON schema validation. Previously, attempting to use the "uri" format in a schema would raise a `NotImplementedError` especially when serving with vLLM. 

This change implements regex support for URI validation, allowing users to specify URI fields in their guided JSON requests.


rlouf:
Awesome, and good idea to refactor this! Would you mind adding a couple of tests?

outlines/fsm/json_schema.py
```diff
@@ -32,12 +32,14 @@
 DATE = r'"(?:\d{4})-(?:0[1-9]|1[0-2])-(?:0[1-9]|[1-2][0-9]|3[0-1])"'
 TIME = r'"(2[0-3]|[01][0-9]):([0-5][0-9]):([0-5][0-9])(\\.[0-9]+)?(Z)?"'
 UUID = r'"[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"'
+URI = r'"(https?:\/\/)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)"'
```
lapp0, line 35: 1) The `www.` is unnecessary

```
>>> pattern = r'"(https?:\/\/)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)"'
>>> re.match(pattern, '"http://www.example.com"')
<re.Match object; span=(0, 24), match='"http://www.example.com"'>
>>> pattern = r'"(https?:\/\/)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)"'
>>> re.match(pattern, '"http://www.example.com"')
<re.Match object; span=(0, 24), match='"http://www.example.com"'>
```

2) `@` can only appear once. You might separate the pattern so the domain portion is preceded by something like

`([-a-zA-Z0-9:%._\+~#=]+@)?` 

3) Could you please leave a commend or add documentation stating that URI only supports a subset of https://datatracker.ietf.org/doc/html/rfc3986 

4) `\b` isn't supported by `interegular`, you can simply remove this.

lapp0 (commented):
Thanks so much for adding this! A few changes to the pattern requested. Let me know if you have any questions.

lapp0 (approved):
LGTM provided tests pass.

RobinPicard:
Hi @Muhtasham! We recently released Outlines v1. Turning a json schema into a regex is now handled in outlines-core[https://github.com/dottxt-ai/outlines-core]. As the `URI` format is already supported there, I'm going to close the PR.

PR #1212, opened 2024-10-16
Add optional argument original_output to vLLM's generate function

### LINKED ISSUE #1199: Give the possibility to obtain the full response when calling the vLLM generate function
I'm using [InspectAI](https://inspect.ai-safety-institute.org.uk/) to evaluate language models. In particular, I'm evaluating the benefits of structured text using Outlines with language models. I would like to obtain the full response when calling the vLLM generate function since InspectAI expects to get the full response. Would it be possible to give the possibility to the user to get the full response. The default should still be the same as now which is a filtered response.

LouSalaun:
This pull request fixes #1199. An optional argument `original_output` is added to the vLLM's generate function: https://github.com/dottxt-ai/outlines/blob/a2fd35cfa58b9a9fb87e885a851af40019c10b62/outlines/models/vllm.py#L44-L53
`original_output` is a boolean. If True then the function returns the original (full) output of the vLLM API call.

### DIFF
--- a/outlines/models/vllm.py
+++ b/outlines/models/vllm.py
@@ -50,6 +50,7 @@ class VLLM:
         *,
         sampling_params: Optional["SamplingParams"] = None,
         use_tqdm: bool = True,
+        original_output: bool = False,
     ):
         """Generate text using vLLM.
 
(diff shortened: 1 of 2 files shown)


outlines/models/vllm.py
```diff
@@ -134,6 +140,10 @@ def generate(
             lora_request=self.lora_request,
             use_tqdm=use_tqdm,
         )
+
+        if original_output:
+            return results
```
cpfiffer, line 145: This part won't pass coverage checks because it isn't tested. Could you add a test here: https://github.com/dottxt-ai/outlines/blob/main/tests/generate/test_integration_vllm.py

cpfiffer (changes requested):

hedrergudene:
Any updates on this thread? It would be a highly desirable feature

PR #1228, opened 2024-10-24
add beam.cloud deployment example

mernit:
This PR shows an example of deployment an outlines API to beam.cloud.


rlouf:
Thank you! Could you run `pre-commit` locally to fix the styling issues?

PR #1236, opened 2024-10-31
fix: Make `convert_token_to_string` pickleable

saattrupdan:
This moves the `convert_token_to_string` helper function out as an independent function, to make it pickleable.

This is causing issues when storing logits processors in a class that needs to be pickleable.


saattrupdan:
@rlouf, does this seem good to you?

rlouf:
Tests were failing for unrelated reasons, I just rebased on `main`

RobinPicard:
Hi @saattrupdan ! We've recently released Outlines v1 that modifies a bit our models. What you were working on is now called VLLMOffline and is located in the `outlines/models/vllm_offline.py` file. Otherwise, it looks like what you propose is still applicable though.

PR #1240, opened 2024-11-04
Refactor the interface for the OpenAI API

rlouf:
Following #1220 we refactor the interface of the OpenAI integration. We build the `messages` argument to the client based on the input type, which can either be a simple string, or a `Vision` object that contains a prompt and an image (for vision models). We also build the `response_format` argument to the client based on the output format specified by the user, currently only `Json` since that is the only output format that OpenAI supports.

The `Vision` and `Json` object handle the conversion from the user input to the format that is expected from models, here Base64 and JSON Schema representations respectively. Even if those are not used at the higher-level user interface I think they are still a useful intermediate representation for that reason.

Overall the change simplifies the code and will make it easier to maintain.


outlines/models/openai.py
```diff
@@ -1,297 +1,154 @@
...
+from pydantic import BaseModel
 
-from outlines.base import vectorize
-from outlines.caching import cache
+__all__ = ["OpenAI"]
 
-__all__ = ["OpenAI", "openai", "azure_openai"]
 
+@dataclass
+class Vision:
```
rlouf, line 17: Move to the `prompts` module

outlines/models/openai.py
```diff
@@ -1,297 +1,154 @@
...
-    presence_penalty: float = 0
-    response_format: Optional[Dict[str, str]] = None
-    seed: Optional[int] = None
-    stop: Optional[Union[str, List[str]]] = None
-    temperature: float = 1.0
-    top_p: int = 1
-    user: str = field(default_factory=str)
 
+@dataclass
+class Json:
```
rlouf, line 39: Move to the `types` module

tests/models/test_openai.py
```diff
@@ -1,125 +1,107 @@
...
+    model = OpenAI(MODEL_NAME)
+
+    class Foo(BaseModel):
+        bar: int
+
+    schema = json.dumps(Foo.model_json_schema())
+
+    result = model.generate("foo?", Json(schema))
+    assert isinstance(result, str)
+    json.loads(result)
```
torymur, line 107:
```suggestion
    assert "bar" in json.loads(result)
```

outlines/models/openai.py
```diff
@@ -1,297 +1,156 @@
...
+    def format_none_output_type(self, _: None):
+        """Generate the `response_format` argument to the client when no
+        output type is specified by the user.
 
-    """
+        """
+        return {}
 
-    @error_handler
-    @cache()
```
torymur, line 199: I see `@cache()` is out, is there any plan for it?

rlouf: Good question. I think we need to re-think caching across the library. I am ok with dropping it for now. Wdyt?

torymur: Yeah, I agree 👍

torymur (approved):
LGTM! 🚀

PR #1243, opened 2024-11-05
Add Gemini integration

rlouf:
This PR provides a thin wrapper around the Gemini client. Gemini provides structured generation by accepting the following to define output types:

- A pydantic `BaseModel`. The output type is assumed to be JSON.
- A `TypedDict`. The output type is assumed to be JSON.
- A `List` type with a `BaseModel` or `TypedDict` argument. The output type is assumed to be a list of JSON objects.
- An `Enum`. The output type is assumed to be one of the fields in the enum.

This integration is already fairly complex:
1. We need to define a new `Choice` output type to handle the `Enum` output type;
2. The Gemini SDK does not accept JSON Schema string. `Json` thus needs to keep track of the original definition object. We should thus add a method to translate the definition object to JSON Schema instead of translating during initialization.
3. The Gemini SDK also accepts `TypedDict`s
4. The Gemini SDK accepts `List` type with arguments.

## TODO

- [x] Handle `list` and `List` more elegantly
- [x] Use `Json` to infer the type of the definition object / reject invalid definition objects
- [x] Add `Json` method to convert the definition object to JSON Schema
- [x] Use `Choice` to infer the type of the definition object / reject invalid definition objects

Builds on #1240.


outlines/models/gemini.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            raise NotImplementedError
+
+    @format_output_type.register(NoneType)
+    def format_none_output_type(self, output_type):
+        return {}
+
+    @format_output_type.register(Json)
+    def format_json_output_type(self, output_type):
+        if issubclass(output_type.original_definition, BaseModel):
```
rlouf, line 78: We could let the `Json` object do type inference for the definition, and use a simple `match` statement here.

outlines/models/gemini.py
```diff
@@ -0,0 +1,121 @@
...
+    def format_vision_input(self, model_input: Vision):
+        """Generate the `messages` argument to pass to the client when the user
+        passes a prompt and an image.
+
+        """
+        return {"contents": [model_input.prompt, model_input.image]}
+
+    @singledispatchmethod
+    def format_output_type(self, output_type):
+        if output_type.__origin__ == list:
```
rlouf, line 57: Having to handle `List` types here is not ideal.

outlines/models/gemini.py
torymur: +1, a good indication of this is that this part stays the same:
```
            return {
                "response_mime_type": "application/json",
                "response_schema": output_type.original_definition,
            }
```
Keeping this translation clean of any other extra logic is desirable to clearly separate concerns.

outlines/models/gemini.py
rlouf: Do you mean define our own `List` type?

torymur: Yeah, maybe it might worth clearly defining such custom types separately for the model. And since they're specific to the model they won't be under common ones:
 `from outlines.types import Choice, Json`

 But instead from somewhere like:
 `from outlines.gemini.types import List`?

Considering the variations of the models, it seems it might be quite common situation to manage the difference between them all. What do you think?

torymur: Yeah, extending the common when we are sure it's a good addition is ok, I wonder how to handle purely custom ones, which we wouldn't want to support for any other model but this particular one. For that we'd need some good examples, if List is out to be a purely custom one.

rlouf:
I am not a big fan of he `list` object, and would rather use `list[Object]`. We may however not have a choice;

outlines/types/__init__.py
```diff
@@ -25,10 +29,35 @@ class Json:
...
+    definition: Union[str, dict]
 
-        self.definition = definition
+    def to_json_schema(self):
+        if isinstance(self.definition, str):
+            schema = json.loads(self.definition)
+        elif isinstance(self.definition, dict):
+            schema = self.definition
+        elif isinstance(self.definition, type(BaseModel)):
+            schema = self.definition.model_json_schema()
```
rlouf, line 40: We also need to support `TypedDict`. We may be able to do this [via Pydantic](https://docs.pydantic.dev/2.3/usage/types/dicts_mapping/).

outlines/types/__init__.py
```diff
@@ -25,10 +29,35 @@ class Json:
...
+@dataclass
+class List:
+    definition: list
+
+
+@dataclass
+class Choice:
+    """Represents a multiple choice"""
+
+    definition: EnumMeta
```
rlouf, line 63: Can be a list of strings as well, which we can convert to `Enum`s

outlines/models/gemini.py
```diff
@@ -0,0 +1,116 @@
...
+            return {
+                "response_mime_type": "application/json",
+                "response_schema": output_type.definition,
+            }
+        elif isinstance(output_type.definition, _TypedDictMeta):
+            return {
+                "response_mime_type": "application/json",
+                "response_schema": output_type.definition,
+            }
+        else:
```
rlouf, line 83: We could support JSON Schema strings and JSON Schema dicts via [datamodel-code-generator](https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/#installation), but this is out of scope for this PR. Let's open an issue instead.

torymur (approved):
LGTM! 🚀

torymur:
> I am not a huge fan of the current design (re-defining `List` for instance). I would rather use `list[type]` but that would only make sense if `Json` and `Choice` were type builders. We can merge this and refactor later.

In current context `List` reads very appropriately alongside the others.

outlines/types/__init__.py
rlouf: Done, that was an easy change.

outlines/types/__init__.py
rlouf: Done, that was an easy change.

outlines/types/__init__.py
rlouf: Added.

PR #1249, opened 2024-11-07
Add a receipt processing cookbook

cpfiffer:
I added a receipt processing cookbook. 

- Uses Qwen or Pixtral
- General purpose message templating, no messy model-specific token adding
- Easy function for compressing images down for lower processing/memory requirements

Should help illustrate a simple use case for vision models.


rlouf:
Could you fix the merge conflict? Looks good otherwise.

cpfiffer:
Yep, fixed!

Issue #1232, opened 2024-10-28
<Please write a descriptive title>


cpfiffer:
### Describe the issue as clearly as possible:

When using outlines with the Llama 3.2 Vision model, simple regex pattern generation works, but JSON schema-based generation fails with index out of bounds errors.

### Steps/code to reproduce the bug:

```python
from io import BytesIO
import json
from urllib.request import urlopen
from PIL import Image
from pydantic import BaseModel
import outlines
from transformers import MllamaForConditionalGeneration, AutoProcessor
from transformers import AutoTokenizer
from outlines.fsm.json_schema import build_regex_from_schema

def img_from_url(url):
    img_byte_stream = BytesIO(urlopen(url).read())
    return Image.open(img_byte_stream).convert("RGB")

image_url="https://upload.wikimedia.org/wikipedia/commons/9/98/Aldrin_Apollo_11_original.jpg"
image= img_from_url(image_url)

model_id = "meta-llama/Llama-3.2-11B-Vision-Instruct"
tokenizer = AutoTokenizer.from_pretrained(model_id) 
model = MllamaForConditionalGeneration.from_pretrained(
    model_id,
    device_map="auto",
)
processor = AutoProcessor.from_pretrained(model_id)

llm = outlines.models.TransformersVision(model, tokenizer,processor)

# Works
pattern = "Mercury|Venus|Earth|Mars|Saturn|Jupiter|Neptune|Uranus|Pluto"
planet_generator = outlines.generate.regex(llm, pattern)
out = planet_generator(
    "What: <|image|>",
    [image]
)
print(out)
# Fails
class ImageData(BaseModel):
    caption: str

im_reg = build_regex_from_schema(json.dumps(ImageData.model_json_schema()))
print(im_reg)
generator = outlines.generate.regex(llm, im_reg)

out = generator(
    "What: <|image|>",
    [image]
)
print(out)
```

### Expected result:

```shell
Mercury, Venus, Earth, etc.
```

### Error message:

```shell
On CPU:


> python vision.py                                              ~/dottxt/outlines/debug
Loading checkpoint shards: 100%|██████████████████████████| 5/5 [00:42<00:00,  8.41s/it]
Pluto
\{[ ]?"caption"[ ]?:[ ]?"([^"\\\x00-\x1F\x7F-\x9F]|\\["\\])*"[ ]?\}
Traceback (most recent call last):
  File "/home/cameron/dottxt/outlines/debug/vision.py", line 44, in <module>
[...]

PR #1251, opened 2024-11-07
Fix IndexError caused by invalid token IDs in CFGGuide

RohitRathore1:
It fixes issue #1232 

These changes fix the IndexError caused by invalid token IDs in allowed_tokens_concat  by handling eos_token_id appropriately and adjusting token handling in CFGGuide. The updates maintain backward compatibility and ensure that existing functionality continues to work as expected.

Tested on CPU:

```
Loading checkpoint shards: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:08<00:00,  1.76s/it]
Saturn
\{[ ]?"caption"[ ]?:[ ]?"([^"\\\x00-\x1F\x7F-\x9F]|\\["\\])*"[ ]?\}
{"caption":"Command module pilot Buzz Aldrin walks across the lunar surface behind the deployed Lunar folloteneer's Ramp. The bottom of a Life Science Branch leg lock is framed in a footprint on the lunar surface behind the left leg of Aldrin's suit. The videocamera on the fullmomteiner's chest is visible atop the open hatch. Apollo 11, Aug. #42; CC AS11-40-5924,"}
```


outlines/fsm/guide.py
```diff
@@ -149,14 +152,20 @@ def get_next_instruction(self, state: CFGState) -> Instruction:
...
         if state.parser_state is None:
-            return Write(torch.tensor([self.eos_token_id]))
+            if self.eos_token_id is not None:
+                return Write(torch.tensor([self.eos_token_id]))
+            else:
+                return None  # No instruction if eos_token_id is not set
 
         valid_tokens = list(
-            self.iter_valid_token_ids(state, self.tokenizer.vocabulary.values())
+            self.iter_valid_token_ids(state, list(self.tokenizer.vocabulary.values()))
```
rlouf, line 161: Why converting this to a list?

RohitRathore1: The reason for converting `self.tokenizer.vocabulary.values()` to a list when passing it to the `iter_valid_token_ids` method is to ensure that we're working with a concrete, indexable collection of token IDs

rlouf:
Thank you for contributing! Could you add a test that fails on current `main` and passes here?

RohitRathore1:
> Thank you for contributing! Could you add a test that fails on current `main` and passes here?



> Thank you for contributing! Could you add a test that fails on current `main` and passes here?

Hi @rlouf to me it doesn't seem related to my changes. Let me have a deeper look

rlouf:
Thank you for making the change :) Is it good for review?

RohitRathore1:
> Thank you for making the change :) Is it good for review?

Hi @rlouf Yes, it is ready to review now. I have provided the correct model name which was the reason to fail  oneof our test case.

PR #1256, opened 2024-11-07
Add PDF cookbook

cpfiffer:
Adds a cookbook on extracting structured output from PDFs.

I included some extra bells and whistles here by showing how to do JSON, regex, and `choice`, which should help provide inspiration to people working with PDFs.


rlouf:
LGTM!

PR #1268, opened 2024-11-15
 Cookbook receipe on extract event details from text

scampion:
If useful, an example to illustrate how to extract event details from a message like :

```
Hello Kitty, my grandmother will be here , I think it's better to postpone our
appointment to review math lessons to next Monday at 2pm at the same place, 3 avenue des tanneurs, 
I think that one hour will be enough
see you 😘 
```

to 

```json
{
  "title":"Math lessons",
  "location":"3 avenue des tanneurs",
  "start":"2024-11-18T14:00:00Z"
}
```

Indeed, the next Monday is the 18 November, the smallest models cannot infer this example of ‘next Monday’ with precision and a real benchmark would be welcome, or maybe another approach with a function call

### DIFF
--- a/docs/cookbook/extract_event_details.md
+++ b/docs/cookbook/extract_event_details.md
@@ -1,0 +1,70 @@
+This recipe demonstrates how to use the `outlines` library to extract structured event details from a text message.
+We will extract the title, location, and start date and time from messages like the following:
+
+```plaintext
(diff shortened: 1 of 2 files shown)


scampion:
Need to mention https://github.com/jrudoler/ics-generator

PR #1269, opened 2024-11-16
Cookbook receipe on extract event details from text

scampion:
If useful, an example to illustrate how to extract event details from a message like :

```
Hello Kitty, my grandmother will be here , I think it's better to postpone our
appointment to review math lessons to next Friday at 2pm at the same place, 3 avenue des tanneurs, 
I think that one hour will be enough
see you 😘 
```

to 

```json
{
  "title":"Math lessons",
  "location":"3 avenue des tanneurs",
  "start":"2024-11-22T14:00:00Z"
}
```

Indeed, the next Friday is the 22 November, the smallest models cannot infer relative date like this example of ‘next Friday’ with precision and a real benchmark would be welcome, or maybe another approach with a function call

Btw, moving the python code in a dedicated file allow to benefit of the precommit hooks and homogenous code
In the future, it may also allow to test example dynamically gererate output to be sure that the documentation is consistant (like for the scikit learn documentation )

### DIFF
--- a/docs/cookbook/extract_event_details.md
+++ b/docs/cookbook/extract_event_details.md
@@ -1,0 +1,70 @@
+This recipe demonstrates how to use the `outlines` library to extract structured event details from a text message.
+We will extract the title, location, and start date and time from messages like the following:
+
+```plaintext
+Hello Kitty, my grandmother will be here, I think it's better to postpone
+our appointment to review math lessons to next Monday at 2pm at the same
+place, 3 avenue des tanneurs, one hour will be enough see you 😘
+```
+
+Let see how to extract the event details from the message.
+
+```python
+
+from datetime import datetime
+from pydantic import Field, BaseModel
+from outlines import models, generate
+
+# Load the model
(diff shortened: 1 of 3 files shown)


docs/cookbook/extract_event_details.md
```diff
@@ -0,0 +1,34 @@
...
+Hello Kitty, my grandmother will be here, I think it's better to postpone
+our appointment to review math lessons to next Monday at 2pm at the same
+place, 3 avenue des tanneurs, one hour will be enough see you 😘
+```
+
+Let see how to extract the event details from the message with the MLX
+library dedicated to  Apple Silicon processor  (M series).
+
+```python
+--8<-- "docs/cookbook/extract_event_details.py"
```
rlouf, line 14: Will that inline the content of the python script?

scampion: Yes and that python script will be evaluated by the precommit hooks. 

the mkdoc.yml need also the `pymdownx.snippets:` config

rlouf: Nice, didn't know about this

rlouf:
Thank you!

PR #1277, opened 2024-11-21
Add json call with multi-function enums

g-prz:
This PR aims at solving #1217

Issue #1275, opened 2024-11-19
outlines.generate.choice does not escape regex characters


cpfiffer:
### Describe the issue as clearly as possible:

Providing strings that include regex characters like `.` are not escaped, and are directly interpreted when using `outlines.generate.choice`.

For example, this can lead to the choice generator

```python
# Choices
choices = ["Dr. Smith", "Prof. Jones"]

# Generator
riddler = generate.choice(model, choices)
```

producing an output like

```
Dru Smith
```

where the `u` should in fact be the literal `.`

### Steps/code to reproduce the bug:

```python
# Best import of all time
from outlines import models, generate
import torch
from transformers import AutoTokenizer

# Load a language model into memory
model = models.transformers(
    "microsoft/Phi-3-mini-4k-instruct",
    device="cuda",
    model_kwargs={"torch_dtype": torch.bfloat16},
)

tokenizer = AutoTokenizer.from_pretrained("microsoft/Phi-3-mini-4k-instruct")
def to_prompt(text):
    return tokenizer.apply_chat_template([text], tokenize=False)

riddle = """
Dr. Smith teaches either Biology or Chemistry.
Prof. Jones teaches the subject Dr. Smith doesn't teach.
If the Biology teacher wears glasses,
and Prof. Jones doesn't wear glasses, who teaches Biology?
"""

# Choices
choices = ["Dr. Smith", "Prof. Jones"]

# Generator
riddler = generate.choice(model, choices)

# Generate a response
response = riddler(to_prompt(riddle))
print(response)
```

### Expected result:

```shell
"Dr. Smith"
```

### Error message:

```shell
N/A
```

### Outlines/Python version information:

Version information
<details>
```
accelerate==1.1.1
aiohappyeyeballs==2.4.3
aiohttp==3.11.6
aiosignal==1.3.1
airportsdata==20241001
annotated-types==0.7.0
asttokens==2.4.1
async-timeout==5.0.1
attrs==24.2.0
certifi==2024.8.30
charset-normalizer==3.4.0
cloudpickle==3.1.0
comm==0.2.2
datasets==3.1.0
debugpy==1.8.8
decorator==5.1.1
dill==0.3.8
diskcache==5.6.3
exceptiongroup==1.2.2
executing==2.1.0
filelock==3.16.1
frozenlist==1.5.0
fsspec==2024.9.0
huggingface-hub==0.26.2
idna==3.10
interegular==0.3.3
ipykernel==6.29.5
[...]

Issue #1218, opened 2024-10-20
Use `Enum`s for multiple choices


rlouf:
### Presentation of the new feature

To make the model choose between different options we currently use `generate.choice`. Here I suggest we also use `Enum`s to represent multiple choices:

```python 
from enum import Enum


class choices(Enum):
    a: str = "Choice A"
    b: str = "Choice B"
```

### Where does it fit in Outlines?

This will probably make more sense once the API has been refactored.

PR #1279, opened 2024-11-22
Handle enum in choice

g-prz:
This PR aims at solving #1218 (requires #1277 to be merged)

Quick fix to solve #1275

### DIFF
--- a/outlines/generate/choice.py
+++ b/outlines/generate/choice.py
@@ -1,7 +1,9 @@
 import json as pyjson
+from enum import Enum
 from functools import singledispatch
-from typing import Callable, List
+from typing import Callable, List, Union
 
+from outlines.fsm.json_schema import build_regex_from_schema, get_schema_from_enum
 from outlines.generate.api import SequenceGeneratorAdapter
 from outlines.models import OpenAI
 from outlines.samplers import Sampler, multinomial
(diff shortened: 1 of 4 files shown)


outlines/generate/choice.py
```diff
@@ -12,12 +15,19 @@
...
+    if isinstance(choices, type(Enum)):
+        regex_str = build_regex_from_schema(pyjson.dumps(get_schema_from_enum(choices)))
+    else:
+        choices = [re.escape(choice) for choice in choices]  # type: ignore
+        regex_str = r"(" + r"|".join(choices) + r")"
 
     generator = regex(model, regex_str, sampler)
-    generator.format_sequence = lambda x: x
+    if isinstance(choices, type(Enum)):
+        generator.format_sequence = lambda x: pyjson.loads(x)
```
g-prz, line 28: Hey @rlouf 🙋‍♂️ 
I have a design related question: in case we pass an enum to `choice`, are we expecting to return an instance of that enum or its value? If the latter we are fine, if the former we need to reconstruct the enum in here
Thx 🙏

rlouf: Not sure what you mean but returning N instance, but we definitely just want the value.

g-prz: We are good for review then

rlouf:
Thank you for contributing!

PR #1207, opened 2024-10-13
Add jax compatible api

sky-2002:
This PR adds a JAX compatible API, refer issue #1027 

Looking forward for review and open to feedback(especially in writing tests).

PR #1297, opened 2024-11-28
Fix import in test base processor

g-prz:
Quick fix for jax.numpy import (following #1207)

### DIFF
--- a/tests/processors/test_base_processor.py
+++ b/tests/processors/test_base_processor.py
@@ -1,6 +1,5 @@
 from typing import List
 
-import jax.numpy as jnp
 import numpy as np
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


g-prz:
It is actually imported in a try/except [clause](https://github.com/dottxt-ai/outlines/pull/1297/commits/ea4e185b81c12af076c2080e06281cfea933cedc#diff-6f68118774f2f08cb589f033f05d0552d1dd7c46737b0f6ca500d863607ded83R22-R27) and creates a jax array accordingly (test is raising an error in the CI otherwise)

rlouf:
I updated the requirements for the tests in #1296

g-prz:
Nice, then we can close that PR
Thx 🙏

PR #1298, opened 2024-11-29
Refactor the llama.cpp interface

rlouf:


rlouf:
@torymur this is ready for review

docs/reference/models/llamacpp.md
```diff
@@ -69,87 +57,39 @@ See the [llama-cpp-python documentation](https://llama-cpp-python.readthedocs.io
...
 ```
 
-This also works with generators built with `generate.regex`, `generate.json`, `generate.cfg`, `generate.format` and `generate.choice`.
 
-### Load LoRA adapters
+## Generate text
+
 
-You can load LoRA adapters dynamically:
+To generate text you must first create a `Generator` object by passing the model instance and, possibley, the expected output type:
```
torymur, line 71:
```suggestion
To generate text you must first create a `Generator` object by passing the model instance and, possibly, the expected output type:
```

docs/reference/models/llamacpp.md
```diff
@@ -69,87 +57,39 @@ See the [llama-cpp-python documentation](https://llama-cpp-python.readthedocs.io
...
 ```python
-from outlines import models, generate
-
-
-model = models.llamacpp("TheBloke/phi-2-GGUF", "phi-2.Q4_K_M.gguf")
-generator = generate.text(model)
-
-answer = generator("A prompt", presence_penalty=0.8)
+tokens = generator.stream("A prompt")
 ```
```
torymur, line 91: 🔥

docs/reference/models/openai.md
```diff
@@ -163,42 +71,64 @@ client = AsyncOpenAI(
...
 
-It is possible to specify the values for `seed`, `presence_penalty`, `frequence_penalty`, `top_p` by passing an instance of `OpenAIConfig` when initializing the model:
+## Models that follow the OpenAI standard
+
+Outlines supports models that follow the OpenAI standard. You will need to initialize the OpenAI client properly configured and pass it to `outlines.models.OpenAI`
 
 ```python
-from outlines.models.openai import OpenAIConfig
+import os
+from openai import AsyncOpenAI
```
torymur, line 82: I don't think it's needed for this example:

```suggestion
```

outlines/generate/__init__.py
```diff
@@ -6,3 +12,45 @@
...
+        if isinstance(model, get_args(APIModel)):
+            self.output_description = output_type
+        else:
+            if output_type is not None:
+                regex_string = output_type.to_regex()
+                self.output_description = RegexLogitsProcessor(
+                    regex_string, model.tokenizer
+                )
+            else:
+                self.output_description = None
```
torymur, line 53: You only manipulating with `self.output_description` and don't assign `self.model = model`, which would fully redefine class attrs in `__init__` and could be okay, but for such partial redefinition in `@dataclass` case I see people suggest `__post_init__`, which will be called after `__init__`? But I'm not sure here.

Overall, took me a second here to understand this fork: `APIModel` <-> `output type`, `Model` <-> `processor`. I can't put my finger on it with a proper design idea, but the fact that we have two good matching pairs suggests a good potential for separation.

torymur (approved):
Sorry for delay!
Nice to see The Generator in docs & in the code 🔥 Left some minor comments, but overall LGTM! 🚀

outlines/generate/__init__.py
rlouf: Yeah I agree, maintain a single interface for API-based models and local models is annoying.

One possible way out of this design issue is to have models build and hold the `LogitsProcessor` but it does feel unintuitive. The other solution is to add a `Generator` building function that dispatches to an `APIGenerator` and a `LogitsProcessorGenerator`, but I feel the duplication of code is not necessarily justified. It might be more a case of documenting the behavior better?

I'll see if I can come up with something better.

torymur: > The other solution is to add a Generator building function that dispatches to an APIGenerator and a LogitsProcessorGenerator, but I feel the duplication of code is not necessarily justified.

Actually, this might be enough, I'd vote for duplications like this to make design more self-descriptive rather than bringing it on a higher level of clarifications in docs

rlouf: I made the change. It actually does not look like duplication so much, and makes the code much easier to follow.

torymur: It was worth it, crystal clear now 🔥

torymur (approved):

PR #1306, opened 2024-12-02
Fix GuideLogitsProcessor for MPS device

### LINKED ISSUE #1282: Cookbook "Receipt Data Extraction with VLMs" is not reproducible

I tried to run the Cookbook ["Receipt Data Extraction with VLMs"](https://github.com/dottxt-ai/outlines/blob/b55d31463cb6ed38fc0109e018f53ce0cdafbe19/docs/cookbook/receipt-digitization.md) using "Qwen/Qwen2-VL-**2**B-Instruct" but ran into the described error. In short, the model seems to output just a `!` character, which leads to a json parsing error.

A colleague of mine was able to reproduce the error on his machine.
Both our machines are MacBooks.

What is also interesting: downgrading to transformers 4.45 leads to the same error, but with output ` ``` `.

```python
# As described in the cookbook

# LLM stuff
import outlines
import torch
from transformers import AutoProcessor, AutoModelForCausalLM
from typing import Literal, Optional, List

# Image stuff
from PIL import Image
import requests

# Rich for pretty printing
from rich import print

import outlines.caching as cache

cache.disable_cache()

# To use Qwen-2-VL:
from transformers import Qwen2VLForConditionalGeneration
model_name = "Qwen/Qwen2-VL-2B-Instruct"
model_class = Qwen2VLForConditionalGeneration

model = outlines.models.transformers_vision(
    model_name,
    model_class=model_class,
    model_kwargs={
        "device_map": "auto",
        "torch_dtype": torch.bfloat16,
        "trust_remote_code": True
    },
    processor_kwargs={
        "device": "mps", # set to "cpu" if you don't have a GPU
    },
)

def load_and_resize_image(image_path, max_size=1024):
    """
    Load and resize an image while maintaining aspect ratio

    Args:
        image_path: Path to the image file
        max_size: Maximum dimension (width or height) of the output image

    Returns:
        PIL Image: Resized image
    """
    image = Image.open(image_path)

    # Get current dimensions
    width, height = image.size

    # Calculate scaling factor
    scale = min(max_size / width, max_size / height)

    # Only resize if image is larger than max_size
    if scale < 1:
        new_width = int(width * scale)
        new_height = int(height * scale)
        image = image.resize((new_width, new_height), Image.Resampling.LANCZOS)

    return image

# Path to the image
image_path = "https://github.com/dottxt-ai/outlines/raw/main/docs/cookbook/images/trader-joes-receipt.jpg"

# Download the image
response = requests.get(image_path)
with open("receipt.png", "wb") as f:
    f.write(response.content)

# Load + resize the image
image = load_and_resize_image("receipt.png")

class Item(BaseModel):
    name: str
    quantity: Optional[int]
    price_per_unit: Optional[float]
    total_price: Optional[float]

class ReceiptSummary(BaseModel):
    store_name: str
    store_address: str
    store_number: Optional[int]
    items: List[Item]
    tax: Optional[float]
    total: Optional[float]
    # Date is in the format YYYY-MM-DD. We can apply a regex pattern to ensure it's formatted correctly.
    date: Optional[str] = Field(pattern=r'\d{4}-\d{2}-\d{2}', description="Date in the format YYYY-MM-DD")
    payment_method: Literal["cash", "credit", "debit", "check", "other"]

# Set up the content you want to send to the model
messages = [
    {
        "role": "user",
        "content": [
            {
                # The image is provided as a PIL Image object
                "type": "image",
                "image": image,
            },
            {
                "type": "text",
                "text": f"""You are an expert at extracting information from receipts.
                Please extract the information from the receipt. Be as detailed as possible --
                missing or misreporting information is a crime.

                Return the information in the following JSON schema:
                {ReceiptSummary.model_json_schema()}
            """},
        ],
    }
]

# Convert the messages to the final prompt
processor = AutoProcessor.from_pretrained(model_name)
prompt = processor.apply_chat_template(
    messages, tokenize=False, add_generation_prompt=True
)

# Prepare a function to process receipts
receipt_summary_generator = outlines.generate.json(
    model,
    ReceiptSummary,

    # Greedy sampling is a good idea for numeric
    # data extraction -- no randomness.
    sampler=outlines.samplers.greedy()
)

# Generate the receipt summary
result = receipt_summary_generator(prompt, [image])
print(result)
```

### Expected result:

```shell
# As described in the cookbook
```

### Error message:

```shell
The argument `trust_remote_code` is to be used with Auto classes. It has no effect here and is ignored.
`Qwen2VLRotaryEmbedding` can now be fully parameterized by passing the model config through the `config` argument. All other arguments will be removed in v4.46
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:04<00:00,  2.07s/it]
.venv/lib/python3.11/site-packages/transformers/generation/configuration_utils.py:590: UserWarning: `do_sample` is set to `False`. However, `temperature` is set to `0.0` -- this flag is only used in sample-based generation modes. You should set `do_sample=True` or unset `temperature`. This was detected when initializing the generation config instance, which means the corresponding file may hold incorrect parameterization and should be fixed.
  warnings.warn(
.venv/lib/python3.11/site-packages/transformers/generation/configuration_utils.py:590: UserWarning: `do_sample` is set to `False`. However, `temperature` is set to `0.0` -- this flag is only used in sample-based generation modes. You should set `do_sample=True` or unset `temperature`.
  warnings.warn(
huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks...
To disable this warning, you can either:
	- Avoid using `tokenizers` before the fork if possible
	- Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true | false)
  File "./receipt-digitization.py", line 132, in <module>
    result = receipt_summary_generator(prompt, [image])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "outlines/generate/api.py", line 565, in __call__
    return self._format(completions)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "outlines/generate/api.py", line 488, in _format
    return self.format_sequence(sequences)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "outlines/generate/json.py", line 50, in <lambda>
    generator.format_sequence = lambda x: schema_object.model_validate_json(x)
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.11/site-packages/pydantic/main.py", line 651, in model_validate_json
    return cls.__pydantic_validator__.validate_json(json_data, strict=strict, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for ReceiptSummary
  Invalid JSON: expected value at line 1 column 1 [type=json_invalid, input_value='!', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/json_invalid
```

Version information
<details>
```
0.1.5.dev0+gc406da8.d20241121
Python 3.11.10 (main, Oct 16 2024, 08:56:36) [Clang 18.1.8 ]
accelerate==1.1.1
aiohappyeyeballs==2.4.3
aiohttp==3.11.6
aiosignal==1.3.1
airportsdata==20241001
annotated-types==0.7.0
anyio==4.6.2.post1
attrs==24.2.0
beartype==0.15.0
certifi==2024.8.30
cfgv==3.4.0
chardet==5.2.0
charset-normalizer==3.4.0
cloudpickle==3.1.0
coverage==7.6.7
cramjam==2.9.0
datasets==3.1.0
diff_cover==9.2.0
dill==0.3.8
diskcache==5.6.3
distlib==0.3.9
distro==1.9.0
exllamav2==0.2.4
fastparquet==2024.11.0
filelock==3.16.1
frozenlist==1.5.0
fsspec==2024.9.0
h11==0.14.0
httpcore==1.0.7
httpx==0.27.2
huggingface-hub==0.26.2
identify==2.6.2
idna==3.10
iniconfig==2.0.0
interegular==0.3.3
Jinja2==3.1.4
jiter==0.7.1
jsonschema==4.23.0
jsonschema-specifications==2024.10.1
lark==1.2.2
llama_cpp_python==0.3.2
markdown-it-py==3.0.0
MarkupSafe==3.0.2
mdurl==0.1.2
mlx==0.20.0
mlx-lm==0.19.3
mpmath==1.3.0
multidict==6.1.0
multiprocess==0.70.16
nest-asyncio==1.6.0
networkx==3.4.2
ninja==1.11.1.1
nodeenv==1.9.1
numpy==1.26.4
openai==1.55.0
-e git+https://github.com/dottxt-ai/outlines.git@c406da8f9ecacb96b93982b8acaad2f5d12dab6e#egg=outlines
outlines_core==0.1.17
packaging==24.2
pandas==2.2.3
pillow==11.0.0
platformdirs==4.3.6
pluggy==1.5.0
pre_commit==4.0.1
propcache==0.2.0
protobuf==5.28.3
psutil==6.1.0
py-cpuinfo==9.0.0
pyarrow==18.0.0
pycountry==24.6.1
pydantic==2.10.0
pydantic_core==2.27.0
Pygments==2.18.0
pytest==8.3.3
pytest-benchmark==5.1.0
pytest-cov==6.0.0
pytest-mock==3.14.0
python-dateutil==2.9.0.post0
pytz==2024.2
PyYAML==6.0.2
referencing==0.35.1
regex==2024.11.6
requests==2.32.3
responses==0.25.3
rich==13.9.4
rpds-py==0.21.0
safetensors==0.4.5
sentencepiece==0.2.0
six==1.16.0
sniffio==1.3.1
sympy==1.13.1
tokenizers==0.20.3
torch==2.5.1
tqdm==4.67.0
transformers==4.46.3
typing_extensions==4.12.2
tzdata==2024.2
urllib3==2.2.3
virtualenv==20.27.1
websockets==14.1
xxhash==3.5.0
yarl==1.18.0
```
</details>

### Context for the issue:

_No response_

hoesler:
While debugging #1282, I found the the issue to be caused by `non_blocking=True` on a Mac with an MPS device.

The usage of `non_blocking=True` is only safe for CPU->GPU, as far as I understand from [this guide](https://pytorch.org/tutorials/intermediate/pinmem_nonblock.html#other-copy-directions-gpu-cpu-cpu-mps).
For other directions, especially CPU->MPS in my case, this results in a all-zero vector instead of the real token ids which generates wrong tokens and results in errors like described in #1282.
I also tried to use `torch.mps.synchronize()`, but it doesn't help.

I'm haven't benchmarked the difference in speed, but I suspect it to be neglectable because the created vector is accessed [directly afterwards](https://github.com/dottxt-ai/outlines/blob/36f1bf27a80f5877b88a915749341664cef4a1ab/outlines/processors/structured.py#L115).

Fixes #1282

### DIFF
--- a/benchmarks/bench_processors.py
+++ b/benchmarks/bench_processors.py
@@ -37,15 +37,12 @@ def get_mock_processor_inputs(array_library, num_tokens=30000):
     logits: (4, 30,000 ) dtype=float
     input_ids shape: (4, 2048) dtype=int
     """
-    if array_library == "torch":
-        logits = torch.rand((4, num_tokens), dtype=torch.float)
+    if array_library.startswith("torch"):
+        device = array_library.split("_")[1] if "_" in array_library else "cpu"
+
+        logits = torch.rand((4, num_tokens), dtype=torch.float, device=device)
         input_ids = torch.randint(
-            low=0, high=num_tokens, size=(4, 2048), dtype=torch.int
-        )
-    elif array_library == "torch_cuda":
-        logits = torch.rand((4, num_tokens), dtype=torch.float, device="cuda")
-        input_ids = torch.randint(
-            low=0, high=num_tokens, size=(4, 2048), dtype=torch.int, device="cuda"
+            low=0, high=num_tokens, size=(4, 2048), dtype=torch.int, device=device
         )
     elif array_library == "numpy":
         logits = np.random.rand(4, num_tokens).astype(np.float32)
@@ -88,6 +85,8 @@ class LogitsProcessorPassthroughBenchmark:
         params += ["mlx"]
     if torch.cuda.is_available():
         params += ["torch_cuda"]
+    if torch.mps.is_available():
+        params += ["torch_mps"]
     if is_jax_allowed():
         params += ["jax"]
 
@@ -108,9 +107,10 @@ class LogitsProcessorStructuredBenchmark:
     array_libraries = ["torch", "numpy"]
     if is_mlx_lm_allowed():
         array_libraries += ["mlx"]
-    # PR TODO
     if torch.cuda.is_available():
         array_libraries += ["torch_cuda"]
+    if torch.mps.is_available():
+        array_libraries += ["torch_mps"]
 
     # accept very many or very few tokens, respectively
     patterns = [r"[^Z]*", "Z*"]
--- a/outlines/processors/structured.py
+++ b/outlines/processors/structured.py
@@ -108,7 +108,7 @@ class GuideLogitsProcessor(OutlinesLogitsProcessor):
         batch_indices = []
         for i, guide_state in enumerate(sequence_states):
             allowed_tokens = self.guide.get_next_instruction(guide_state).tokens.to(
-                mask.device, non_blocking=True
+                mask.device
             )
             allowed_tokens_batch.append(allowed_tokens)
             batch_indices.append(
--- a/outlines/processors/structured.py
+++ b/outlines/processors/structured.py
@@ -102,22 +102,19 @@ class GuideLogitsProcessor(OutlinesLogitsProcessor):
 
             sequence_states.append(self._guide_states[curr_state_key])
 
-        mask = torch.ones_like(logits, dtype=torch.bool)
-
         allowed_tokens_batch = []
         batch_indices = []
         for i, guide_state in enumerate(sequence_states):
-            allowed_tokens = self.guide.get_next_instruction(guide_state).tokens.to(
-                mask.device
-            )
+            allowed_tokens = self.guide.get_next_instruction(guide_state).tokens
             allowed_tokens_batch.append(allowed_tokens)
             batch_indices.append(
                 torch.full_like(allowed_tokens, i)
             )  # Store batch index for each allowed token
 
-        allowed_tokens_concat = torch.cat(allowed_tokens_batch)
-        batch_indices_concat = torch.cat(batch_indices)
+        allowed_tokens_concat = torch.cat(allowed_tokens_batch).to(logits.device)
+        batch_indices_concat = torch.cat(batch_indices).to(logits.device)
 
+        mask = torch.ones_like(logits, dtype=torch.bool)
         mask[batch_indices_concat, allowed_tokens_concat] = False
         logits.masked_fill_(mask, float("-inf"))
 


cpfiffer:
I think we'd want to see some benchmarks + tests on this one for sure -- I'm worried that this might have some unanticipated consequences. 

I'm curious if it's also possible to specify this by a keyword argument somewhere.

hoesler:
> I think we'd want to see some benchmarks + tests on this one for sure -- I'm worried that this might have some unanticipated consequences.

As I cant't do a benchmark comparison on my MPS machine, would you mind doing that? Which tests in addition to the existing ones do you imagine?
I think the unwanted consequences are what we see now by enabling that feature for potential performance benefits. Reading the linked article, I guess blocking mode is the default for a reason.
Maybe @lapp0 can comment on that, as he introduced this code in #1192.

> I'm curious if it's also possible to specify this by a keyword argument somewhere.

Do you suggest here to run different code depending on the device in use or by a user setting?

lapp0:
I believe this was introduced before MLX was integrated. Good find. Could you see what this does to the benchmarks on your Apple Silicon machine?

dylanjcastillo:
This also fixes #1316

hoesler:
@lapp0 I obviously can't compare blocking vs non_blocking but with the fix, this is what I get:

```
asv run --bench bench_processors.LogitsProcessorStructuredBenchmark.time_structured_generation main..fix-guide-logits-processor-mps
Couldn't load asv.plugins._mamba_helpers because
No module named 'libmambapy'
· Creating environments
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[ 0.00%] · For Outlines commit aa333be9 <fix-guide-logits-processor-mps>:
[ 0.00%] ·· Benchmarking virtualenv-py3.11
[50.00%] ··· Running (bench_processors.LogitsProcessorStructuredBenchmark.time_structured_generation--).
[100.00%] ··· bench_processors.LogitsProcessorStructuredBenchmark.time_structured_generation                                                                                                                        ok
[100.00%] ··· ======================== ============ ============
              --                                 param2
              ------------------------ -------------------------
               array_library, pattern     [^Z]*          Z*
              ======================== ============ ============
                       torch             16.5±1ms     106±10μs
                       numpy             17.5±1ms     120±10μs
                        mlx             19.1±0.5ms    185±10μs
                     torch_mps          18.6±0.3ms   1.73±0.1ms
              ======================== ============ ============
```

I don't know what to make of the overall significant difference between patterns and especially the huge slowdown for torch_mps for pattern Z*. The latter could be related to MPS performing worse if the number of dims is small (see https://github.com/pytorch/pytorch/issues/77799).

hoesler:
Another idea: Wouldn't it be even better to remove the necessity of device synchronization by creating the `allowed_tokens` tensors directly on the device?

```python
        with torch.device(mask.device):
            for i, guide_state in enumerate(sequence_states):
                allowed_tokens = self.guide.get_next_instruction(guide_state).tokens
                allowed_tokens_batch.append(allowed_tokens)
                batch_indices.append(
                    torch.full_like(allowed_tokens, i)
                )  # Store batch index for each allowed token
```

This seems to be [the recommended approach](https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#create-tensors-directly-on-the-target-device).

hoesler:
The following seems to perform even better. Concat the indexing tensors in cpu and move only these to target device:

```python
        allowed_tokens = self.guide.get_next_instruction(guide_state).tokens
        ....
        allowed_tokens_concat = torch.cat(allowed_tokens_batch).to(logits.device)
        batch_indices_concat = torch.cat(batch_indices).to(logits.device)
```

```
              ------------------------ -----------------------
               array_library, pattern     [^Z]*         Z*
              ======================== ============ ==========
                       torch             17.4±2ms    90.4±6μs
                       numpy            16.6±0.9ms   89.4±5μs
                        mlx             18.4±0.7ms   139±9μs
                     torch_mps          19.0±0.9ms   775±30μs
              ======================== ============ ==========
```

This also removes the need to use non-blocking syncs. Will push that.

rlouf:
Thank you for the fix ⭐

PR #1308, opened 2024-12-02
Add TypeAdapter support for JSON

hoesler:
Add support for pydantic [TypeAdapter](https://docs.pydantic.dev/latest/api/type_adapter/) to allow schema objects to be generated from primitive types or lists for example (e.g. `TypeAdapter(List[Transaction])`).

### DIFF
--- a/outlines/generate/json.py
+++ b/outlines/generate/json.py
@@ -1,9 +1,9 @@
 import json as pyjson
 from enum import Enum
 from functools import singledispatch
-from typing import Callable, Optional, Union
+from typing import Callable, Optional, Type, Union, cast
 
-from pydantic import BaseModel
+from pydantic import BaseModel, TypeAdapter
 
 from outlines.fsm.json_schema import (
     build_regex_from_schema,
@@ -20,7 +20,7 @@ from .regex import regex
 @singledispatch
 def json(
     model,
-    schema_object: Union[str, object, Callable],
+    schema_object: Union[str, Type[BaseModel], TypeAdapter, Type[Enum], Callable],
     sampler: Sampler = multinomial(),
     whitespace_pattern: Optional[str] = None,
 ) -> SequenceGeneratorAdapter:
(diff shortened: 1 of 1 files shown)


rlouf:
Thank you! Could you add some tests?

hoesler:
I actually looked into adding a test, but found that the [relevant test is skipped](https://github.com/dottxt-ai/outlines/blob/36f1bf27a80f5877b88a915749341664cef4a1ab/tests/generate/test_generate.py#L255). So what would you suggest here?

hoesler:
BTW, I guess this could also replace the recently added enum support from #1277, as you can just do something like:

```
class Foo(Enum):
    x: 1
    y: 2

generator = outlines.generate.json(model, TypeAdapter(Foo))
```

rlouf:
Could you give me an example of how you would use this?

hoesler:
My motivation was to generate a list of objects:

```python
class Transaction(BaseModel):
   ...

generator = outlines.generate.json(model, TypeAdapter(List[Transaction]))
```

But one could use it for primitives as well. See for example the comment above for Enums.

rlouf:
I wonder if there is a way to let users pass `List[Object]` and hide the use of `TypeAdapter` from them?

hoesler:
Sure, I could do that. But I don’t know if I would recommend it. While convenient, the function might become a bit hard to extend in the future or simply complex to define. Maybe there will be types you want to handle differently, maybe another library other than Pydantic will be added… But in the end it’s your decision.Am 08.12.2024 um 16:44 schrieb Rémi Louf ***@***.***>:﻿
I wonder if there is a way to let users pass List[Object] and hide the use of TypeAdapter from them?

rlouf:
You're probably right, we'll leave it as is for now. We just need to get to full coverage before merging!

hoesler:
Happy to add tests, but please give me some guidance here.
Intuitively I would add a schema_object parameter to `test_generate_json` in [test_generate.py](https://github.com/dottxt-ai/outlines/blob/36f1bf27a80f5877b88a915749341664cef4a1ab/tests/generate/test_generate.py#L255). But this test is skipped, because it fails for some models? Shouldn't we just skip those models instead of all? And by executing the test with all the different concrete models, doesn't that make it an integration test? If I'm right, how do you decide if a test goes to this file or to any of the test_integration_*.py files?

And should I remove the specific handling of the Enum type in favor of the TypeAdapter?

rlouf:
Is it still necessary now that we support most basic Python types?

RobinPicard:
I can't see a use case. Let's say that we don't add it for now and wait to see if someone has an example in which it's necessary.

PR #1311, opened 2024-12-02
Use `outlines-core` to translate JSON Schemas into regexes

rlouf:
Outlines is currently using its own logic to translate JSON Schema definitions into regular expressions. However, this logic was ported to `outlines-core` recently. In this PR I remove Outlines' implementation and import `outlines-core`'s instead.

### TODO

- [x] ~~It seems like `outlines-core` does not support the `enum` keyword when the elements are objects. Craft a MRE and open an issue there.~~ That's completely expected since `enum` is for constant values only. I fixed this by converting `enum` into `oneOf`.


rlouf:
Coverage will necessarily fail, they correspond to the MLX array conversion and we cannot test MLX on GH runners. Merging.

Issue #1270, opened 2024-11-17
Support Audio models


rlouf:
See for instance https://huggingface.co/fixie-ai/ultravox-v0_4_1-llama-3_1-8b

g-prz:
Hey @rlouf 
Interested in working on that one, I would propose an integration similar to `TransformersVision` and [Qwen2-Audio](https://huggingface.co/Qwen/Qwen2-Audio-7B-Instruct) seems to be a good candidate for testing. 
Finally, having a cookbook like the one for [vision models](https://github.com/dottxt-ai/outlines/blob/main/docs/cookbook/read-pdfs.md) could be a good example. 
Let me know if anything seems not aligned?

rlouf:
I'm all for it

PR #1317, opened 2024-12-03
Add support of audio transformers

g-prz:
This PR proposes an integration of audio transformers into outlines framework (#1270)


g-prz:
Hey @rlouf 🙋‍♂️ 
I'm not sure, but it seems that the runner disk is full and qwen2-audio files are taking a bit too much space, I can try to find a gguf version or maybe a smaller model? Wdyt?

rlouf:
We generally use HF's internal testing models which are much smaller, see the other tests. Can you try to find if there is one for audio?

g-prz:
Hey @rlouf 
I updated `test_generate_text`
I have 95% coverage, the missing parts are in case of import errors (not sure we want to modify the test env to have the coverage) or processor with no tokenizer, to what extend do you want to increase the coverage? 
Otherwise good for review on my side

g-prz:
Hey @rlouf updated here, still having no coverage if `transformers` is not installed an not sure we want to modify the test env to test these lines?

Issue #1335, opened 2024-12-12
Samplers should know know their own `SamplingParameters`


dgerlanc:
### What behavior of the library made you think about the improvement?

Since we already have classes for samplers, we should encapsulate the known behaviors of the samplers within that class.

### How would you like it to behave?

Currently, the `SamplingParameters` are processed in conditional statement at line 431 of `outlines/generate/api.py`

I think we should move the generation of the sampling parameters to the individual sampler classes.

The new code in `SequenceGeneratorAdapter` would be:

```python
def __init__(self, model, logits_processor, sampler):
    self.model = model
    self.logits_processor = logits_processor
    self.sampling_params = sampler.sampling_params()
```

And each of these classes would get a `sampling_parameters` property or method.

* MultinomialSampler
* GreedySampler
* BeamSearchSampler

This will allow us to add new samplers without modifying the code in `SequenceGeneratorAdapter`

kkpo4553:
Hello, we tried to solve the issue.

This is what we did:

Refactor the code to move the generation of sampling parameters to individual sampler classes. This change improves encapsulation and makes it easier to add new samplers without modifying the SequenceGeneratorAdapter.

You can review changes in this commit: https://github.com/kkpo4553/dottxt-ai-outlines-1335/commit/baf66bce0572e0bcc91b15a34a1c470e63ef5338.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

dgerlanc:
Resolved by #1340

PR #1340, opened 2024-12-15
Move generation of sampling params

sky-2002:
This PR addresses #1335 
I have kept the `SamplingParameters` in `outlines/generate/api.py`  cause its imported at many places and may break examples ig. 
Would you want to change that @dgerlanc ??
Otherwise if this looks good, I can make docs changes if needd, lmk.


rlouf:
LGTM, thank you for contributing!

PR #1349, opened 2024-12-20
[MLX] [bugfix] Preserve dtype of array when converting to torch

neilmehta24:
We have noticing the following error with a recent version of outlines when used with MLX:
```
TypeError: argument 'token_id': 'float' object cannot be interpreted as an integer

At:
  /.../outlines_core/fsm/guide.py(294): get_next_state
  /.../outlines/processors/structured.py(101): process_logits
  /.../outlines/processors/base_logits_processor.py(90): __call__
```

The issue is that the MLX array of tokens, which are integers, are being force-converted to floats, even though outlines expects an integer array. This is because all MLX arrays are being converted to `float32`, even when it's not necessarily appropriate, like in this case. Looking at the [commented link](https://ml-explore.github.io/mlx/build/html/usage/numpy.html#pytorch), the advice was to convert to `float32` only for `bfloat16`, because numpy does not support `bfloat16`. Now the MLX `_to_torch` implementation matches the other array libraries, none of the other libraries are being force-casted to float

### DIFF
--- a/outlines/processors/base_logits_processor.py
+++ b/outlines/processors/base_logits_processor.py
@@ -112,9 +112,7 @@ class OutlinesLogitsProcessor(Protocol):
             # https://ml-explore.github.io/mlx/build/html/usage/numpy.html#pytorch
             if tensor_like.dtype == mx.bfloat16:
                 tensor_like = tensor_like.astype(mx.float32)
-            return torch.from_dlpack(
-                np.array(tensor_like, copy=False)
-            )
+            return torch.from_dlpack(np.array(tensor_like, copy=False))
 
         elif is_jax_array_type(type(tensor_like)):
             import jax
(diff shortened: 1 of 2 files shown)


rlouf (approved):

neilmehta24:
@rlouf The CI configuration prevents me from reaching the coverage requirements. From the [CI logs](https://github.com/dottxt-ai/outlines/actions/runs/12457418146/job/34772164331#step:7:23):
```
outlines/processors/base_logits_processor.py (0.0%): Missing lines 113-115
```

But note that the [CI doesn't install](https://github.com/dottxt-ai/outlines/actions/runs/12457418146/job/34771948716#step:4:572) or run any MLX code, so to my understanding it's impossible to improve the CI coverage report when adding MLX-specific logic.

However, I did check locally that there isn't a regression in coverage, and that the tests are passing

Coverage:
Screenshot 2024-12-22 at 4 03 05 PM

Test:
Screenshot 2024-12-22 at 4 05 17 PM


So this should be good to go, please let me know what you think.

rlouf:
Yes, sorry for the delay in reviewing. Thank you for contributing this fix!

PR #1355, opened 2025-01-02
Add `from_file` class method to the `Prompt` object

### LINKED ISSUE #1345: Add `from_file` class method to the `Prompt` object
We keep coming back to the following pattern: first write the prompt in docstrings, and move it to a text file once it gets too large. It would be nice if Outlines made it easy for users to generate a prompt template from a text file. The file can contain plain text, or be written using the Jinja2 templating syntax.

I propose to add a `from_file` classmethod, which reads the file, extracts the names of the free variables (if any) in the template, and returns a `Prompt` instance.

yvan-sraka:
Fix #1345: For now, I'm just trying to make sure I'm editing the right part of the codebase. I haven't managed to run the tests yet…

<details><summary>EDIT: I opened a separate issue about the falling tests 🙂</summary>
<p>

I noticed that the project use `pytest`, so I assumed that running `pip install ".[test]"` would install all the dependencies needed to run the full test suite. However, it seems that's not the case _(it's currently complaining that I don't have `vllm`)_. Should I open an issue _(I can add the missing dependency under the `test` section of the `pyproject.toml`)_?

</p>
</details>

https://github.com/dottxt-ai/outlines/issues/1356


rlouf:
Yes, please open an issue 😄

outlines/prompts.py
```diff
@@ -27,20 +21,57 @@ def __post_init__(self):
...
+
+        # Find all undeclared variables including those from included files
+        variables = find_all_variables(template, loader)
+
+        # Create a signature with the variables as parameters
+        parameters = [
+            inspect.Parameter(var, inspect.Parameter.POSITIONAL_OR_KEYWORD)
+            for var in variables
+        ]
+        signature = inspect.Signature(parameters)
```
rlouf, line 71: Can we add a test for this?

tests/test_prompts.py
```diff
@@ -312,3 +314,68 @@ def args_prompt(fn):
         args_prompt(with_all)
         == "args: x1, y1, z1, x2: bool, y2: str, z2: Dict[int, List[str]], x3=True, y3='Hi', z3={4: ['I', 'love', 'outlines']}, x4: bool = True, y4: str = 'Hi', z4: Dict[int, List[str]] = {4: ['I', 'love', 'outlines']}"
     )
+
+
+@pytest.fixture
+def temp_prompt_files():
```
rlouf, line 320: Can we also add a test when [template inheritance](https://jinja.palletsprojects.com/en/stable/templates/#template-inheritance) is used?

rlouf:
Added a couple comments, although since this is a draft PR you are probably addressing them already.

tests/test_prompts.py
yvan-sraka: I'm indeed struggling a bit to support [template inheritance](https://jinja.palletsprojects.com/en/stable/templates/#template-inheritance) and [simple includes](https://jinja.palletsprojects.com/en/stable/templates/#include). I've finally figured out the overall design, and I think I'll need to tweak the current `Prompt` internals. For instance, `template` is currently a `str`, but it might be more more semantically correct for it to be a `jinja2.Template` instead _(but this would require some refactoring)_.

https://github.com/dottxt-ai/outlines/blob/d32dfde3ae7eac71f57af96d5141975ef9322b12/outlines/prompts.py#L13-L42

Alternatively, I could just try to quickly overload our `create_jinja_template` method to optionally accept a file path as alternative to a string containing the template content _(also, I decided not to allow include and inheritance to reference files that are outside the folder or subfolders of the file given to `from_file`, is this behavior intuitive?)_ WDYT? any opnion on what would be the best design?

https://github.com/dottxt-ai/outlines/blob/d32dfde3ae7eac71f57af96d5141975ef9322b12/outlines/prompts.py#L190-L219

rlouf: > _(also, I decided not to allow include and inheritance to reference files that are outside the folder or subfolders of the file given to `from_file`, is this behavior intuitive?)_ 

It is intuitive, and I think we should add a comment in the code (if you haven't already).

> WDYT? any opnion on what would be the best design?

I would say go for the cleanest solution, even if it requires some refactoring.

outlines/prompts.py
```diff
@@ -301,10 +324,10 @@ def get_schema_dict(model: Dict):
     return json.dumps(model, indent=2)
 
 
-@get_schema.register(type(BaseModel))
-def get_schema_pydantic(model: Type[BaseModel]):
+@get_schema.register(type(pydantic.BaseModel))
+def get_schema_pydantic(model: Type[pydantic.BaseModel]):
     """Return the schema of a Pydantic model."""
-    if not type(model) == type(BaseModel):
+    if not isinstance(model, type(pydantic.BaseModel)):
```
yvan-sraka, line 330: These changes are unrelated lint fixes, I can eventually move them to another PR!

outlines/prompts.py
```diff
@@ -34,12 +36,23 @@ def __call__(self, *args, **kwargs) -> str:
...
+            bound_arguments.apply_defaults()
+            return self.template.render(**bound_arguments.arguments)
+        else:
+            return self.template.render(**kwargs)
+
+    @classmethod
+    def from_file(cls, path: Path):
+        """Create a Prompt instance from a file containing a Jinja template.
 
-    def __str__(self):
```
yvan-sraka, line 41: I removed the `__str__` implementation from `Prompt` because it now keeps a `jinja2.Template` as its internal state. And, it's not straightforward to serialize a template made from more than one file...

That should be the only breaking change in this PR. If keeping this feature is important, I can work around it by adding a new internal variable to the class to store the`str` version of the template. And when `from_file` is used, it would return the content of the file passed as an argument, regardless of any other files it might depend on... However, I'm still unsure when the `__str__` method would be useful anyway, WDYT?

yvan-sraka:
Working on this PR made me wonder: **why do we integrate Jinja2** _(or any specific templating engine)_ into Outlines? I mean, we could have just provided a Prompt interface that takes an already rendered `str` _(and eventually keep a function decorator if that sounds fancy and helps better infer the template arguments)_, leaving the templating up to the user. That way, users could choose a different templating engine if they wanted, and e.g., how to use Outlines in combination with Jinja2 could be rather provided as an example in the documentation. WDYT?

outlines/prompts.py
rlouf: I'm fine with this breaking change

rlouf:
I am not sure what you mean by "taking an already render `str`", could you provide an example?

outlines/prompts.py
rlouf: We can keep them in this PR as long as they have their own commit.

outlines/prompts.py
```diff
@@ -183,17 +196,16 @@ def render(template: str, **values: Optional[Dict[str, Any]]) -> str:
     A string that contains the rendered template.
 
     """
-    jinja_template = create_jinja_template(template)
-    return jinja_template.render(**values)
+    return template.render(**values)
```
torymur, line 199: From interface point of view `render` functionality has changed too if defined this way. Before it would accept `str` and create jinja template (what's now done by `template_from_str`), why was it redefined to accept jinja template right away? 

Because I'm not sure of the overall usability of `render` if it accepts the jinja template. By reading the test cases and docs with examples it seems this function was here to confirm how the transformation from `str` to jinja template looked like. So, by calling it, user would make sure that it was transformed as expected.

outlines/prompts.py
```diff
@@ -19,12 +21,12 @@ class Prompt:
...
-    template: str
-    signature: inspect.Signature
+    template: jinja2.Template
+    signature: Optional[inspect.Signature]
 
     def __post_init__(self):
-        self.parameters: List[str] = list(self.signature.parameters.keys())
-        self.jinja_environment = create_jinja_template(self.template)
+        if self.signature is not None:
+            self.parameters: List[str] = list(self.signature.parameters.keys())
```
torymur, line 29: I didn't find where `self.parameters` is used in the logic, with the exception of tests, wonder if we can drop it? 🤔

rlouf:
You can already use any templating engine you want with Outlines, by rendering the template and passing the returned string to a generator:

```python
import outlines

model = outlines.models.openai("gpt-4o")

# Input

examples = [
    {"question": "What is the capital of France?", "answer": "Paris"},
    {"question": "What is 2 + 2?", "answer": "4"},
]
question = "What is the Earth's diameter?"


# Using Jinja2 ...
import jinja2

env = jinja2.Environment(loader=jinja2.FileSystemLoader("."))
template = env.get_template("prompt.jinja2")
rendered = template.render(examples=examples, question=question)

results = generator(rendered)


# ... or, Using Mako ...
import Mako

template = mako.template.Template(filename="prompt.mako")
rendered = template.render(examples=examples, question=question)

results = generator(rendered)
```

The role of the `Prompt` object is to wrap a Jinja2 template so that it returns a string when called with the template's variable values. The confusion may come from poor naming on my part; this object should probably be called `Template` instead.

outlines/prompts.py
torymur: Ideally, since now there are 2 separate flows to jinja template:
1) from str
2) from file
Then it's expected to call `render` with jinja template, yes, but this old kind of `render` more looks like `Prompt.from_str`, which will allow to confirm the str to jinja transformation

tests/test_prompts.py
```diff
@@ -135,7 +158,6 @@ def test_prompt_kwargs():
     def test_kwarg_tpl(var, other_var="other"):
         """{{var}} and {{other_var}}"""
 
-    assert test_kwarg_tpl.template == "{{var}} and {{other_var}}"
```
torymur, line 138: Why not keep these two `.template` checks?

yvan-sraka: Because `template` is no longer a `str` but a `jinja2.template`, I could try adapting them instead of suppressing them, however, they seemed a bit redundant to me, WDYT?

yvan-sraka:
Thank you very much @rlouf for the clarification. I’m not sure about the naming either, but if it's extensible, that already sounds like a great design! Apologies for sharing ideas maybe a bit too quickly as they pop from my mind, I probably should have checked the documentation first!

outlines/prompts.py
yvan-sraka: Yes, agreed! I'll check for usages and remove it :)

outlines/prompts.py
yvan-sraka: The old design was slightly inefficient because a `jinja2.template` should not be constructed more than once per template. Previously, it was being reconstructed for each call to `render`...

I understand this could be a potential breaking change. Thinking about it, I had to modify `render`-related tests, but I don't think users should be able to call `render` without a `Prompt` object, WDYT? And yes, it would make sense to expose a `from`_str method for parity with `from_file`!

yvan-sraka: I've updated the `render` code so it no longer breaks previous usage, but instead displays a deprecation warning. WDYT?

rlouf:
> Thank you very much @rlouf for the clarification. I’m not sure about the naming either, but if it's extensible, that already sounds like a great design! Apologies for sharing ideas maybe a bit too quickly as they pop from my mind, I probably should have checked the documentation first!

No worries! Questions are always welcome

tests/test_prompts.py
torymur: Yeah, probably type checks will be enough, though I would suggest adding assertions to verify `.signature`: check it to be present from decorator call `from_str` and None otherwise.

outlines/prompts.py
torymur: Yeah, I checked new code, `from_str` looks good!

Though I wonder about two things:
1) since this will be in major update, maybe we can get away even without deprecation warning for whatever we'll decide for `render` just this time
2) this `render` function still creates more questions of why it exists to me, you both in your examples used interface like `prompt.template.render(params)`, which is much clearer anyway and with `from_str` available, especially. Also, there is similar logic in `__call__`, so maybe this standalone `render` could be dropped all together? It seems it was something historical as well, considering graph mentioning: `Parse a Jinaj2 template and translate it into an Outlines graph.`

So, I would even remove `render` function and integrate and refactor its docs in for `template_from_str`.
@yvan-sraka @rlouf what's your take?

yvan-sraka: I thought about it, and tbh, I think that's a bit overkill. I’d be fine with not exposing it to the user, as well as other methods. However _(and maybe this is where my dusty understanding of Python module visibility falls short)_, AFAIU, those methods aren't exposed atm, e.g.:
```python
def render(
    template: Union[jinja2.Template, str], **values: Optional[Dict[str, Any]]
) -> str:
    r"""
    Examples
    --------

    Outlines follow Jinja2's syntax

    >>> import outlines
    >>> outline = outlines.render("I like {{food}} and {{sport}}", food="tomatoes", sport="tennis")
    I like tomatoes and tennis
    
    ... snipped"""
```
```
% python3 demo.py  
 File "/Users/yvan/outlines/demo.py", line 2, in <module>  
  outline = outlines.render("I like {{food}} and {{sport}}", food="tomatoes", sport="tennis")  
AttributeError: module 'outlines' has no attribute 'render'  
```
So, I guess it's the documentation that's rather misleading, right?

tests/test_prompts.py
yvan-sraka: I added a few more `assert` statements to check that the `signature`. It's not `None` only when the `Prompt` is constructed using the decorator _(otherwise, it can't infer the signature argument since there is no function)_.

outlines/prompts.py
yvan-sraka: I'm not sure that `template_from_str` should be user-facing. Why would the user use it, and how could that be more ergonomic than `Prompt.from_str`? What about renaming them to `__template_from_str` and `__template_from_file`, so it's more obvious that they are internals? Where is the documentation that we should not forget to update?

torymur: I was more thinking about integration of these docs/tests of `render` into tests of `template_from_str`, jjust to preserve these assumptions, but not to make `template_from_* ` user-facing. I agree, that these are internal, we can make it explicit, yes. 

I was also searching for docs, seems they're generated based on docstrings: https://dottxt-ai.github.io/outlines/latest/api/prompts/

yvan-sraka:
Help! I don't understand [the failed test coverage](https://github.com/dottxt-ai/outlines/actions/runs/12654790343/job/35264763034?pr=1355). It says that only [line 90 of `prompt.py`](https://github.com/dottxt-ai/outlines/pull/1355/commits/32587ee0e5c170c9f465c5ec4022329e62a2202e#diff-ff53e0881f88b197f0a25271f293a9eaab48666f2a05fdf6aafd386344bf6307R90) is not covered, which doesn't make sense to me…

torymur:
It seems that tests don't cover that part of the flow, where if is True and `cleaned_template += "\n"` is actually being executed:
```
        ends_with_linebreak = content.replace(" ", "").endswith("\n\n")
        if ends_with_linebreak:
            cleaned_template += "\n"
```

This is one of the cases, I suppose, where one of the old `render` tests would exactly cover (I would integrate them all into `template_from_str`):
```
    tpl = """
        A test line
            An indented line

    """
    assert render(tpl) == "A test line\n    An indented line\n"
```

torymur (approved):
Turned out nicely, looking good! 🚀

tests/test_prompts.py
```diff
@@ -145,7 +152,6 @@ def test_kwarg_tpl(var, other_var="other"):
     assert p == "test and kwarg"
 
     p = test_kwarg_tpl("test", "test")
-    assert p == "test and test"
```
yvan-sraka, line 148: I guess I removed this line by mistake

yvan-sraka:
> _(Do not merge this PR yet, I will rewrite the commit history a bit once we've settled on deprecating or removing `outlines.prompts.render`!)_

I've cleaned up the git history, so it's ready to be merged now :)

yvan-sraka:
> _(Do not merge this PR yet, I will rewrite the commit history a bit once we've settled on deprecating or removing `outlines.prompts.render`!)_

I've cleaned up the git history, so it's ready to be merged now :)

rlouf:
Great job! 🎉

rlouf (approved):

rlouf:
No :(

PR #1360, opened 2025-01-04
Fix `json_schema` imports in cot example

tylerjthomas9:
- Replace `outlines.integrations` import with import from `outlines.fsm.json_schema` (#1061) 
- Fix `build_regex_from_schema` to import from `outlines_core` instead of `outlines` (#1311)


rlouf:
Thank you!

PR #1364, opened 2025-01-08
Add `Outline`s

### LINKED ISSUE #1346: Add `Application`s
(name to be defined)

Following https://github.com/dottxt-ai/outlines/discussions/1226 and https://github.com/dottxt-ai/outlines/discussions/1220 we will be able to implement the following interface:

```python
from outlines import Application, models

def template(a: int) -> str:
    return f"What is 2 times {a}?"

fn = Application(model, template, int)
```

where fn is a callable object whose arguments are the arguments to the prompt template:

```python
fn(1)
# 2

fn(3)
# 6
```

yvan-sraka:
Fix #1346, WDYT about the integration with `outlines.prompt`? _(Here, the template argument is just a function that returns a `str`, so it could be either an `@outlines.prompt`-decorated function or a regular one)_


tests/test_outline.py
```diff
@@ -0,0 +1,56 @@
+from unittest.mock import MagicMock
+
+import pytest
+
+from outlines.outline import Outline
+
+
+def test_outline_int_output():
+    # Mock the model
+    model = MagicMock()
```
yvan-sraka, line 10: I thought that using a real model here might result in this being a flaky test, WDYT?

rlouf: Yes, good idea. Plus it would add a lot of overhead for no obvious improvement.

outlines/outline.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    def __init__(self, model, template, output_type):
+        self.model = model
+        self.template = template
+        self.output_type = output_type
+
+    def __call__(self, *args):
+        # Generate the prompt using the template function
+        prompt = self.template(*args)
```
rlouf, line 37: How should we support the corner case where users pass a string?

rlouf:
> Fix #1346, WDYT about the integration with `outlines.prompt`?

We should allow functions that are not decorated, such as

```python
def build_prompt(a: int) -> str:
    return f"What is {a} squared?"
```

Because we do not want to force the use of templates on users.

rlouf:
Also you can deprecate the code in `function.py` as this is going to replace it.

outlines/outline.py
```diff
@@ -0,0 +1,52 @@
+import ast
+from dataclasses import dataclass
+
+
+@dataclass
+class Outline:
+    """
+    Outline is a class that creates a callable object to generate responses
+    based on a given model and a prompt template.
```
rlouf, line 9: and an output type.

outlines/outline.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Args:
```
rlouf, line 11: We've used the Numpy docstring style so far and should stick to it.

outlines/outline.py
```diff
@@ -0,0 +1,52 @@
...
+        self.model = model
+        self.template = template
+        self.output_type = output_type
+
+    def __call__(self, *args):
+        # Generate the prompt using the template function
+        prompt = self.template(*args)
+
+        # Generate the response using the model
+        response = self.model.generate(prompt)
```
rlouf, line 40: This should use either `outlines.generate.json`, `outlines.generate.regex`, etc. The interface is necessarily going to be brittle for now. You can take a look at the v1 branch to see how this is going to be more robust with the undergoing refactor.

outlines/outline.py
yvan-sraka: How is that a corner case? I'll add a test to demonstrate that it already works.

yvan-sraka:
> > Fix #1346, WDYT about the integration with `outlines.prompt`?
> 
> We should allow functions that are not decorated, such as
> 
> ```python
> def build_prompt(a: int) -> str:
>     return f"What is {a} squared?"
> ```
> 
> Because we do not want to force the use of templates on users.

That's already how this PR works, I just wanted to confirm that this was the intended design and not an omission!

> Also you can deprecate the code in `function.py` as this is going to replace it.

Should I handle that in this PR or a separate one? By _deprecate_, do you mean adding a user-facing warning when it's used, rather than removing it entirely?

outlines/outline.py
yvan-sraka: Hmm, okay, but which one of those should I use by default? And how should I let users switch to the other one, by adding a fourth optional argument, e.g. `generation_kind`, WDYT?

outlines/outline.py
rlouf: My bad, I should have reviewed more carefully

rlouf:
> Should I handle that in this PR or a separate one? By _deprecate_, do you mean adding a user-facing warning when it's used, rather than removing it entirely?

We can do it in this PR. Let's add a user-facing warning, and announce that it will be removed in favour of the `Outline` interface in v1.0

outlines/outline.py
yvan-sraka: I just realized that maybe I misunderstood the PR, and it should infer the correct regex to match the user's required output type. E.g., if a user asks for an `int`, I should ask `outlines.generate.regex` something like `^[+-]?[0-9]+$`. IIUC, I'm unsure how to scale this to arbitrary return types. Should we define a list of supported return types? Using `ast.literal_eval` was the most flexible solution I came up with, but I'm not sure how well it handles custom data structures. In such cases, I think we should check that users provide types that implement deserialization from JSON. WDYT?

outlines/outline.py
yvan-sraka: What about having that checked by a pre-commit hook? https://github.com/dottxt-ai/outlines/pull/1368

outlines/outline.py
rlouf: Let's only accept Pydantic models / JSON Schema strings for now and use `outlines.generate.json`. We'll be able to easily generalise after we push the refactor in the v1 branch.

yvan-sraka: I've updated the PR _(and the usage example in the docstring)_. Does it still look like what we want?

outlines/function.py
```diff
@@ -10,6 +10,9 @@
     from outlines.generate.api import SequenceGenerator
     from outlines.prompts import Prompt
 
+# Raising a warning here caused all the tests to fail…
+print("The 'function' module is deprecated and will be removed in a future release.")
+
```
torymur, line 15: Maybe better place would be putting warning in `__post_init__` of `Function`?

outlines/outline.py
```diff
@@ -0,0 +1,61 @@
...
+        return f"What is 2 times {a}?"
+
+    fn = Outline(model, template, OutputModel)
+
+    result = fn(3)
+    print(result)  # Expected output: OutputModel(result=6)
+    """
+
+    def __init__(self, model, template, output_type):
+        if not (isinstance(output_type, str) or issubclass(output_type, BaseModel)):
```
torymur, line 44: This seems like an oversimplified check to confirm str to be a json schema string: `isinstance(output_type, str)`. Even though v1 will handle this much more gracefully, perhaps, even for a temp solution it worth to verify str being an actual json schema?

rlouf: It's surprisingly difficult to find a tool that validates a Json Schema. Would be very happy if you found one!

tests/test_outline.py
```diff
@@ -0,0 +1,41 @@
...
+
+    with patch("outlines.generate.json", return_value=mock_generator):
+        outline_instance = Outline(
+            mock_model,
+            template,
+            '{"type": "object", "properties": {"result": {"type": "integer"}}}',
+        )
+        result = outline_instance(3)
+
+    assert result["result"] == 6
```
torymur, line 41: Understanding that this is a draft yet, but still to mention it: tests for misc errors flows are missing

outlines/outline.py
```diff
@@ -0,0 +1,61 @@
...
+            raise TypeError(
+                "output_type must be a Pydantic model or a JSON Schema string"
+            )
+        self.template = template
+        self.output_type = output_type
+        self.generator = generate.json(model, output_type)
+
+    def __call__(self, *args):
+        prompt = self.template(*args)
```
torymur, line 53: Maybe worth to support **kwargs as well?

outlines/function.py
rlouf: Moving my previous comment here for visibility:

> I would be nice to specify which release (1.0.0) so people can pin their version to < this release?

outlines/function.py
```diff
@@ -20,13 +20,23 @@ class Function:
...
     """
 
     prompt_template: "Prompt"
     schema: Union[str, Callable, object]
     model_name: str
     generator: Optional["SequenceGenerator"] = None
 
+    def __post_init__(self):
+        print(
```
rlouf, line 36: We should favor warnings in general, unless there's a good reason not to do so (maybe there is here?). The reason is that if tests are properly set up in another codebase they will fail when a warning is raised, it's another way to warn about the change.

outlines/outline.py
```diff
@@ -0,0 +1,68 @@
...
+        self.output_type = output_type
+        self.generator = generate.json(model, output_type)
+
+    def __call__(self, *args, **kwargs):
+        prompt = self.template(*args, **kwargs)
+        response = self.generator(prompt)
+        try:
+            if isinstance(self.output_type, str):
+                return json.loads(response)
+            return self.output_type.model_validate_json(response)
```
rlouf, line 65: Let's instantiate the model with the JSON data, like we do [here](https://github.com/dottxt-ai/outlines/blob/03d387893b95f151c74ff6b7d7864e15cb484eee/outlines/generate/json.py#L52) to keep a consistent interface.

outlines/outline.py
```diff
@@ -0,0 +1,68 @@
...
+    fn = Outline(model, template, OutputModel)
+
+    result = fn(3)
+    print(result)  # Expected output: OutputModel(result=6)
+    """
+
+    def __init__(self, model, template, output_type):
+        if isinstance(output_type, str):
+            try:
+                jsonschema.Draft7Validator.check_schema(json.loads(output_type))
```
rlouf, line 47: Good find! Is there a test for this?

tests/test_outline.py
```diff
@@ -0,0 +1,82 @@
...
+def test_outline():
+    mock_model = Mock()
+    mock_generator = Mock()
+    mock_generator.return_value = '{"result": 6}'
+
+    with patch(
+        "outlines.generate.json", return_value=lambda *args, **kwargs: mock_generator()
+    ):
+        outline_instance = Outline(mock_model, template, OutputModel)
+        result = outline_instance(3)
```
rlouf, line 26: And add an assertion to make sure it's an instance of a Pydantic `BaseModel` once you make the change requested above.

outlines/outline.py
yvan-sraka: I got this linter warning:

> The method "parse_raw" in class "BaseModel" is deprecated
> The `parse_raw` method is deprecated; if your data is JSON use `model_validate_json`, otherwise load the data then use `model_validate` instead.Pylance

yvan-sraka:
I'm a bit frustrated by this [failing test](https://github.com/dottxt-ai/outlines/actions/runs/12751020237/job/35537074359#step:6:133), I can't reproduce it locally, and I'm not sure how to work around it :/

outlines/outline.py
rlouf: You're right, my bad

torymur:
Update the tests to not use deprecated `Function`?: https://github.com/dottxt-ai/outlines/blob/79100b293f861ee8bd74390527ec25cb6fe3b670/tests/test_function.py

torymur:
Also, found mention in quickstart docs, needs to be updated too: https://github.com/dottxt-ai/outlines/blob/79100b293f861ee8bd74390527ec25cb6fe3b670/docs/quickstart.md?plain=1#L186-L190

rlouf:
Can you run `pytest` with the `--pdb` flag?

yvan-sraka:
> Can you run `pytest` with the `--pdb` flag?

Thanks for the tips! I ran it inside the GitHub Action job and got the following traceback:

```python
outlines/fsm/guide.py:76: in cached_create_states_mapping
    return uncached_create_states_mapping(regex_string, tokenizer, *args, **kwargs)
/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/outlines_core/fsm/guide.py:141: in create_states_mapping
    return create_states_mapping_from_fsm(regex_fsm, tokenizer, frozen_tokens)
/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/outlines_core/fsm/guide.py:178: in create_states_mapping_from_fsm
    states_to_token_maps, empty_token_ids = create_fsm_index_tokenizer(
/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/outlines_core/fsm/regex.py:473: in create_fsm_index_tokenizer
    tokens_to_token_ids, empty_token_ids = reduced_vocabulary(tokenizer)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

tokenizer = <Mock name='mock.tokenizer' id='140081937198256'>

    @lru_cache
    def reduced_vocabulary(
        tokenizer,
    ) -> Tuple[Dict[str, List[int]], Set[int]]:
        """Create a map from decoded vocabulary tokens to lists of equivalent token ids."""
        # TODO FIXME: See if we can get the underlying Rust tokenizers from HF and
        # do all this in Rust
        empty_token_ids = set()
        vocabulary: Dict[str, List[int]] = {}
>       for token, token_idx in tokenizer.vocabulary.items():
E       TypeError: 'Mock' object is not iterable

/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/outlines_core/fsm/regex.py:394: TypeError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/outlines_core/fsm/regex.py(394)reduced_vocabulary()
-> for token, token_idx in tokenizer.vocabulary.items():
(Pdb) p tokenizer
<Mock name='mock.tokenizer' id='140081937198256'>
(Pdb) p dir(tokenizer)
['assert_any_call', 'assert_called', 'assert_called_once', 'assert_called_once_with', 'assert_called_with', 'assert_has_calls', 'assert_not_called', 'attach_mock', 'call_args', 'call_args_list', 'call_
count', 'called', 'configure_mock', 'method_calls', 'mock_add_spec', 'mock_calls', 'reset_mock', 'return_value', 'side_effect', 'vocabulary']
(Pdb) p tokenizer.vocabulary
<Mock name='mock.tokenizer.vocabulary' id='140081937198736'>
(Pdb) p tokenizer.vocabulary.items
<Mock name='mock.tokenizer.vocabulary.items' id='140081937198832'>
````

This made me realize that `from outlines_core.fsm.regex import reduced_vocabulary` is only used in `tests/generate/test_integration_transformers.py`, so the test might fail because of an update in `outline_core`, rather than an issue with the tests in `tests/test_outline.py`, WDYT?

**Comment by yvan-sraka** [comment]
[I upgraded my Mock game](https://github.com/dottxt-ai/outlines/pull/1364/files#diff-aa41978f733a456b8f0212bec5ce261a1d0fc2d31def5692def7934ef5eae099R9-R13), but I’m starting to believe I shouldn’t put too much effort into that and should focus on patching something else to work around it...

```
=========================== short test summary info ============================
FAILED tests/test_outline.py::test_outline - ValueError: The vocabulary does not allow us to build a sequence that matches the input
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
```

PR #1365, opened 2025-01-08
Add `model_name` attribute to `model` classes

### LINKED ISSUE #1361: Add `model_name` attribute to model classes
We will need to be able to pull a model name from a model class instance in the future. We should thus add a `model_name` attribute.

yvan-sraka:
Fix #1361, did I forget anything?


rlouf:
I meant the name of the pre-trained model. This won't be needed immediately, and will likely be refactored in v1, we can safely wait until then to avoid wasted efforts.

PR #1368, opened 2025-01-09
Add a few dotfiles to improve newcomers' experience

yvan-sraka:
WDYT about Astral’s tools, [Ruff](https://docs.astral.sh/ruff/) and [uv](https://docs.astral.sh/uv/)?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.yaml]
+indent_size = 2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,33 +1,20 @@
 repos:
 - repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: v4.4.0
+  rev: v5.0.0
   hooks:
     -   id: check-merge-conflict
     -   id: debug-statements
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
-- repo: https://github.com/pycqa/isort
-  rev: 5.12.0
-  hooks:
-    - id: isort
-      args: [--profile, black]
-- repo: https://github.com/asottile/pyupgrade
-  rev: v3.3.1
-  hooks:
-    - id: pyupgrade
-      args: [--py37-plus]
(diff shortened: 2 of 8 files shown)


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
rlouf, line 1: TIL. I like that it has native support in many editors.

.pre-commit-config.yaml
```diff
@@ -31,3 +31,13 @@ repos:
       args: [--allow-redefinition]
       exclude: ^examples/
       additional_dependencies: [types-tqdm, types-Pillow]
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: v0.8.6
```
rlouf, line 35: Would we need to remove `flake8`, `isort` and `black`? The documentation says it has drop-in parity.

rlouf:
I don't have anything against `uv` _a priori_ although last time I tried it installing PyTorch was a headache. We would need to make sure everything works out of the box on every supported platform, and document the decision if we were to go this route.

yvan-sraka:
Yeah, my concern is, iiuc, that for now, we distribute the library without any kind of lock file. I will advocate for at least adding an `uv.lock` _(or similar)_ to ensure the ability to reproduce a developer environment. Also, if I end up adding support for Nix, I would likely use [`uv2nix`](https://pyproject-nix.github.io/uv2nix/usage/hello-world.html).

rlouf:
Let's open a discussion to weigh pros and cons of `uv`. `ruff` is raising some interesting style issues already, I'm not opposed to using it.

rlouf:
Overall this looks good to me, and before fixing the code style check I want to understand why ruff asks for explicit re-exports here before merging. An issue on the `ruff` repository refers to [PEP484](https://peps.python.org/pep-0484/) but I am still not sure how this is related.

yvan-sraka:
> Overall this looks good to me, and before fixing the code style check I want to understand why ruff asks for explicit re-exports here before merging. An issue on the `ruff` repository refers to [PEP484](https://peps.python.org/pep-0484/) but I am still not sure how this is related.

I’ve read a bit about this online, but I’m not sure to have any strong opinion. That said, it seems like ignoring certain lints _([as vllm does](https://github.com/vllm-project/vllm/blob/main/pyproject.toml#L51-L61))_ while fixing others is a reasonable strategy, WDYT about ignoring F401 but addressing E721 and E731?

torymur (approved):
LGTM and ruff lints to ignore seems reasonable 👍

PR #1371, opened 2025-01-10
add filters to prompt function

derfred:
Allow giving custom filters to the prompt decorator

```
def reverses: str) -> str:
    return s[::-1]

@prompt(filters={ 'reverse': reverse })
def reverse_prompt(text):
    '''{{ text | reverse }}'''

prompt = reverse_prompt("Hello")

print(prompt)
>>> "olleH"
```

### DIFF
--- a/outlines/prompts.py
+++ b/outlines/prompts.py
@@ -6,7 +6,7 @@ import re
 import textwrap
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Type, cast
+from typing import Any, Callable, List, Dict, Optional, Type, cast, Union
 
 import jinja2
 import pydantic
@@ -40,7 +40,7 @@ class Prompt:
             return self.template.render(**kwargs)
 
     @classmethod
-    def from_str(cls, content: str):
+    def from_str(cls, content: str, **filters: Dict[str, Callable]):
         """
         Create an instance of the class from a string.
 
@@ -53,10 +53,10 @@ class Prompt:
         -------
         An instance of the class with the provided content as a template.
         """
-        return cls(cls._template_from_str(content), None)
+        return cls(cls._template_from_str(content, **filters), None)
 
     @classmethod
-    def from_file(cls, path: Path):
+    def from_file(cls, path: Path, **filters: Dict[str, Callable]):
         """
         Create a Prompt instance from a file containing a Jinja template.
 
@@ -75,10 +75,10 @@ class Prompt:
         """
         # We don't use a `Signature` here because it seems not feasible to infer one from a Jinja2 environment that is
         # split across multiple files (since e.g. we support features like Jinja2 includes and template inheritance)
-        return cls(cls._template_from_file(path), None)
+        return cls(cls._template_from_file(path, **filters), None)
 
     @classmethod
-    def _template_from_str(_, content: str) -> jinja2.Template:
+    def _template_from_str(_, content: str, **filters: Dict[str, Callable]) -> jinja2.Template:
         # Dedent, and remove extra linebreak
         cleaned_template = inspect.cleandoc(content)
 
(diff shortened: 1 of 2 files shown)


rlouf:
Thank you for contributing! I assume the reason you proposed this particular design is because the existing filters are "baked in" and cannot be modified by the user?

I'd like to explore other designs before committing to this one. For instance adding `filters` as an attribute of the `Prompt` class:

```python
@dataclass
class Prompt:
    template: jinja.Template
    signature: Optional[inspect.signature]
    filters: dict = field(default_factory = lambda : {
    "name": get_fn_name,
    "description": get_fn_description
})
```

So you would write:

```python
def reverse_string(s: str) -> str:
    return s[::-1]

@prompt
def reverse_prompt(text):
    '''{{ text | reverse }}'''

reverse_prompt.filter["reverse"] = reverse_string
prompt = reverse_prompt("Hello")
```

Which is arguably not ideal, and makes more sense when used as:

```
template = Prompt.from_str("{{ text | reverse }})
template.filters["reverse"] = reverse_string
```

This is clearly a design smell, let's try to figure this out!

outlines/prompts.py
```diff
@@ -118,10 +125,14 @@ def _template_from_file(_, path: Path) -> jinja2.Template:
...
         )
+
+        for name, filter_fn in filters.items():
+            env.filters[name] = filter_fn
+
         return env.get_template(os.path.basename(path))
 
 
-def prompt(fn: Callable) -> Prompt:
+def prompt(fn: Optional[Callable] = None, **filters: Dict[str, Callable]) -> Callable:
```
rlouf, line 135: `filters` should be a simple argument that has type `Dict[str, Callable]` and maps the filter names to the functions.

outlines/prompts.py
```diff
@@ -118,10 +125,14 @@ def _template_from_file(_, path: Path) -> jinja2.Template:
             keep_trailing_newline=True,
             undefined=jinja2.StrictUndefined,
         )
+
+        for name, filter_fn in filters.items():
```
rlouf, line 129: I generally don't mind a little repetition but I think we should refactor the code and add a `create_environment` class that can be used in both class methods. @yvan-sraka are there any reasons we *shouldn't* do that?

derfred:
thanks for your review @rlouf. I've taken your comments into account. While doing so the following alternative approach came to me:

```
env = outlines.prompts.prompt_environment([reverse])

@env
def my_prompt(text):
  "reversed: {{ text | reverse }}"
```

what do you think?

rlouf:
Thank you for being attentive to the general design! I think your original design to be simpler, we should go with it (including my comments).

derfred:
ok, great! Do you have any more comments about the implementation?

outlines/prompts.py
yvan-sraka: I agree, ideally, it would be less surprising for users to have the most similar environment (and behavior) for any method used to create a `Prompt`. Otherwise, it's a great feature addition!

outlines/prompts.py
yvan-sraka: I agree, ideally, it would be less surprising for users to have the most similar environment (and behavior) for any method used to create a `Prompt`. Otherwise, it's a great feature addition!

outlines/prompts.py
```diff
@@ -166,11 +199,22 @@ def prompt(fn: Callable) -> Prompt:
...
 
     return Prompt(template, signature)
 
 
+def _add_filters(
+    env: jinja2.Environment, filters: Union[List[Callable], Dict[str, Callable]]
+):
+    if isinstance(filters, list):
+        for filter_fn in filters:
+            env.filters[filter_fn.__name__] = filter_fn
```
rlouf, line 212: Let's stick to `filters` being a dictionary.

derfred: I removed the list option.

rlouf (approved):

rlouf:
Looks good to me! @yvan-sraka ?

yvan-sraka (approved):
**LGTM**! Great contribution, thanks :)

derfred:
Awesome!

Issue #1347, opened 2024-12-19
Example is underwhelming


TimZaman:
### Describe the issue as clearly as possible:

I used `README.md`'s example, and changed the review to "Review: This restaurant is bad!" -instead of 'good'- or any similar negative phrasing. The answer given back is almost exclusively "Positive".

I understand this is the model's behaviour, but maybe either:
1. this model isn't up to this task
2. the prompt is too hard for this example.
 
**Suggestion:** We could reconsider a more illustrative example

### Steps/code to reproduce the bug:

```python
import outlines

model = outlines.models.transformers("microsoft/Phi-3-mini-4k-instruct")

prompt = """You are a sentiment-labelling assistant.
Is the following review positive or negative?

Review: This restaurant is bad!
"""

generator = outlines.generate.choice(model, ["Positive", "Negative"])
answer = generator(prompt)
print(f'{answer=}')
```

### Expected result:

```shell
"Negative"
```

### Error message:

```shell
"Positive"
```

### Outlines/Python version information:

outlines v 0.1.11
Python 3.10.14 | packaged by conda-forge | (main, Mar 20 2024, 12:51:49) [Clang 16.0.6 ]

### Context for the issue:

first time user trying out the example

denadai2:
what happens without outlines?

cpfiffer:
This might be a result of not using the proper chat templating in the example. This works better:

```python
import outlines
from transformers import AutoTokenizer

model_name = "microsoft/Phi-3-mini-4k-instruct"
tokenizer = AutoTokenizer.from_pretrained(model_name)
model = outlines.models.transformers(model_name)

def template(prompt: str) -> str:
    return tokenizer.apply_chat_template(
        [{"role": "user", "content": prompt}],
        tokenize=False,
        add_bos=True,
        add_generation_prompt=True,
    )

prompt = """You are a sentiment-labelling assistant.
Is the following review positive or negative?

Review: This restaurant is bad!
"""

generator = outlines.generate.choice(model, ["Positive", "Negative"])
answer = generator(template(prompt))
print(f'{answer=}')
```

cpfiffer:
This is a general issue with the documentation and in outlines use more generally. We have a few outstanding issues for this, like #987, #756,

There's a PR #1019 which might address this, though currently it seems to be in limbo.

PR #1372, opened 2025-01-10
Update README to include chat templating

cpfiffer:
The existing README has underwhelming or incorrect results (Example is underwhelming #1347) due to lack of templating for instruct models.

This adds special tokens for each instruct model call, as well as provide comments on how to obtain/produce special tokens.


README.md
```diff
@@ -107,7 +129,7 @@ generator = outlines.generate.choice(model, Sentiment)
 answer = generator(prompt)
 ````
```
torymur, line 130: Seems this part also needs to be adjusted, but maybe we can show only the difference:

````python
from enum import Enum

class Food(str, Enum):
      pizza = "Pizza"
      pasta = "Pasta"
      salad = "Salad"
      dessert = "Dessert"

...

generator = outlines.generate.choice(model, Food)
````

File: README.md [hunk 2]
```diff
@@ -65,23 +65,45 @@ is to ensure that there is a well-defined interface between their output and
...
+<|im_end|>
+<|im_start|>assistant
 """
 
-generator = outlines.generate.choice(model, ["Positive", "Negative"])
+generator = outlines.generate.choice(model, ["Pizza", "Pasta", "Salad", "Dessert"])
 answer = generator(prompt)
+print(f'{answer=}')
+
+# answer=Pizza
```
torymur, line 106: From our perspective any from the list is acceptable, so maybe leave llm some wiggle room in case it would decide to be funny, so something like?:

```suggestion
answer = generator(prompt)

# Likely answer: Pizza
```

yvan-sraka:
[Rendered](https://github.com/cpfiffer/outlines/tree/readme-update?tab=readme-ov-file#structured-generation)

README.md
cpfiffer: IMO every example in a README should be completely copy-able with no slice-and-dice on the user's part. This is of course personal preference, so up to y'all.

cpfiffer:
Great, appreciate the comments!

README.md
yvan-sraka: If one day we generate documentation with `mdbook`, it offers a feature for [hiding code lines](https://rust-lang.github.io/mdBook/format/mdbook.html#hiding-code-lines) from the user. I realized there isn't a similar feature in GitHub Flavored Markdown _(yet?)_. The closest thing I'm aware of is [collapsed sections](https://github.com/github/docs/blob/main/content/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections.md)...

yvan-sraka (approved):
This overall **LGTM**. I like that we provide a lot of examples in the `README.md` and understand why the Positive/Negative one felt counterintuitive... That said, I found most of the examples a bit abstract/useless and would prefer more concrete, real-world use cases!

The example that probably fits this category best is character generation, though it would be even more interesting if one of the character generation fields was an unconstrained _(or less constrained)_ string, something like a description field _(eventually with a limit, e.g., less than 500 characters of lore)!_ Otherwise, I think it could be nice to hide internet culture jokes in example because, hey, we’re _cool kids!_ Ofc, these comments are for future improvement, the PR is already great and ready to be merged as is!

README.md
```diff
@@ -386,6 +413,39 @@ prompt = labelling("Just awesome", examples)
...
+yields
+
+```
+<|im_start|>system
+You extract information from text.<|im_end|>
+<|im_start|>user
+What food does the following text describe?<|im_end|>
+<|im_start|>assistant
+```
+
```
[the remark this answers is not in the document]
cpfiffer: Agreed. I'm not sure what the best place is for it. We could add it to the prompting reference page or a new page. Not sure which is better. It wouldn't be too hard to cook up a new page for the chat templating issue, as we can provide lots of little bits of context.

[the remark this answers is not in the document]
torymur: My 5 cents here is that since chat templating issue kind of grows all over the examples and at the same time has nothing to do with structure-generating showcases it could be properly stated in the description before all the sections, and considering how many nuances could be there, probably, deserves its own page with a reference from the description.

cpfiffer:
I moved the chat template tips to the top of the README, and added a specific page in the docs for it.

torymur (approved):
LGTM!

Issue #88, opened 2023-05-02
Add `length` constraint to the generation


rlouf:
This should apply to the _word_, not the tokens.

PR #1378, opened 2025-01-15
Add length limit (maximum number of words) in generation

g-prz:
This PR aims at proposing a limit of words that are generated, similarly to a max number of tokens (#88).


g-prz:
@rlouf seems ready for review

g-prz:
@rlouf WDYT?

g-prz:
@rlouf WDYT?

PR #1382, opened 2025-01-17
Fix readme to reflect routing through enum

farnasirim:


rlouf (approved):

rlouf:
Good catch, thank you!

PR #1384, opened 2025-01-18
Add support for OpenAI style function call support

punitvara:
User should pass OpenAI style dictionary of function and it should be internally converted to same json schema as callable function


RobinPicard (commented):
It looks fine to me, but could you do the following please?
- Update the type-hinting and docstrings in the function to include it as a possible input type
- Update the documentation in `docs/reference/generation/json.md` to specify that this schema_object type is available, explain the necessary dict structure and link to the Openai documentation about it
- Write a test for this new type in `tests/generate/test_generate.py`

punitvara:
yes sure. @RobinPicard will update PR.

Issue #1359, opened 2025-01-03
Call models with a prompt and output type directly


rlouf:
In the current re-design of the API we [added `Generator` objects](https://github.com/dottxt-ai/outlines/blob/v1.0/outlines/generate/__init__.py) that accept a model and an output type as parameters:

```python
from outlines import Generator, models
from pydantic import BaseModel

class Foo(BaseModel):
    bar: str

model = models.openai("gpt-4o")
generator = Generator(model, Foo)
result = generator("prompt")
```

The rationale behind this design is that, for open source models, the index for structured generation can take some time to compile and we hold it in this object so it can be re-used. There are a few issues with this design:

1. API models do not have this problem
2. This adds an extra step in the happy `pip install outlines -> text generated path`
3. It does not correspond to most people's mental model. Other libraries make you call the model instance with the prompt to generate text.

I suggest that we modify the original design to make simple use cases straightforward, but still allows flexibility in more complex cases. We can add a `__call__` method to model instances so the previous workflow can be re-written as:

```python
from outlines import Generator, models
from pydantic import BaseModel

class Foo(BaseModel):
    bar: str

model = models.openai("gpt-4o")
result = model("prompt", Foo)
```

While keeping the `Generator` pattern for cases where a user may want to call the same model repeatedly with the same structure:

```python
from outlines import Generator, models
from pydantic import BaseModel

class Foo(BaseModel):
    bar: str

model = models.openai("gpt-4o")
generator = Generator(model, Foo)

result_1 = generator("prompt_1")
result_2 = generator("prompt_2")
```

RobinPicard:
Your suggestion makes a lot of sense to me. I think that removing the generation creation step by having the possibility to call the model directly makes the basic example look a lot simpler and more intuitive to new users.

PR #1385, opened 2025-01-20
Create a base class for the models

RobinPicard:
Updated on 2025-02-11

The objective of this PR is to create a base class all models would inherit from. This base class defines the interface for models and includes a `__call__` method that allows models to be called with a prompt and output type directly following the suggestion in #1359 .

Formatting methods for the input and output types are implemented in a separate class inheriting from the base class `ModelTypeAdapter`. Each model must set a `type_adapter` attribute in its initialization. The rationale for separating those from the model class is that the accepted types for each model and the formatting logic are quite different concerns compared to the specifics of the model's implementation. Separating them also allows us to have an easier to read documentation of the accepted types for each model.

Another change proposed in this PR is to set a mandatory `model_type` attribute in each model class. This would allow us not to have to rely on a static list of models through the `APIModel` and `LocalModel` classes.

The use of the base model class is implemented for the models that have already been refactored: OpenAI, Gemini, Anthropic and Llamacpp.

### DIFF
--- a/outlines/models/__init__.py
+++ b/outlines/models/__init__.py
@@ -9,6 +9,7 @@ codebase.
 from typing import Union
 
 from .anthropic import Anthropic
+from .base import Model, ModelTypeAdapter
 from .exllamav2 import ExLlamaV2Model, exl2
 from .gemini import Gemini
 from .llamacpp import LlamaCpp
--- a/outlines/models/base.py
+++ b/outlines/models/base.py
@@ -1,0 +1,82 @@
+from abc import ABC, abstractmethod
+
+
+class ModelTypeAdapter(ABC):
+    """Base class for all model type adapters.
+
+    A type adapter instance must be given as a value to the `type_adapter`
+    attribute when instantiating a model.
+    The type adapter is responsible for formatting the input and output types
+    passed to the model to match the specific format expected by the
+    associated model.
+
+    """
+
+    @abstractmethod
+    def format_input(self, model_input):
+        """Format the user input to the expected format of the model.
+
+        For API-based models, it typically means creating the `messages`
+        argument passed to the client. For local models, it can mean casting
+        the input from str to list for instance.
+        This method is also used to validate that the input type provided by
+        the user is supported by the model.
+
+        """
+        ...
+
+    @abstractmethod
+    def format_output_type(self, output_type):
+        """Format the output type to the expected format of the model.
+
+        For API-based models, this typically means creating a `response_format`
+        argument. For local models, it means formatting the logits processor to
+        create the object type expected by the model.
+
+        """
+        ...
+
+
+class Model(ABC):
+    """Base class for all models.
+
+    This class defines a shared `__call__` method that can be used to call the
+    model directly.
+    All models inheriting from this class must define a `type_adapter`
+    attribute of type `ModelTypeAdapter`. The methods of the `type_adapter`
+    attribute are used in the `generate` method to format the input and output
+    types received by the model.
+
+    """
+
+    type_adapter: ModelTypeAdapter
+
+    def __call__(self, model_input, output_type=None, **inference_kwargs):
+        """Call the model.
+
+        Users can call the model directly, in which case we will create a
+        generator instance with the output type provided and call it.
+        Thus, those commands are equivalent:
+        ```python
+        generator = Generator(model, Foo)
+        generator("prompt")
+        ```
+        and
+        ```python
+        model("prompt", Foo)
+        ```
+
+        """
+        from outlines.generate import Generator
+
+        return Generator(self, output_type)(model_input, **inference_kwargs)
+
+    @abstractmethod
+    def generate(self, model_input, output_type=None, **inference_kwargs):
+        """Generate a response from the model.
+
+        The output_type argument contains a logits processor for local models
+        while it contains a type (Json, Enum...) for the API-based models.
+
+        """
+        ...
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -1,7 +1,9 @@
 import pickle
 import warnings
+from functools import singledispatchmethod
 from typing import TYPE_CHECKING, Dict, Iterator, List, Set, Tuple, Union
 
+from outlines.models.base import Model, ModelTypeAdapter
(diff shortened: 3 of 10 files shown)


rlouf:
That's smart. I am not a big fan of mixins, however, and I wonder if we should instead define at least one base class for all models, and also require that `format_input` and `format_output_type` are defined? That's a little too many levels of inheritance to my liking, but we can figure this out later (`OpenAIBase` does not need to be a class, for instance)

outlines/models/base.py
```diff
@@ -0,0 +1,24 @@
...
+    @abstractmethod
+    def format_input(self, model_input):
+        ...
+
+    @abstractmethod
+    def format_output_type(self, output_type):
+        ...
+
+
+class Model(ABC):
```
rlouf, line 16: What's the rationale for having two different classes here?

RobinPicard:
I agree that mixins aren't great. I like the idea of having a base class for all models as it would allow, on top of avoiding repetitions, to define/document the interface of the models somewhere in a very straightforward manner. If we want to keep the formatting logic outside of the model classes not to make them have too many methods, but not rely on inheritance either, maybe a solution could be to require the models to have a formatter attribute that would inherit from a base ModelFormatter class.

outlines/models/base.py
RobinPicard: The rationale was to keep the separation currently found in branch v1.0 between OpenAIBase (preparing the arguments for the model's main methods) and OpenAI (where the main methods are).

outlines/models/base.py
```diff
@@ -0,0 +1,24 @@
...
+    """Base class for all models."""
```
rlouf, line 17: Explain why we are creating this class in the docstring!

outlines/models/base.py
```diff
@@ -0,0 +1,24 @@
+from abc import ABC, abstractmethod
+
+
+class ModelFormatter(ABC):
```
rlouf, line 4: I would call it `TypeAdapter` or something like this. Wdyt?

rlouf: Works for me

cpfiffer:
I don't have too much to add from a technical perspective, but I like the single inheritance format. Much easier to extend.

I poked around and it doesn't seem like this has much of an impact on most Outlines users, right? Only people deep into the internals are going to be impacted, and even then minorly.

outlines/models/__init__.py
```diff
@@ -19,6 +20,3 @@
 from .vllm import VLLM, vllm
 
 LogitsGenerator = Union[Transformers, LlamaCpp, OpenAI, ExLlamaV2Model, MLXLM, VLLM]
-
-LocalModel = LlamaCpp
-APIModel = Union[AzureOpenAI, OpenAI, Anthropic, Gemini]
```
rlouf, line 24: I would keep these types instead of using a class attribute to distinguish between API and local models.

rlouf:
A few nitpicks, looks good otherwise. To fix the CI:

- We need to add `datasets` as a dependency;
- We need to skip tests marked as `api_call` in the regular CI

outlines/models/llamacpp.py
```diff
@@ -134,7 +180,7 @@ def from_pretrained(cls, repo_id, filename, **kwargs):
             model = Llama.from_pretrained(repo_id, filename, **kwargs)
             return cls(model)
 
-    def generate(self, prompt: str, logits_processor, **inference_kwargs) -> str:
+    def generate(self, model_input, logits_processor, **inference_kwargs):
```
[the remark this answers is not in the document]
RobinPicard: I removed it because I had a mypy error for inconsistent argument name/type with the abstract method of the parent class. Should we just remove the abstract method `generate` of the base class to give more freedom to the model implementations' `generate` methods?

outlines/generate/__init__.py
```diff
@@ -70,10 +70,10 @@ def __call__(self, prompt, **inference_kwargs):
 
 
 def Generator(
-    model: Union[LocalModel, APIModel],
+    model: Model,
     output_type: Optional[Union[Json, List, Choice, Regex]] = None,
 ):
-    if isinstance(model, APIModel):  # type: ignore
+    if model.model_type == "api":  # type: ignore
```
yvan-sraka, line 76: The `# type: ignore` mypy instruction seems no longer required, right? The two dataclasses start to look pretty similar, is the strategy to merge them into one?

outlines/models/base.py
```diff
@@ -0,0 +1,84 @@
...
+    This class defines a shared `__call__` method that can be used to call the
+    model directly.
+    All models inheriting from this class must define a `type_adapter`
+    attribute of type `ModelTypeAdapter`. The methods of the `type_adapter`
+    attribute are used in the `generate` method to format the input and output
+    types received by the model.
+
+    """
+
+    model_type: Union[Literal["api"], Literal["local"]]
```
yvan-sraka, line 53: TIL about `typing.Litteral` in Python, nice! Are they like _"atoms"_ in Scheme, and a way to build kind of enums?

yvan-sraka (approved):
Add a few superficial comments, nice work!

outlines/generate/__init__.py
RobinPicard: Yes, but I'm going to revert that change in the end anyway.

outlines/models/base.py
RobinPicard: Yes! But they are going away as we're going to keep the `LocalModel` and `ApiModel` types

rlouf: `Union[Literal["api"], Literal["local"]]` is equivalent to `Literal["api", "local"]` fyi

outlines/generate/__init__.py
```diff
@@ -73,7 +73,7 @@ def Generator(
     model: Union[LocalModel, APIModel],
     output_type: Optional[Union[Json, List, Choice, Regex]] = None,
 ):
-    if isinstance(model, APIModel):  # type: ignore
+    if isinstance(model, LocalModel):  # type: ignore
```
rlouf, line 76: The opposite, no?

RobinPicard: Ah yes, sorry

rlouf (approved):

Issue #1380, opened 2025-01-17
Prepare `outlines` to use `outlines-core` v0.2


torymur:
`outlines-core` v0.2 introduces major interface changes, which require additional work on the `outlines` side.

Currently in `outlines` we're importing these from `outlines_core`:

```python
from outlines_core.fsm.guide import Generate
from outlines_core.fsm.guide import Guide as CoreGuide
from outlines_core.fsm.guide import RegexGuide as CoreRegexGuide
from outlines_core.fsm.guide import Write
from outlines_core.fsm.guide import (
    create_states_mapping as uncached_create_states_mapping,
)
```

Now we're providing interfaces and usage like:

``` python
import json

from outlines_core.json_schema import build_regex_from_schema
from outlines_core.fsm.guide import Guide, Index, Vocabulary

schema =  {
  "title": "Foo",
  "type": "object",
  "properties": {"date": {"type": "string", "format": "date"}}
}
regex = build_regex_from_schema(json.dumps(schema))

vocabulary = Vocabulary.from_pretrained("openai-community/gpt2")
index = Index(regex, vocabulary)
guide = Guide(index)

# Get current state of the Guide:
current_state = guide.get_state()

# Get allowed tokens for the current state of the Guide:
allowed_tokens = guide.get_tokens()

# Advance Guide to the next state via some token_id and return allowed tokens for that new state:
next_allowed_tokens = guide.advance(allowed_tokens[-1])

# To check if Guide is finished:
guide.is_finished()

# If it's finished then this assertion holds:
assert guide.get_tokens() == [vocabulary.get_eos_token_id()]
```
More interface details for `Vocabulary`, `Index` and `Guide` will be available at: https://github.com/dottxt-ai/outlines-core/blob/main/python/outlines_core/outlines_core_rs.pyi

Some previously imported structures like `Write` and `Generate` might make sense to move back to `outlines` and adapt them, if they'll be used in other Guides. How best to proceed there is open for discussion.

PR #1386, opened 2025-01-22
Bump `outlines-core` to version `0.2.3`

yvan-sraka:
This is a very early half-baked draft aimed at addressing #1380, Currently, I’m struggling to make it work, so far, my development environment is just replacing the Python interfaces in `outlines-core` with those from the `fsm` subfolder:

```
outlines-core on git main [xr!+?] is pkg v0.0.0 via py v3.12.8 via rs v1.84.0  
[OCI] > git status  
On branch main  
Your branch is up to date with 'outline-core/main'.  
Changes to be committed:  
  (use "git restore --staged <file>..." to unstage)  
        modified:   python/outlines_core/__init__.py  
        deleted:    python/outlines_core/fsm/__init__.py  
        renamed:    python/outlines_core/fsm/json_schema.py -> python/outlines_core/json_schema.py  
        renamed:    python/outlines_core/fsm/outlines_core_rs.pyi -> python/outlines_core/outlines_core_rs.pyi
```

I also edited my outlines Conda environment to use the local version of outlines-core:

```diff
diff --git a/environment.yml b/environment.yml  
index 6fc980f..6553e74 100644  
--- a/environment.yml  
+++ b/environment.yml  
@@ -20,4 +20,4 @@ dependencies:  
   - transformers  
   - pip  
   - pip:  
-    - -e ".[test]"  
+    - -e ".[test]" -e ../outlines_core  
diff --git a/pyproject.toml b/pyproject.toml  
index 28060f7..42da6b0 100644  
--- a/pyproject.toml  
+++ b/pyproject.toml  
@@ -40,7 +40,7 @@ dependencies = [  
    "pycountry",  
    "airportsdata",  
    "torch",  
-   "outlines_core==0.1.26",  
+   "outlines_core",  
 ]  
 dynamic = ["version"]
```

Overall, I try to update the `outlines.fsm.guide` module to match the logic in the example from the issue. Are we planning to keep supporting all the `Guide` variants? For example, IIUC, `outlines-core` no longer provides `RegexGuide`, right? So I integrated some code from https://github.com/dottxt-ai/outlines-core/blob/0.1.27/python/outlines_core/fsm/guide.py, as per @torymur’s advice.

### DIFF
--- a/outlines/fsm/guide.py
+++ b/outlines/fsm/guide.py
@@ -1,30 +1,77 @@
 import collections
 import copy
 import warnings
-from typing import TYPE_CHECKING, Any, Generator, Union
-
-import torch
-from lark.indenter import DedentError
-from lark.lexer import UnexpectedCharacters, UnexpectedToken
-from outlines_core.fsm.guide import Generate
-from outlines_core.fsm.guide import Guide as CoreGuide
-from outlines_core.fsm.guide import RegexGuide as CoreRegexGuide
-from outlines_core.fsm.guide import Write
-from outlines_core.fsm.guide import (
-    create_states_mapping as uncached_create_states_mapping,
+from dataclasses import dataclass
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Generator,
+    Union,
+    List,
+    Optional,
+    Callable,
+    Tuple,
+    Set,
+    Dict,
+    FrozenSet,
+    Sequence,
+    cast,
 )
 
+import interegular
+import torch
+from interegular.fsm import (
+    Alphabet,
+    FSM,
+    anything_else,
+    _AnythingElseCls,
+    TransitionKey,
+    State,
+)
+from lark.indenter import DedentError
+from lark.lexer import UnexpectedCharacters, UnexpectedToken
+from outlines_core import Guide as CoreGuide, Index, Vocabulary
+from outlines_core.json_schema import build_regex_from_schema
 from outlines import grammars
-from outlines.fsm.parsing import PartialLark, PartialParserState
+from outlines.fsm.parsing import PartialLark, PartialParserState, make_deterministic_fsm
 
 if TYPE_CHECKING:
     from outlines.models.tokenizer import Tokenizer
 
 
+@dataclass(frozen=True)
+class Write:
+    """Write instruction.
+
+    Attributes
+    ----------
+    tokens
+        The sequence of tokens to be added to the current sequence by the
+        generation process.
+
+    """
+
+    tokens: List[int]
+
+
+@dataclass(frozen=True)
+class Generate:
+    """Generate instruction
+
+    Attributes
+    ----------
+    tokens
+        The tokens that lead to a valid completion if generated.  A value
+        of ``None`` indicates that all tokens are allowed.
+    """
+
+    tokens: Optional[List[int]]
+
+
 Instruction = Union[Write, Generate]
 
 
-class Guide(CoreGuide):
+class Guide:
     """Base definition of a generation guide.
 
     A generation guide defines the behavior of a finite-state machine that guides
@@ -34,7 +81,38 @@ class Guide(CoreGuide):
 
     """
 
-    initial_state: Any
+    def __init__(self, index: Index):
+        self._core_guide = CoreGuide(index)
+        self.initial_state = self._core_guide.get_state()
+
+    def get_next_instruction(self, state: Any) -> Instruction:
+        """Retrieve the list of allowed tokens for the given state."""
+        return self._core_guide.get_tokens()  # type: ignore
+
+    def get_next_state(self, state: Any, token_id: int) -> Any:
+        """Advance the guide's state with a token and return the new state."""
+        self._core_guide.advance(token_id)
+        return self._core_guide.get_state()
+
+    def is_final_state(self, state: Any) -> bool:
+        """Check if the given state is a final state."""
+        return self._core_guide.is_finished()
+
+    def copy(self) -> "Guide":
+        """Create a copy of the Guide."""
+        new_guide = Guide(self._core_guide.index)  # Assuming index is accessible
+        return new_guide
+
+    def __repr__(self) -> str:
+        return repr(self._core_guide)
+
+    def __str__(self) -> str:
+        return str(self._core_guide)
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, Guide):
+            return False
+        return self._core_guide == other._core_guide
 
 
 class StopAtEOSGuide(Guide):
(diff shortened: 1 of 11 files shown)


outlines/fsm/guide.py
```diff
@@ -21,9 +14,6 @@
     from outlines.models.tokenizer import Tokenizer
 
 
-Instruction = Union[Write, Generate]
```
torymur, line 24: I wouldn't drop this as of now:
- it's a hint of what instruction actually is and what it means in outlines
- it's very easy to make a mistake in returning expected instructions in `get_next_instruction` 
And looking at the changes you're already breaking these expectations.

I'd suggest moving them from `outlines-core` as they are, at this point it's unclear if it's worth redefining them (probably not at all, since it's already a challenge to adjust new functionality to the current expectations in `outlines`): https://github.com/dottxt-ai/outlines-core/blob/0.1.27/python/outlines_core/fsm/guide.py

```
@dataclass(frozen=True)
class Write:
    """Write instruction.

    Attributes
    ----------
    tokens
        The sequence of tokens to be added to the current sequence by the
        generation process.

    """

    tokens: List[int]


@dataclass(frozen=True)
class Generate:
    """Generate instruction

    Attributes
    ----------
    tokens
        The tokens that lead to a valid completion if generated.  A value
        of ``None`` indicates that all tokens are allowed.
    """

    tokens: Optional[List[int]]


Instruction = Union[Write, Generate]```

File: outlines/fsm/guide.py [hunk 2]
```diff
@@ -147,14 +117,14 @@ def get_next_instruction(self, state: CFGState) -> Instruction:
...
+            return self.eos_token_id
 
         valid_tokens = list(
             self.iter_valid_token_ids(state, self.tokenizer.vocabulary.values())
         )
         if len(valid_tokens) == 1:
-            return Write(torch.tensor(valid_tokens))
-        return Generate(torch.tensor(valid_tokens))
+            return valid_tokens[0]
+        return None
```
torymur, line 127: These are examples of broken instructions now, I think we need to harshly limit changes and do not propagate it further in the code's logic, by asking for `get_next_instruction` consumers to adjust to new types of instructions. If it expects a tensor here, it should get one. And we need to try to limit the scope to the `RegexGuide`. At least until we see what's what. We might create a plan for improving Guide's logics in following PRs once we confirm that it could be.

outlines/fsm/guide.py
```diff
@@ -274,3 +247,60 @@ def must_terminate_state(self, state: CFGState) -> bool:
...
+
+
+def create_states_mapping_from_fsm(fsm, tokenizer):
+    states_mapping = {}
+    for state in fsm.states:
+        states_mapping[state] = {
+            "allowed_token_ids": list(fsm.transitions[state].keys()),
+            "is_final": fsm.is_final(state),
+        }
+    return states_mapping
```
torymur, line 280: We would want to avoid these kind of "supporters" in the new `RegexGuide` and figure out the way to provide a better interface for it in Guide from `outlines-core`. So, in the experiments here, we need to try using only what's provided to understand what's missing and add it. Here, I'm not sure why do you even need to have `self.states_mapping`?

outlines/fsm/guide.py
```diff
@@ -274,3 +247,60 @@ def must_terminate_state(self, state: CFGState) -> bool:
...
+
+
+class RegexGuide(Guide):
+    def __init__(self, regex, tokenizer):
+        self.regex = regex
+        self.tokenizer = tokenizer
+        self.eos_token_id = tokenizer.eos_token_id
+        self.fsm = build_regex_from_schema(regex)
```
torymur, line 288: `build_regex_from_schema(regex)` build regex from JSON schema, not fsm

outlines/fsm/guide.py
```diff
@@ -274,3 +247,60 @@ def must_terminate_state(self, state: CFGState) -> bool:
...
+        self.states_mapping = create_states_mapping_from_fsm(self.fsm, tokenizer)
+        self.initial_state = 0
```
torymur, line 290: I would suggest to start with the least amount of attributes possible and then iterate on what's actually needed.

torymur:
> It doesn't yet use the new Index and Vocabulary interfaces. I'll have to make it look more like this?

Yes, you will need `Index` and `Vocabulary` interfaces. But `GuideWrapper` in your example doesn't look very helpful, it just unnecessarily repeats the existing logic. `RegexGuide` is already a wrapper of the Guide just for `outlines` context, it should adhere to defined protocols of `RegexGuide` for now and that's it.

In other words, I would suggest to start with that removed `RegexGuide` from `outlines-core` and method by method try to fulfill all the requirements `outlines` asks from it with new interfaces:
https://github.com/dottxt-ai/outlines-core/blob/0.1.27/python/outlines_core/fsm/guide.py

Then we can iterate over what could be missing in the `outlines_core` interfaces if we see some struggling in the `RegexGuide`'s logic. Or not, if it's just a backward compatibility thing, for example, then it should be in `outlines` only. _We might generally improve on guides interfaces in outlines, but I suspect, that our hands are already full for now._

For example, you can start with something very basic like this and then see where it leads you, method by method:

```python

class RegexGuide(Guide):
    """Guide to generate text in the language of a regular expression."""

    def __init__(self, guide, eos_tensor):
        # EOS tensor is used in instructions, so better to be created just once
        self.eos_tensor = eos_tensor
        self._guide = guide

    @classmethod
    def from_regex(cls, regex, tokenizer):
        # Tokenizer has something like `name_or_path` to get to the model's name
        vocabulary = Vocabulary.from_pretrained(tokenizer.name_or_path())
        # Unclear yet if we need attr access to Index/Vocabulary, you will see it when it develops
        index = Index(regex, vocabulary)
        guide = Guide(index)

        eos_tensor = torch.tensor([vocabulary.eos_token_id()])
        return cls(guide, eos_tensor)
```

torymur:
It was finally published to pypi, please try using `0.2.2` version of `outlines-core`.

yvan-sraka:
```
==================================== ERRORS ====================================
_________________ ERROR collecting tests/fsm/test_cfg_guide.py _________________
ImportError while importing test module '/home/runner/work/outlines/outlines/tests/fsm/test_cfg_guide.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/fsm/test_cfg_guide.py:7: in <module>
    from outlines import grammars, models
outlines/__init__.py:3: in <module>
    import outlines.generate
outlines/generate/__init__.py:3: in <module>
    from .choice import choice
outlines/generate/choice.py:7: in <module>
    from outlines_core.fsm.json_schema import build_regex_from_schema
E   ModuleNotFoundError: No module named 'outlines_core.fsm'
=========================== short test summary info ============================
ERROR tests/fsm/test_cfg_guide.py
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 2.98s ===============================
```

rlouf:
You can also import `Guide` as `from outlines_core import Guide`

outlines/fsm/guide.py
```diff
@@ -274,3 +307,21 @@ def must_terminate_state(self, state: CFGState) -> bool:
     def copy(self) -> "CFGGuide":
         """Create a copy of the Guide."""
         return CFGGuide(self.cfg_string, self.tokenizer)
+
+
+class BetterFSM:
+    def __init__(self):
+        self.finals = None
+        self.flat_transition_map = None
```
rlouf, line 315: Do we still need this?

torymur: No we shouldn't need this

yvan-sraka: Should I remove everything that's `fsm`-related?

outlines/fsm/parsing.py
```diff
@@ -623,7 +619,7 @@ def match(self, text, pos, last_fsm_state_seq: Optional[Tuple[int, ...]] = None)
             text_part,
         )
 
-        state_seq = walk_fsm(
+        state_seq = walk_fsm(  # type: ignore  # noqa: F821
```
yvan-sraka, line 622: I'm not sure if I should remove the whole `PartialScanner`, or workaround that?

rlouf: We need to keep everything that is related to grammars for now

rlouf:
We have several choices here: expose the necessary functions in `outlines-core`, copy the old Python code back, or wait until we decide what we're doing with the current CFG code. I would choose (1) or (3), with a preference for (3) in the current context.

rlouf:
Let's close for now. We'll come back to it after v1.0 is released.

PR #1387, opened 2025-01-22
Reproducible Python virtual environment using `uv` and `nix`

yvan-sraka:
This PR is a draft aimed at producing a reproducible Python virtual environment. `uv` seems like the right tool for this because it has a lock file mechanism _(but we could alternatively use `poetry`)_, and `nix` would be my preferred choice for managing system dependencies _(I’m not aware of many alternatives for pinning system dependencies, apart from possibly using a container with a pinned version of Alpine Linux?)_

This is still a draft, the `shell.nix` is currently too convoluted _(and ultimately I would like to rather rely on [uv2nix](https://github.com/dottxt-ai/outlines/pull/1368#issuecomment-2586566084))_, and the goal of this PR is to enable the use of this environment in GitHub Action jobs! Using this setup in CI would help avoid non-reproducible issues, such as [this one](https://github.com/dottxt-ai/outlines/pull/1364#issuecomment-2598353952).

**n.b.** that VLLM is currently [broken on NixOS](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/development/python-modules/vllm/default.nix#L180), debugging its derivation might be necessary at some point: `nix-build  /nix/var/nix/profiles/per-user/root/channels/nixos/ --run-env -A python3Packages.vllm`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,59 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1738136902,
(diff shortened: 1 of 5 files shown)


rlouf (approved):

rlouf:
@yvan-sraka I see this was marked as a draft. Is it in a mergeable state?

yvan-sraka:
> @yvan-sraka I see this was marked as a draft. Is it in a mergeable state?

You can merge it! I would have preferred to test it by [making sure the tests pass first](https://github.com/dottxt-ai/outlines/issues/1404). I guess fixing the tests may require bumping dependencies, but that could be done in another PR. Also, I'm not 100% sure the Nix expression isn't missing anything since the tests didn't pass...

[The sibling PR in `outlines-core` is finished](https://github.com/dottxt-ai/outlines-core/pull/141), the tests pass, and the next step could be to integrate it into CI!

maresb:
Why was pre-commit added as a runtime dependency? This seems wrong.

maresb:
Ah, this was also mentioned in https://github.com/dottxt-ai/outlines/issues/1465

Issue #1383, opened 2025-01-18
Dynamic schema creation


cpfiffer:
I recently noticed [GenSON](https://github.com/wolverdude/GenSON), a library for generating JSON schemas dynamically. 

I think we should evaluate ways of providing tools for dynamic JSON schema generation, possibly by simply showcasing how to use GenSON or a similar library. If anyone has other ideas, they're very welcome!

Typically, the way we specify schemas in Outlines is to use Pydantic, like so:

```python
class CalendarEvent(BaseModel):
    description: str
    date: datetime
```

For well structured programs, this is a __great idea__. Providing users a simple, clean interface in Pydantic enforces good practice. Users are required to provide a Pydantic model that can be used in a type-safe way everywhere in their application.

However, there are many cases where this can be problematic. Pydantic can make it difficult to program with Outlines when the schema must be modified in-place.

My simplest example is this:

```python
class CalendarEvent(BaseModel):
    id: str  # Note the addition of the id field
    description: str
    date: datetime
```

This ID is stored in my database, or uniquely generated before the model generates the object.

When you give this to the model, it will make up an ID that may not be unique. What I would like to do instead is:

```python
class CalendarEvent(BaseModel):
    id: Literal['abc']
    description: str
    date: datetime
```

This will _force_ the model to use the ID I provide, and I won't have to do any post-generation clean up to enforce a unique UUID. 

This is difficult to do, currently. I wrote an example of how to dynamically create Pydantic models, but it is quite clunky and does not have a convenient user interface. I've included an example of this in the detail block at the end of the issue.

## Other examples

Here's a few other cases where dynamic schema creation might be a useful user interface feature.

1. **Function calling**. Currently we do gnarly regular expressions, or set up mega-function calling objects. Here we can flexibly define functions that the model may choose from during runtime.
2. **General runtime usage**. I often run into cases where I need to change the schema in standard control flow, such as changing enums within larger classes.
3. **Flexibility**. Working with Pydantic dynamically is a pain in general. Pydantic is great when you have a fixed structure, but often you may wish to provide flexibly schemas conditional on a model response. Imaging that two disconnected systems send each other JSON -- if you build a schema from the message Alice sends to Bob, Bob can just replicate that schema and kick it back to Alice in a format Alice understands.
4. **Simplicity**. You don't always need internal Python objects that you get from Pydantic. I would often be happy with just a dict for throwaways, especially when I don't want to have a gigantic models/ directory packed with tiny Pydantic classes.

[...]

PR #1390, opened 2025-01-22
Add `genson` integration for json generation

g-prz:
This PR aims at integrating support of the `genson` package (in `generate.json`) to be able to use dynamic json schema generation as proposed in #1383.

### DIFF
--- a/docs/reference/generation/json.md
+++ b/docs/reference/generation/json.md
@@ -103,3 +103,76 @@ print(add(**result))
 ```
 
 A great advantage of passing functions directly to specify the structure is that the structure of the LLM will change with the function's definition. No need to change the code at several places!
+
+
+## From a dynamic JSON schema builder - GenSON
+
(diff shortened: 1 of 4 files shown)


rlouf:
Thank you! We need to add this to the documentation, and make it clear that the user needs to call `generate.json` again each time the schema is updated, or this will lead to confusion.

rlouf:
The change looks good to me!

> Here it is ! However genson behavior can be a bit unstable regarding the [required fields](https://github.com/wolverdude/GenSON/issues/78#issuecomment-2607570441), the user has to be careful with the use of that lib

Would you mind adding an [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) to warn users, and link to GenSON's documentation?

g-prz:
Done, also added the same example (and another link to the docs)

rlouf:
LGTM! Thank you for contributing!

PR #1403, opened 2025-02-05
Add a DSL to generate regular expressions

### LINKED ISSUE #1302: Add Regex DSL
## Why?

Regular expressions are a very compact DSL to generate DFAs, and can be intimidating at first. We cannot expect users to be fluent in this DSL.

## How?

We can make this easier by designing a simple Python DSL that compiles into regular expressions. It can be as simple as defining functions that return a regex string:

```python
def one_or_more(pattern: str):
    return f"({pattern})+"
```

This is in practice slightly more complex, as some characters need to be escaped in literal expressions (such as `(`), and we cannot expect users who are not familiar with regular expressions to know that. A solution would be to create a `RegexStr` object that abstracts this away, for instance:

```python
def escape(string):
    if string == "(":
        return f"\{string}"
    if string == ")":
        return f"\{string}"
    else:
        return string


class RegexStr(str):
    def __add__(self, other):
        if isinstance(other, RegexStr):
            return RegexStr(f"{self}{other}")
        else:
            return RegexStr(f"{self}{escape(other)}")

    def __radd__(self, other):
        if isinstance(other, RegexStr):
            return RegexStr(f"{other}{self}")
        else:
            return RegexStr(f"{escape(other)}{self}")

    def __repr__(self):
        # return regex in string format
        pass


def exactly(quantity, *regex_strs):
    regex_str = ''.join(regex_strs)
    if regex_str == '':
        raise ValueError('regex_strs argument must have at least one nonblank value')
    return RegexStr(regex_str + '{' + str(quantity) + '}')

print("(" + exactly(3, "abc") + ")")
```

This is however a very rough design, and I am open to alternatives.

rlouf:
This PR introduces a DSL to simplify the use of regular expressions in Outlines.

## Why a DSL

Not everyone is, and wants to be, familiar with regular expressions. However, regex-based structured generation [can be very powerful](https://blog.dottxt.co/performance-gsm8k.html), sometimes even more so than JSON Schema-based structured generation. This PR thus introduces new abstractions that allow users to create regular expressions from very simple and intuitive primitives.

## Requirements

Besides providing an intuitive DSL we want:

- The ability to use the regular expressions thus defined as types in Pydantic models;
- Pre-defined types for commonly used constructs:
    - `sentence`
    - `digit`
    - `paragraph`
    - And other types.
- A few custom types that can be useful: `email`, `phone_number`, etc.
- Make it easy to debug / iterate on the regular expression

## Features

The DSL works by building a graph from several kinds of nodes, or types. The generated graph can be inspected, used to match strings of text, used as a type in a pydantic `BaseModel`, or converted to a regular expression.

### Basic types

There is only one object that the user should ever have to manipulate, `Regex`.  While we also define a `String` type, we try to automatically convert Python strings to `String` objects whenever possible. For instance:

```python
from outlines.types import regex

temp = "The answer is " + regex("[0-9]")
repr(temp)
# Sequence(terms=[String(value="The answer is "), Regex(pattern="[0-9]")])

temp = "Word" | regex("[0-9]")
repr(temp)
# Alternatives(terms=[String(value="Word"), Regex(pattern="[0-9]")])
```

#### Optional

The `Optional` node corresponds to the `(...)?` pattern in regular expression. It can be instantiated directly with a factory function:

```python
from outlines.types import regex, optional

temp = optional(regex("[0-9]"))
repr(temp)
# Optional(term=Regex(pattern="[0-9]"))
```

We also provide the following method:

```python
from outlines.types import regex

temp = regex("[0-9]").optional()
repr(temp)
# Optional(term=Regex(pattern="[0-9]"))
```

#### KleneeStar

The Kleene star is an operator (`(...)*`) that means "zero or more". We provide a factory function and method to generate `KleeneStar` objects:

```python
from outlines.types import regex, zero_or_more

temp = zero_or_more(regex("[0-9]"))
repr(temp)
# KleeneStar(term=Regex(pattern="[0-9]"))

temp = regex("[0-9]").zero_or_more()
repr(temp)
# KleeneStar(term=Regex(pattern="[0-9]"))
```

#### KleenePlus

The Kleene plus is an operator (`(...)+`) that means "one or more". We provide a factory function and method to generate `KleenePlus` objects:

```python
from outlines.types import regex, one_or_more

temp = one_or_more(regex("[0-9]"))
repr(temp)
# KleenePlus(term=Regex(pattern="[0-9]"))

temp = regex("[0-9]").one_or_more()
repr(temp)
# KleenePlus(term=Regex(pattern="[0-9]"))
```

#### Sequence

`Sequence` objects represent sequences of sub-patterns, and loosely is equivalent to an "and" operator. We overload the `__add__` and `__radd__` methods of the `Term` base object to generate sequences. Strings are automatically converted to a `String` object:

```python
from outlines.types import regex

temp = "digit: " + regex("[0-9]")
repr(temp)
# Sequence(terms=[String(value="digit :"), Regex(pattern="[0-9]")])
```

#### Alternatives

`Altenatives` objects are equivalent to "or" operators. We overload the `__or__` and `__ror__` methods of the `Term` base object to generate alternatives. Strings are automatically converted to a `String` object:

```python
from outlines.types import regex

temp = "word" | regex("[0-9]")
repr(temp)
# Sequence(terms=[String(value="digit :"), Regex(pattern="[0-9]")])
```

#### QuantifyExact

`QuantifyExact` objects are used to specify the number of repetitions of a sub-pattern. We provide a `repeat` function and `times` method to create these objects:

```python
from outlines.types import regex, repeat

digits = repeat(regex("[0-9]"), 2, 2)
repr(digits)
# QuantifyExact(term=Regex(pattern="[0-9]"), 2)

digits = regex("[0-9]").times(2)
repr(digits)
# QuantifyExact(term=Regex(pattern="[0-9]"), 2)
```

#### QuantifyMininum

`QuantifyMinimum` objects are used to define repetitions of a sub-pattern, with a minimum of repetitions. We provide a `repeat` function and `repeat` method to create these objects:

```python
from outlines.types import regex, repeat

digits = repeat(regex("[0-9]"), 2, None)
repr(digits)
# QuantifyMinimum(term=Regex(pattern="[0-9]"), 2)

digits = regex("[0-9]").repeat(2, None)
repr(digits)
# QuantifyMinimum(term=Regex(pattern="[0-9]"), 2)
```

#### QuantifyMaximum

`QuantifyMinimum` objects are used to define repetitions of a sub-pattern, with a maximum of repetitions. We provide a `repeat` function and `repeat` method to create these objects:

```python
from outlines.types import regex, repeat

digits = repeat(regex("[0-9]"), None, 2)
repr(digits)
# QuantifyMaximum(term=Regex(pattern="[0-9]"), 2)

digits = regex("[0-9]").repeat(None, 2)
repr(digits)
# QuantifyMaximum(term=Regex(pattern="[0-9]"), 2)
```

#### QuantifyBetween

`QuantifyBetween` objects are used to define a bounded number of repetitions of a sub-pattern. We provide a `repeat` function and `repeat` method to create these objects:

```python
from outlines.types import regex, repeat

digits = repeat(regex("[0-9]"), 1, 2)
repr(digits)
# QuantifyBetween(term=Regex(pattern="[0-9]"), 1, 2)

digits = regex("[0-9]").repeat(None, 2)
repr(digits)
# QuantifyBetween(term=Regex(pattern="[0-9]"), 1, 2)
```

### Pre-defined types

We provide pre-defined types based on regular expressions, and that can be composed to define a more complex expression:

- `sentence`
- `paragraph`
- `quoted_text`
- `char`
- `integer`
- `newline`
- (...)

```python
from outlines.types import paragraph

document = "A document with 3 paragraphs:" + newline + paragraph.times(3) + newline + "The end."
```

### Generating JSON

It is possible to use a Pydantic model or JSON Schema to append JSON to the template:

```python
from outlines.types import json_schema, paragraph, newline


class User(BaseModel):
    name: str
    age: int


template = "Think hard and generate JSON." + newline + paragraph.times(4) + newline + json_schema(User)
```

### Pydantic integration

The expressions can be used as types in Pydantic objects, and are translated to a `string` type with the `pattern` keyword when converted to JSON Schema:

```python
from outlines.types import Regex, sentence

digit = Regex("[0-9]")
answer_template = "So the answer is " + digit

class GSM8K(BaseModel):
    reasoning: sentence.times(4)
    answer: answer_template
```

### Debugging

The DSL offers two ways to test and debug the generated expressions. The first one is to be able to test whether the expression matches a given string:

```python
from outlines.types import regex

digit = regex("[0-9]")
answer_template = "So the answer is " + digit

does_match = answer_template.matches("So the answer is 1")
print(does_match)
# True

does_match = answer_template.matches("So the answer is one")
print(does_match)
# False
```

The second one is pretty-printing of the graph generated:

```python
from outlines.types.dsl import String, Regex, KleeneStart

a = String("a")
b = String("b")
c = Regex("[0-9]")
d = KleeneStar(a | b) + c

print(d)
#└── Sequence
#    ├── KleeneStar(*)
#    │   └── Alternatives(|)
#    │       ├── String('a')
#    │       └── String('b')
#    └── Regex('[0-9]')
```

## Todo

- [x] Add more basic types: sentence, pargraph, quoted text
- [x] Move the python types defined in `fsm` to `types`
- [x] Add `json` factory to convert a Pydantic model / json schema string to a regular expression
- [x] Add more tests
  - [x] Regexes
  - [x] Factory functions
  - [x] Custom types
- [x] Add documentation
- [x] Clean the imports

Closes #1302.


cpfiffer:
I don't really have feature/style notes to add here. It looks solid to me.

outlines/types/dsl.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+Term.one_or_more = one_or_more
+Term.optional = optional
+Term.repeat = repeat
+Term.times = times
+Term.zero_or_more = zero_or_more
+
+
+def to_regex(term: Term) -> str:
```
414owen, line 324: Just want to draw attention to a few terms that will need extra sets of parentheses:

```python
>>> to_regex(QuantifyExact(String("test"), 3))
'test{3}'
>>> to_regex(Sequence([Regex("dog|cat"), String("fish")]))
'dog|catfish'
```

I realise this PR is still in draft though :)

outlines/types/dsl.py
```diff
@@ -0,0 +1,359 @@
...
+    def __radd__(self: "Term", other: Union[str, "Term"]) -> "Sequence":
+        if isinstance(other, str):
+            other = String(other)
+
+        return Sequence([other, self])
+
+    def __or__(self: "Term", other: Union[str, "Term"]) -> "Alternatives":
+        if isinstance(other, str):
+            other = String(other)
+
```
RobinPicard, line 48: Same comment as above but with `Alternatives`

outlines/types/__init__.py
```diff
@@ -1,4 +1,31 @@
...
+char = Regex(r"\w")
+newline = Regex(r"(\r\n|\r|\n)")  # Matched new lines on Linux, Windows & MacOS
+whitespace = Regex(r"\s")
+
+# Document-specific types
+sentence = Regex(r"[A-Z].*\s*[.!?]")
+paragraph = Regex(rf"{sentence.pattern}(?:\s+{sentence.pattern})*\n+")
+
+
+# Custom types
```
yvan-sraka, line 25: Would be nice to have specification, references or sources that tell where does complex Regexes come from!

outlines/types/airports.py
```diff
@@ -6,6 +6,4 @@
 AIRPORT_IATA_LIST = [
     (v["iata"], v["iata"]) for v in airportsdata.load().values() if v["iata"]
 ]
-
```
yvan-sraka, line 9: Seems like an unrelated code format change.

outlines/types/__init__.py
yvan-sraka: I infer that `email` have been taken from Stack Overflow https://stackoverflow.com/a/201378/14773537

yvan-sraka: IDEM, that would be nice to not lose that comment about ISBN:

>  Matches any ISBN number.  Note that this is not completely correct as not all
> 10 or 13 digits numbers are valid ISBNs. See https://en.wikipedia.org/wiki/ISBN
> Taken from O'Reilly's Regular Expression Cookbook:
> https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s13.html
> TODO: Can this be represented by a grammar or do we need semantic checks?

tests/types/test_custom_types.py
```diff
@@ -0,0 +1,111 @@
...
+    [
+        (types.locale.us.phone_number, "12", False),
+        (types.locale.us.phone_number, "(123) 123-1234", True),
+        (types.locale.us.phone_number, "123-123-1234", True),
+        (types.locale.us.zip_code, "12", False),
+        (types.locale.us.zip_code, "12345", True),
+        (types.locale.us.zip_code, "12345-1234", True),
+        (types.isbn, "ISBN 0-1-2-3-4-5", False),
+        (types.isbn, "ISBN 978-0-596-52068-7", True),
+        # (types.ISBN, "ISBN 978-0-596-52068-1", True), wrong check digit
```
yvan-sraka, line 21: Why is that test case commented out?

yvan-sraka (approved):
LGTM, add a few minor comments! The [documentation page](https://github.com/rlouf/outlines/blob/regex-dsl/docs/reference/regex_dsl.md) is great :)

outlines/types/dsl.py
```diff
@@ -0,0 +1,359 @@
...
+    >>> class User(BaseModel):
+    >>>     name: str
+    >>>     age: age_type
+
+    """
+
+    def __add__(self: "Term", other: Union[str, "Term"]) -> "Sequence":
+        if isinstance(other, str):
+            other = String(other)
+
```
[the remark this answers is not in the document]
rlouf: Very good catch! Proper testing would have prevented that 😔

rlouf:
Interesting case of "tests pass on my machine"

PR #1410, opened 2025-02-11
Replaced `pycountry` with `iso3166` for Better Licensing Compliance

aman-17:
`pycountry` is GPL-licensed, which can be restrictive depending on how the project is used. `iso3166` is Apache 2.0 licensed, which is way more permissive and open-source friendly. This keeps things simple and avoids any potential licensing headaches down the road. Switching to `iso3166` keeps the project legally compliant and makes it more flexible for both open-source and enterprise use. No unnecessary barriers, just clean and hassle-free licensing.


jakep-allenai:
This would be helpful for us at AI2. Can we get it merged in please?

jakep-allenai:
Thank you!

PR #1412, opened 2025-02-12
Create the CFG type to be used as an output type

RobinPicard:
This PR addresses issue #1300 

We create a new type `CFG` and update the `Generator` class to accept the `CFG` type (creating a `CFGLogitsProcessor`) on top of accepting regex-based types (creating a `RegexLogitsProcessor`)

### DIFF
--- a/outlines/generate/__init__.py
+++ b/outlines/generate/__init__.py
@@ -1,8 +1,9 @@
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Optional, Union, cast, get_args
+from typing import Any, Optional, Union, cast, get_args
 
-from outlines.models import APIModel, LocalModel
-from outlines.types import Choice, Json, List, Regex
+from outlines.models import APIModel, LlamaCpp, LocalModel
+from outlines.processors import CFGLogitsProcessor, RegexLogitsProcessor
+from outlines.types import CFG, Choice, Json, List, Regex
 
 from .api import SequenceGenerator
 from .cfg import cfg
@@ -12,9 +13,6 @@ from .fsm import fsm
 from .json import json
 from .regex import regex
 from .text import text
-
-if TYPE_CHECKING:
-    from outlines.processors import RegexLogitsProcessor
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


outlines/generate/__init__.py
```diff
@@ -60,18 +64,28 @@ def __post_init__(self):
         if self.output_type is None:
             self.logits_processor = None
         else:
-            regex_string = self.output_type.to_regex()
-            self.logits_processor = RegexLogitsProcessor(
-                regex_string, self.model.tokenizer
-            )
+            if isinstance(self.output_type, CFG):
+                if isinstance(self.model, LlamaCpp):
```
rlouf, line 68: We should raise this exception from the `LlamaCpp` class

outlines/types/__init__.py
```diff
@@ -103,3 +103,13 @@ class Regex:
...
         return self.definition
+
+
+@dataclass
+class CFG:
+    """Represents a Context-Free Grammar as a string."""
+
+    definition: str
+
+    def to_string(self):
```
rlouf, line 114: Do we really need this method?

outlines/generate/__init__.py
```diff
@@ -13,9 +14,6 @@
 from .regex import regex
 from .text import text
 
-if TYPE_CHECKING:
```
yvan-sraka, line 16: What was the point of that, and why did you remove it?

outlines/generate/__init__.py
```diff
@@ -33,6 +31,12 @@ class APIGenerator:
     model: APIModel
     output_type: Optional[Union[Json, List, Choice, Regex]] = None
 
+    def __post_init__(self):
```
yvan-sraka, line 34: Wouldn't `__init__` work there?

yvan-sraka (approved):
LGTM, with minor comments :)

outlines/generate/__init__.py
RobinPicard: I think that was a mistake as we need to import `RegexLogitsProcessor` to use it in the `LocalGenerator`

outlines/generate/__init__.py
RobinPicard: I don't want to interfere with the `__init__` generated by `dataclass`

Issue #1324, opened 2024-12-06
Add `ollama` integration


rlouf:
Now that Ollama has some implementation of structured outputs we can add an integration in Outlines via their [python library](https://github.com/ollama/ollama-python).

cpfiffer:
Flagging @ParthSareen, who worked on the Ollama structured outputs.

OriginalGoku:
Great idea

PR #1415, opened 2025-02-17
Create an integration with Ollama

RobinPicard:
This PR addresses issue #1324 

We add Ollama as a new type of Outlines model. As the `ollama` python sdk requires having Ollama installed on one's system, we add a step to the Github actions to install Ollama and we document this requirement in the `ollama.md` file.

Something we may need to change is the name of `APIGenerator`/`LocalGenerator` and of `APIModel`/`LocalModel`. Despite being run locally, Ollama does not support receving a logit processor and instead requires an output type-like argument. Thus, I added the `Ollama` class to `APIModel` as we need it to use the `APIGenerator`. This is however very confusing, so I think we should consider renaming those based on the type of generator needed for the different types of models. What about `LogitsProcessorGenerator` vs. `OutputTypeGenerator`?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,6 +29,10 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install Ollama
+      run: |
+        curl -fsSL https://ollama.com/install.sh | sh
+        ollama --version
     - name: Set up test environment
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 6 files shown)


rlouf:
Although `API(Model|Generator)` makes sense for the Ollama integration, `Local(Model|Generator)` becomes confusing.

We could rename both to `BlackBox(Model|Generator)` and `Steerable(Model|Generator)` for instance, or something similar in spirit. I think would be more representative of what they actually are.

rlouf (approved):

PR #1417, opened 2025-02-17
Add integration with Dottxt

### LINKED ISSUE #1411: Add `dottxt` SDK
We want to add `dottxt` SDK to Outlines. `dottxt` supports the most advanced structured generation features and can act as a drop-in replacement for local models.

RobinPicard:
This PR addresses issue #1411 

We add Dottxt as a new type of Outlines model.

Similarly as in PR #1415, there may be a decision to take regarding the naming of the `Generator` classes (and of the associated model type lists). Here I created a new `Generator` class called `CompiledAPIGenerator`. The reason for this addition is that we need a kind of generator that adds a step during initialization during which we rely on the model for compiling the output type into what the model will eventually use for generation. I tried to make it general so that it could fit other new models if some were to follow the same logic.

An alternative that would have allowed us not to create a new generator class would have been to use the existing `APIGenerator` and to put the output type compilation step in the `generate` method of the model. However, that would lead to a situation in which compilation happens during the first generation instead of during the initilialization of the `Generator` object, which defeats a bit the purpose of creating this generator object instead of directly calling the model.

Closes #1411

### DIFF
--- a/docs/reference/models/dottxt.md
+++ b/docs/reference/models/dottxt.md
@@ -1,0 +1,51 @@
+# Dottxt
+
+!!! Installation
+
+    To be able to use Dottxt in Outlines, you must install the `dottxt` python sdk.
+
+    ```bash
+    pip install dottxt
+    ```
+
+    You also need to have a Dottxt API key. This API key must either be set as an environment variable called `DOTTXT_API_KEY` or be provided to the `outlines.models.Dottxt` class when instantiating it.
+
+## Generate text
+
+Dottxt only supports constrained generation with the `Json` output type. The input of the generation must be a string. Batch generation is not supported.
+Thus, you must always provide an output type.
+
+You can either create a `Generator` object and call it afterward:
+
+```python
+from outlines.models import Dottxt
+from outlines.generate import Generator
+from pydantic import BaseModel
+
+class Character(BaseModel):
+    name: str
+
+model = Dottxt()
+generator = Generator(model, Character)
+result = generator("Create a character")
+```
+
+or call the model directly with the output type:
+
+```python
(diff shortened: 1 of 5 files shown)


rlouf:
I think we should simplify and use the `.json` method directly, which handles the low-level logic.

aschwa:
@RobinPicard  I will wait to review until you update to use `.json`

RobinPicard:
I modified it to use `.json`. That way I rely on the `APIGenerator` instead of creating a new type of `Generator`
@rlouf @aschwa

rlouf (approved):

aschwa:
There is an issues with `outlines-core` versioning, in particular with  importing `outlines_core.json_schema import build_regex_from_schema`

This change happened in outlines-core v0.2.3 but the pyproject is pinned to 0.1.26. If you manually upgrade to v0.2.3 you cause other import errors.

RobinPicard:
> There is an issues with `outlines-core` versioning, in particular with importing `outlines_core.json_schema import build_regex_from_schema`
> 
> This change happened in outlines-core v0.2.3 but the pyproject is pinned to 0.1.26. If you manually upgrade to v0.2.3 you cause other import errors.

There is an issue with that import statement. It was fixed in another commit added to the v1.0 branch. I rebased on it, so it should work fine now @aschwa

rlouf:
I fixed the CI so that tests marked as `api_call` are not run in #1437

aschwa:
I think this import is still calling from the deprecated `outlines_core` https://github.com/dottxt-ai/outlines/blob/3b5483d8fa7ef6e5afbf3b68127b1b429f71d753/outlines/types/__init__.py#L8

PR #1425, opened 2025-02-19
Fix cache dir, make it more robust when there is no $HOME

larroy:
Make cache dir more robust.

We observed that when starting a docker container with a non existing user inside the container (using -u) outlines would try to use /.cache/ as a cache directory which will give permission denied. This PR sets an order of preference for finding the cache home uses XDG_CACHE_HOME if available and OUTLINES_CACHE_DIR is not set.

### DIFF
--- a/outlines/caching.py
+++ b/outlines/caching.py
@@ -51,16 +51,20 @@ def get_cache():
     """
     from outlines._version import __version__ as outlines_version  # type: ignore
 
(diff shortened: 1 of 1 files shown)


rlouf (approved):

larroy:
Thank you @rlouf

larroy:
What are next steps for merging?

rlouf:
Thank you!

PR #1430, opened 2025-02-20
Refactor the Transformers and TransformersVision model

RobinPicard:
This PR takes over what was started by @rlouf in #1344 

The objective is to adapt `Transfomers` and `TransformersVision` to the model interface used in v1.0. To do so, we modify the signature of the `generate` method and make it the only way of generating text with the model. We also modify the handling of model/tokenizer optional arguments to avoid manipulating them ourselves (giving to the user the responsibility of choosing those they need).

I deleted the tests for those models from `tests/generate` and instead put the tests in the `tests/models`.

I also included a commit that fixes some issues on the current version of branch v1.0 that are unrelated to this PR but were preventing me from running the tests.


outlines/models/transformers_vision.py
```diff
@@ -1,138 +1,146 @@
...
+from functools import singledispatchmethod
 
-from outlines.generate.api import GenerationParameters, SamplingParameters
 from outlines.models import Transformers
+from outlines.models.base import ModelTypeAdapter
 
-if TYPE_CHECKING:
-    from outlines.processors import OutlinesLogitsProcessor
+
+class TransformersVisionTypeAdapter(ModelTypeAdapter):
```
rlouf, line 7: `TransformersTypeAdapter`

tests/models/test_transformers_vision.py
```diff
@@ -0,0 +1,235 @@
...
+    )
+    assert isinstance(result, list)
+    assert len(result) == 2
+    result = model(
+        {
+            "prompts": [
+                "<image>Describe this image in one sentence.",
+                "<image>Describe this image in one sentence.",
+            ],
+            "images": [images[0], images[0]],
```
rlouf, line 176: Q: Is this batching?

rlouf:
You can merge after addressing my comment.

rlouf (approved):

tests/models/test_transformers_vision.py
RobinPicard: Yes. This part tests batching (2) + multiple samples (2). There are 2 images, one for each batch. We check that the output is a list of 2 lists each containing 2 elements.

Issue #1423, opened 2025-02-17
Rename the `API(Model|Generator)` and `Local(Model|Generator)` classes


RobinPicard:
As briefly discussed in PR #1415, the current names `API(Model|Generator)` and `Local(Model|Generator)` in `outlines/generate/__init__.py` are confusing as Ollama is in the `API` models while it's actually run locally. We should find names that are more representative of what those generators are to remove the confusion.

PR #1439, opened 2025-02-23
Rename Local/API into Steerable/BlackBox for models/generators

RobinPicard:
Addresses issue #1423

I think `Steerable` is good. I was looking for an antonym and came up with `Enclosed`. Other candidates I considered are `Sealed` and `Locked`.


rlouf:
What do you think of `BlackBoxModel`?

rlouf (approved):

PR #1440, opened 2025-02-23
Rename `Prompt` to `Template` and add deprecation warning to the `prompt` decorator

### LINKED ISSUE #1438: Rename s / `Prompt` / `Template`
`Prompt` is poorly called, it really represents a prompt template.

rlouf:
The current `Prompt` naming does not match what the object actually represents: a prompt *template*. I thus renamed it. I also made the decision to deprecate the `@prompt` decorator. Defining jinja templates in a docstring has always felt awkward, and users still have two ways to define a prompt template:

- Writing a Jinja template in a separate text file and load it with `Template.from_file`
- Defining a function that returns a string; anything can happen in this function, including using and rendering a Jinja template.

Closes #1438

### DIFF
--- a/docs/api/prompts.md
+++ b/docs/api/prompts.md
@@ -1,1 +1,0 @@
-::: outlines.prompts
--- a/README.md
+++ b/README.md
@@ -412,6 +412,27 @@ prompt = labelling("Just awesome", examples)
 answer = outlines.generate.text(model)(prompt, max_tokens=100)
 ```
 
+You can also load a template from a text file. Assume you have the following prompt template defined in `prompt.txt`:
+
+``` text
+You are a sentiment-labelling assistant.
+
+{% for example in examples %}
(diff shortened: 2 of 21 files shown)


docs/api/prompts.md
```diff
@@ -1 +1 @@
-::: outlines.prompts
+::: outlines.templates
```
RobinPicard, line 1: Forgot to rename the file too

RobinPicard (approved):
I think that removing the `prompt` decorator is a good idea. I found it quite confusing.

docs/api/prompts.md
rlouf: They appear as renamed here?

RobinPicard: File name*
Screenshot 2025-02-24 at 09 39 42

rlouf: Ahhh

Issue #1292, opened 2024-11-27
Update the `mlxlm` integration


rlouf:
We need to update the `mlxlm` integration to follow the new user interface.

PR #1444, opened 2025-02-24
Refactor the MLXLM model

RobinPicard:
Addresses issue #1292 

Modify the `MLXLM` class to inherit from the base `Model` class and implement the standard interface of models for the v1.0.

A noteworthy change here is in the method `OutlinesLogitsProcessor.__call__`. The `mlx-lm` library gives to the logits processor the logits in 2D even though there's a single prompt (so input_ids are 1D). As this was causing a bug (the logits processor expect those 2 variables to have the sameshape), I added an `if` statement at the start of the method to make the input_ids 2D when encountering this situation.

### DIFF
--- a/outlines/processors/base_logits_processor.py
+++ b/outlines/processors/base_logits_processor.py
@@ -81,6 +81,15 @@ class OutlinesLogitsProcessor(Protocol):
         torch_logits = self._to_torch(logits)
         input_ids = self._to_torch(input_ids)
 
+        # if input_ids is 1D and logits is 2D with a single sequence,
+        # reshape input_ids to 2D (needed for mlx-lm)
+        if (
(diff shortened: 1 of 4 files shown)


outlines/processors/base_logits_processor.py
```diff
@@ -81,6 +81,15 @@ def __call__(
         torch_logits = self._to_torch(logits)
         input_ids = self._to_torch(input_ids)
 
+        # if input_ids is 1D and logits is 2D with a single sequence,
```
rlouf, line 84: Not for this PR, but I think we'll have to break down the logits processor into library-specific ones. On top of this we'll soon have methods in `outlines-core` to mask on-device that see optimized for `torch`, MLX and `NumPy`. We can already open an issue for that.

Issue #1245, opened 2024-11-05
Run test for the integrations to API  model providers nightly


rlouf:
We need to make sure that the integration with API-based model providers works. We can use API keys for this; in order to limit costs and prevent abuse we will run these tests nightly.

PR #1447, opened 2025-02-25
Create nightly tests run in Github actions

RobinPicard:
Addresses issue #1245 

We would need to add the following 5 secrets to the environment for it to work:
- OPENAI_API_KEY
- ANTHROPIC_API_KEY
- GEMINI_API_KEY
- DOTTXT_API_KEY
- SLACK_TESTS_ERRORS_WEBHOOK_URL

### DIFF
--- a/.github/workflows/tests_nightly.yml
+++ b/.github/workflows/tests_nightly.yml
@@ -1,0 +1,60 @@
+name: Nightly Tests
+
+# Run every day at 6:00 AM UTC
+on:
+  schedule:
+    - cron: "0 6 * * *"
+
+jobs:
+  tests:
+    name: Run the tests
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -1,28 +1,25 @@
 name: Tests
 
+# Runs every day at 06:00 UTC
```
rlouf, line 3: Why are you overwriting `tests.yml` instead of creating a new workflow file?

RobinPicard: I was testing out something I saw on an issue saying the workflow must already exist for you to be able to run it before merging it (it did not work). Removing the commit.

rlouf:
If you create a new workflow and have it run for any PR on `main` it should work. Once we've fixed the issue we can set it to nightly.

Issue #1422, opened 2025-02-17
Remove the custom sampling loop


rlouf:
We won't need the custom sampling loop once v1.0 is ready. Remove it in a separate PR so we can easily revert the change in the future if we need to.

PR #1449, opened 2025-02-25
Remove the custom sampling loop (SequenceGenerator)

RobinPicard:
Addresses issue #1422


rlouf:
I see there's a lot mixed in the last commit, but could we keep `SequenceGeneratorAdapter` for now and delete the rest?

RobinPicard:
If we remove `Sampler` too we would still need to modify many files.

rlouf:
Let's keep them for the moment then, I don't want the branch to be in a broken state even if temporary.

RobinPicard:
I modified the commit to only remove `SequenceGenerator`. I created a separate issue for the rest of: #1450. It should be quite straightforward after we have finished refactoring all models and have updated/deprecated the `generate.<something>` functions.

rlouf (approved):

PR #1452, opened 2025-02-26
Add support for Transformers audio models

RobinPicard:
Addresses issue #1270 

This PR turn the class `TransformersVision` into `TransformersMultiModal` that accommodates both vision and audio models. We should be able to use this same class for video models with minimal change, but I did not find a good model to test it out.

Something I'm unsure about concerns the keys of the dict provided as a prompt when calling a `TransformersMultiModal` model: I have the impression that they are standardized among transformers processors (`text` in all cases, `images` for vision models and `audios` for audio models), but I'm not sure of it. If it turns out they are not, we could remove the check on the keys of the dict and give responsibility to the user for providing what their processor needs.

I did not add model tests for audio models as the model I found (Qwen/Qwen2-Audio-7B-Instruct) is too heavy to run in the CI (and on some people's device).


docs/reference/models/transformers_multimodal.md
```diff
@@ -0,0 +1,196 @@
...
+text = """<|im_start|>user
+Audio 1: <|audio_bos|><|AUDIO|><|audio_eos|>
+<|im_end|>
+<|im_start|>user
+Write a sentence about what you hear in the audio.<|im_end|>
+<|im_start|>assistant
+"""
+prompt = {
+    "text": text,
+    "audios": audio_from_url(URL)
```
rlouf, line 132:
```suggestion
    "audio_files": audio_from_url(URL)
```

docs/reference/models/transformers_multimodal.md
```diff
@@ -0,0 +1,196 @@
...
+You must provider a processor adapted to your model. We currently only support vision and audio models.
+
+## Use the model to generate text from prompts and images/audios
+
+When calling the model, the prompt argument you provide must be a dictionary using the following format:
+```
+{
+    "text": Union[str, List[str]],
+    "images": Optional[Union[Any, List[Any]]],
+    "audios": Optional[Union[Any, List[Any]]],
```
rlouf, line 33:
```suggestion
    "audio_files": Optional[Union[Any, List[Any]]], 
```

docs/reference/models/transformers_multimodal.md
```diff
@@ -0,0 +1,196 @@
...
+    CLIPModel.from_pretrained(model_name),
+    CLIPProcessor.from_pretrained(model_name)
+)
+```
+
+You must provider a processor adapted to your model. We currently only support vision and audio models.
+
+## Use the model to generate text from prompts and images/audios
+
+When calling the model, the prompt argument you provide must be a dictionary using the following format:
```
rlouf, line 28: I like the dictionary approach better than my type-based approach

rlouf (approved):
Just a minor naming issue

docs/reference/models/transformers_multimodal.md
RobinPicard: I used `audios` because it's the keyword used by the Gwen2Audio processor (for the same reason I changed `prompts` into `text`). I think letting users choose whatever keyword their processor uses (apart from text that's shared by all) may be the best solution. We would still use `inputs = self.processor(**model_input, padding=True, return_tensors="pt")`. Another advantage is that it lets users provide optional kwargs for the processor call.

rlouf: If those are the name of the arguments in `transformers` so we can do `inputs = self.processor(**model_input, padding=True, return_tensors="pt")` let's keep them this way.

PR #1454, opened 2025-02-27
Create the `Application` class

RobinPicard:
Addresses issue #1346 

The class will actually be called `Application` instead of `Outline`, but it will otherwise implement the interface suggested in the linked issue. We additionally remove the `functions` module.


outlines/macros.py
```diff
@@ -0,0 +1,58 @@
...
+    my_macro = Macro(model, template, JsonType(OutputModel))
+
+    result = my_macro(num=3)
+    print(result)  # Expected output: { "result" : 6 }
+    """
+    model: Model
+    template: Union[Template, Callable]
+    output_type: Any
+
+    def __post_init__(self):
```
rlouf, line 52: Could we add a `with_backend` method that allows a user to change how a model is going to be run? And lazily load the model and compile lazily so this does not happen when the `Macro` is instantiated, only when it is used?

The idea is to eventually have people only pass a string with the model name.

rlouf:
Wondering if we should simply call it `Application`

rlouf:
Let's use `Application` for now

rlouf:
We had an offline discussion where we decided to pass the model as an argument when calling the application. This makes `Application` object more re-usable.

Issue #1301, opened 2024-11-30
Deprecate `generate.json`, `generate.format`, `generate.regex`, `regex.choice`, `regex.cfg`


rlouf:
We use them as aliases to `build_generator` with the appropriate type and add a deprecation notice.

PR #1462, opened 2025-02-28
Remove the generate functions

RobinPicard:
Addresses issue #1301 

The generate functions are not needed anymore as they are being replaced by the `Generator` object. We remove the files in which they were defined and the associated documentations (some md files that will need to be completely rewrote in the future are not modified here).


rlouf (changes requested):
Let's move `generate/__init__.py` to `generator.py`

rlouf (approved):

Issue #1457, opened 2025-02-27
Enforcing a conditional logic in responses


reuning:
I've been working on using this to categorize twitter posts into 1 of 7 categories. One issue I've had so far is that models seem to be overly willing to identify meaning in posts that don't have a lot of content. I thought about solving this by trying to use a chain-of-thought style approach where the prompt asks to first identify if there is enough information and then code the tweet if there is enough (and if there isn't enough it should return "OTHER"). 

The problem I'm having is that there doesn't seem to be an obvious way to _force_ that requirement. So I will get posts identified as having not enough information and then a non-OTHER code applied. 

Below is what I'm using. I'm afraid I might be missing something obvious. 

```
class Information(str, Enum):
    Yes = "Yes"
    No = "No"

class Code(str, Enum):
    MACROECONOMICS = "MACROECONOMICS"
    CIVIL_RIGHTS = "CIVIL RIGHTS"
    HEALTH = "HEALTH"
    IMMIGRATION = "IMMIGRATION"
    DEFENSE = "DEFENSE"
    CAMPAIGNING = "CAMPAIGNING"
    OTHER = "OTHER"

class Coding(BaseModel):
    information: Information = Field(..., description="Enough information to categorize the post?")
    conclusion: Code = Field(..., description="Final coding of post")

json_schema = Coding.model_json_schema()

instructions = f"""
You are a world class AI model who answers questions in JSON using the following schema: \n<schema>\n{json_schema}\n</schema>
You will use this to classify social media posts from Members of the United States Congress into one of seven categories:

 - MACROECONOMICS includes posts about economic trends, economic policies, unemployment, taxation, or the budget.
 - CIVIL RIGHTS includes posts about racial or gender discrimination, voting rights, the first amendment or privacy.
 - HEALTH  includes posts about health care policies, regulation of insurance companies, mental health, or disease prevention.
 - IMMIGRATION includes posts about immigrants, refugees, undocumented individuals, or deportation 
 - DEFENSE includes posts about the military, alliances, veteran affairs, or the national guard. It does not include posts about the police.
 - CAMPAIGNING includes posts urging people to vote, finding a polling place, donating to a campaign, or sharing an endorsement. It does not include posts about voting rights in general.
 - OTHER includes anything that does not fit into the above categories or where there is not enough information to categorize it.

Before you categorize the post you should first decide if there is enough information in the post to categorize it. If there is not enough information then code the post as "OTHER". 

[...]

PR #1463, opened 2025-02-28
Improve DSL (and make it impossible to unknowingly make a mistake)

### LINKED ISSUE #1459: Add helper functions for the DSL
They should transform `str` arguments into `Terms`, and be expressive:

```python
from outlines.types import either, exactly, s

alternative = either("MACROECONOMICS", "CIVIL RIGHTS")
repetition = exactly(3, "lol")
repetition = at_least(3, "lol")
repetition = at_most(3, "lol")
repetition = between(2, 3, "lol")
repetition = one_or_more("lol")
repetition = zero_or_more("lol")
repetition = optional("lol")

s("lol").exactly(3)
s("lol").at_most(3)
s("lol").at_least(3)
s("lol").between(2,3)
s("lol").zero_or_more()
s("lol").one_or_more()
s("lol").optional()
```

We should also drop `|`: while `+` will work regardless of whether we're adding `Term` or `str` objects, `|` does not work for strings. 

The example in the linked issues would then read:

```python
from outlines.types import either

information_query = "Enough information to categorize the post? "
conclusion = "Final coding of post "
choices = either("macroeconomics", "civil rights")

outline = information_query + either(
    ("No. " + conclusion + "OTHER"),
    ("Yes. " + conclusion + choices)
)

```

In terms of basic types we can support:

```python
digit
char
chars("a-z")  # [a-z]
chars("aeiou")  # [aeiou]
space
tab
newline
(token?)
```

And more common types:

```python
# Identifiers/Code
class_name       # PascalCase identifier
snake_case       # lowercase_with_underscores
kebab_case       # lowercase-with-dashes
hex_color       # #RRGGBB format
version_number   # like 1.2.3

# Data formats
uuid            # standard UUID format
phone           # phone number pattern
credit_card     # credit card number format
postal_code     # zip/postal code pattern

# Date/Time
iso_date        # YYYY-MM-DD
iso_time        # HH:MM:SS
iso_datetime    # YYYY-MM-DDTHH:MM:SS

# Internet
hostname        # valid hostname
path            # file/URL path
query_param     # URL query parameter
```

Note that grouping is implicit, but it follows how people naturally think about text patterns. I don't think we need an explicit `group` operator.

After this we should have a nice and unambiguous interface.

rlouf:
Closes #1459. It addresses the issue raised in #1457. Changes:

- Remove `|`
- Make helper functions convert strings to `String` objects
- Add `either(*arsg)` function for alternatives
- Replace `times` with `exactly` for exact quantification 
- Add `at_least` function
- Add `at_most` function
- Change `between` signature

Needs documentation update and add a few extra types. After this PR the DSL should be in a stable and definitive state.

The solution to the user's original request would look like:

```python
from outlines.types import either

information_query = "Enough information to categorize the post? "
conclusion = "Final coding of post "
choices = either("macroeconomics", "civil rights")

outline = information_query + either(
    ("No. " + conclusion + "OTHER"),
    ("Yes. " + conclusion + choices)
)
```

Which is better UX and answers a problem commonly raised (path-dependent generation). And it is now impossible to make an error because of `|`, and there is a helper function to build `Alternatives`.


RobinPicard (commented):
I think all the changes are great and make the UX clearer. There's still a little issue mentioned in the latest PR before that one regarding the ascii representation of sequences though:
```python
from outlines.types.dsl import String
print(String("a") + String("b") + String("c"))

└── Sequence
    ├── Sequence
    │   ├── String('a')
    │   └── String('b')
    └── String('c')
```

rlouf:
Is your remark with the nested sequences?

RobinPicard (approved):

PR #1476, opened 2025-03-05
Tweak version commands in bug_report.yml

nchammas:
The way these commands show up when a user fills out a new bug report is misleading, as they all get mashed into one line.

Screenshot 2025-03-05 at 11 50 22 AM

Copy-pasting this results in only the first command running, which is presumably not what the template author intended.


RobinPicard (approved):
Thanks for opening a PR to fix that!

Issue #1487, opened 2025-03-08
Make it possible to pass a logits processor to `Generator`


rlouf:
See for instance https://github.com/dottxt-ai/outlines/pull/1408, but also for internal testing. This is an important design decisions, we need to consider pros & cons of each possible choice.

rlouf:
We could have users pass the logits processor as a keyword-only argument to the `Generator` object. I'm not a fan of introducing a new object or forcing users to know about the internals.

As for the implementation, I like the classmethod solution.

Issue #1414, opened 2025-02-13
Write generate tests for v1.0 refactored models/generators


RobinPicard:
As part of the v1.0 project, we are refactoring the models' and generators' interfaces. A consequence of that change is that the generate tests in file `tests/generate/test_generate` are inadequate for those new models. We stopped running those tests on refactored models in PR #1413 as they were necessarily failing.

We should write a new set of generate tests fitting the refactored models/generators.

PR #1494, opened 2025-03-13
Modify Generator to allow passing a logits processor to it

RobinPicard:
Addresses #1487 and #1414 

This PR modifies the `Generator` classes. Instead of accepting only an `output_type` argument, which could be a python type or an Interefular FSM, it now accepts 3 distinct arguments: `output_type`, `fsm` and `logits_processor`. Additionally, the PR creates tests for the `Generator` objects.


outlines/generator.py
```diff
@@ -91,8 +106,51 @@ def stream(self, prompt, **inference_kwargs):
 def Generator(
     model: Union[SteerableModel, BlackBoxModel],
     output_type: Optional[Any] = None,
+    *,
+    logits_processor: Optional[OutlinesLogitsProcessor] = None,
```
rlouf, line 110: Can we call it `processors`? I think in this context it makes sense.

outlines/generator.py
```diff
@@ -91,8 +106,51 @@ def stream(self, prompt, **inference_kwargs):
...
+    fsm: Optional[interegular.fsm.FSM] = None,
```
rlouf, line 111: I would prefer a separate output type instead of a kwarg here.

RobinPicard: So keep what we had before, right?

rlouf: Yes we could have a `FSM` object like we have `CFG` or `JsonSchema`

Issue #1445, opened 2025-02-24
Create library-specific versions of logits processors


RobinPicard:
First suggested in #1444:

> Not for this PR, but I think we'll have to break down the logits processor into library-specific ones. On top of this we'll soon have methods in outlines-core to mask on-device that see optimized for torch, MLX and NumPy. We can already open an issue for that.

The justification for this proposed change is that libraries providing local models (transformers, mlxlm, lamma.cpp...) use different object types and shapes to represent logits.

PR #1498, opened 2025-03-15
Add library specific handling of tensors in logits processors

RobinPicard:
This PR addresses issue #1445 

The PR creates classes of type `TensorHandler` for various tensor libraries. Such classes are used by the logit processors to avoid having to manipulate the `input_its/logits` tensors itself. This solution allows us not to require the user to download all tensor libraries supported by outlines and to treat the tensor in their native type without having to turn them into torch tensors. We also modify the local models to possess an attribute `tensor_library_name` to indicate to the logits processor what `TensorHandler` implementation to use.

Thanks to those change, we remove the mandatory dependencies on torch and numpy.

### DIFF
--- a/outlines/models/base.py
+++ b/outlines/models/base.py
@@ -46,10 +46,11 @@ class Model(ABC):
     attribute of type `ModelTypeAdapter`. The methods of the `type_adapter`
     attribute are used in the `generate` method to format the input and output
     types received by the model.
+    Additionally, local models must define a `tensor_library_name` attribute.
 
     """
-
     type_adapter: ModelTypeAdapter
+    tensor_library_name: str
 
     def __call__(self, model_input, output_type=None, **inference_kwargs):
         """Call the model.
--- a/outlines/models/llamacpp.py
+++ b/outlines/models/llamacpp.py
@@ -1,7 +1,7 @@
 import pickle
 import warnings
 from functools import singledispatchmethod
-from typing import TYPE_CHECKING, Dict, Iterator, List, Set, Tuple, Union
+from typing import TYPE_CHECKING, Dict, Iterator, List, Optional, Set, Tuple, Union
 
 from outlines.models.base import Model, ModelTypeAdapter
 from outlines.models.tokenizer import Tokenizer
@@ -140,13 +140,13 @@ class LlamaCppTypeAdapter(ModelTypeAdapter):
 
 class LlamaCpp(Model):
     """Wraps a model provided by the `llama-cpp-python` library."""
+    tensor_library_name = "numpy"
 
     def __init__(self, model: "Llama"):
         from llama_cpp import Llama
 
         self.model = model
         self.tokenizer = LlamaCppTokenizer(self.model)
-        self.model_type = "local"
         self.type_adapter = LlamaCppTypeAdapter()
 
     def generate(self, model_input, output_type, **inference_kwargs):
--- a/outlines/models/mlxlm.py
+++ b/outlines/models/mlxlm.py
@@ -50,6 +50,7 @@ class MLXLMTypeAdapter(ModelTypeAdapter):
 
 class MLXLM(Model):
     """Represents an `mlx_lm` model."""
+    tensor_library_name = "mlx"
 
     def __init__(
         self,
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -189,12 +189,12 @@ class Transformers(Model):
         # is not available in the environment.
         try:
             from transformers import FlaxPreTrainedModel
-        except ImportError:
+        except ImportError:  # pragma: no cover
             FlaxPreTrainedModel = None
 
         try:
             from transformers import TFPreTrainedModel
-        except ImportError:
+        except ImportError:  # pragma: no cover
             TFPreTrainedModel = None
 
         tokenizer.padding_size = "left"
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -185,10 +185,35 @@ class Transformers(Model):
             the `transformers` API for tokenizers.
 
         """
+        # We need to handle the cases in which jax/flax or tensorflow
+        # is not available in the environment.
+        try:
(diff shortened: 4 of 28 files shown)


outlines/models/base.py
```diff
@@ -51,6 +51,41 @@ class Model(ABC):
...
+        that value will be returned. Otherwise, the model's default tensor
+        library name will be used.
+        """
+        if self._tensor_library_name is not None:
+            return self._tensor_library_name
+
+        return self._default_tensor_library_name
+
+    @property
+    def _default_tensor_library_name(self):
```
rlouf, line 79: What is the idea behind having methods instead of attributes?

outlines/processors/tensor_handlers/__init__.py
```diff
@@ -0,0 +1,17 @@
+from .jax import JAXTensorHandler
+from .mlx import MLXTensorHandler
+from .numpy import NumpyTensorHandler
+from .torch import TorchTensorHandler
+
+
+def get_vector_handler(tensor_library_name: str):
```
rlouf, line 7: Could be a dictionary?

outlines/models/base.py
RobinPicard: I thought it's a clearer way of documenting in the base class what's expected of child classes (and implementation is more straightforward there). Otherwise we would need to add something like that in the child classes:
```python
  def __init__(self, tensor_library_name):
      self._tensor_library_name = tensor_library_name or "torch"
```
But I thought it may be harder to understand that there's this attribute for local models and that the associated classes need to add that line

rlouf: Can't you just specify `_tensor_library_name: str` under the class name?

RobinPicard: I thought mypy would not like it for some reason, my bad.

rlouf:
I was originally skeptical of the tensor handler the approach but it's growing on me. I would use the name `tensor_adapter` though.

> Thanks to those change, we remove the mandatory dependencies on torch and numpy.

🎉

outlines/models/llamacpp.py
```diff
@@ -140,14 +140,15 @@ def format_output_type(self, output_type):
 
 class LlamaCpp(Model):
     """Wraps a model provided by the `llama-cpp-python` library."""
+    _default_tensor_library_name = "numpy"
 
-    def __init__(self, model: "Llama"):
+    def __init__(self, model: "Llama", tensor_library_name: Optional[str] = None):
```
rlouf, line 145: Isn't  `tensor_library_name` for `llama-cpp-python` necessarily `numpy`?

outlines/models/mlxlm.py
```diff
@@ -50,11 +50,13 @@ def format_output_type(self, output_type):
 
 class MLXLM(Model):
     """Represents an `mlx_lm` model."""
+    _default_tensor_library_name = "mlx"
 
     def __init__(
         self,
         model: "nn.Module",
         tokenizer: "PreTrainedTokenizer",
+        tensor_library_name: Optional[str] = None,
```
rlouf, line 59: Isn't it necessarily `mlx`?

outlines/models/transformers.py
```diff
@@ -163,11 +163,13 @@ def format_output_type(self, output_type):
 
 class Transformers(Model):
```
rlouf, line 164: We may be able to do this automatically. It looks like all models implemented implemented with JAX inherit from [`FlaxPretrainedModel`](https://github.com/huggingface/transformers/blob/c9d1e5238a752813ba91a8751a638a09b5efbb73/src/transformers/modeling_flax_utils.py#L168)

outlines/models/vllm.py
```diff
@@ -139,8 +144,8 @@ def generate_stream(self, model_input, output_type, **inference_kwargs):
         )
 
 
-def from_vllm(model: "LLM") -> VLLM:
-    return VLLM(model)
+def from_vllm(model: "LLM", tensor_library_name: Optional[str] = None) -> VLLM:
```
rlouf, line 147: Isn't it necessarily `torch`? I don't think vLLM supports any other backend.

pyproject.toml
```diff
@@ -40,7 +39,6 @@ dependencies = [
    "typing_extensions",
    "iso3166",
    "airportsdata",
-   "torch",
```
rlouf, line 43: 🥳

rlouf (changes requested):
Looks good overall. The main comment I have is that I don't think that any library other than `transformers` supports more than one backend, and even for `transformers` there may be a way to determine the backend automatically.

RobinPicard:
Yes, you're right. I completely removed the possibility of specifying the tensor library to use when initiating a model as there's no case in which it's necessary. As a result I could simplify the logic the `tensor_library_name` attribute (just a simple attribute without distinct default value now).

outlines/models/transformers.py
```diff
@@ -185,11 +185,18 @@ def __init__(
             the `transformers` API for tokenizers.
 
         """
+        from transformers import FlaxPreTrainedModel
```
rlouf, line 188: It looks like this requires `flax` to be installed :/ Let's see if we can do this another way.

RobinPicard: Can't we just add flax to the test dependencies? We already have jax there

rlouf: It will still fail whenever users try to initialise a `transformers` model. We could use `try/except` for the checks (also add `TFPretrainedModel`).

RobinPicard: Ah yes, right

RobinPicard: I've added a `TensorAdapter` for `tensorflow`

rlouf (approved):

rlouf:
Do we have tests that initialize a `Transformers` instance for each of the backend `transformers` supports? If not that would be why coverage is failing.

RobinPicard:
Yes. The issue was that there are lines in the `Transformers` that are executed only if some libraries are not available. I added a coverage exclusion rule for them.

Issue #1500, opened 2025-03-17
Fix test for coverage check


RobinPicard:
These tests are constantly failing when working on branch v1.0. The resulting noise prevents us from noticing when we actually forget to add  tests.

PR #1501, opened 2025-03-17
Run coverage check on target branch instead of always main

RobinPicard:
Addresses issue #1500 

A source of problem was the fast that the coverage check was always comparing the last commit to the main branch, even if the target branch of the PR was not main. As a result, the diff was incorrect for all PRs related to the v1.0 branch. Here we change this to compare it to the actual target branch.


rlouf:
This is now failing for a different reason

RobinPicard:
It seems to be working fine now.

Issue #1466, opened 2025-03-03
Remove `pre-commit` runtime dependency


yvan-sraka:
Why was pre-commit added as a runtime dependency? This seems wrong.

_Originally posted by @maresb in https://github.com/dottxt-ai/outlines/issues/1387#issuecomment-2693582007_

Issue #1274, opened 2024-11-19
Generate.json and generate.regex failing on 0.1.4


sigjhl:
### Describe the issue as clearly as possible:

On 0.1.4, generate.regex and generate.json fails to execute. This is a fresh conda env with only python==3.11, transformers, and outlines==0.14 installed. 

Rolling back to 0.1.3 works as intended.

### Steps/code to reproduce the bug:

```python
Python 3.11.10 | packaged by conda-forge | (main, Oct 16 2024, 01:26:25) [Clang 17.0.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from outlines import models, generate
>>> model = models.mlxlm("meta-llama/Llama-3.2-1B-Instruct")
Traceback (most recent call last):
  File "/Users/sigjhl/miniforge3/envs/ol/lib/python3.11/site-packages/outlines/models/mlxlm.py", line 231, in mlxlm
    import mlx.core as mx
ModuleNotFoundError: No module named 'mlx'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sigjhl/miniforge3/envs/ol/lib/python3.11/site-packages/outlines/models/mlxlm.py", line 234, in mlxlm
    raise ImportError(
ImportError: The `mlx_lm` library needs to be installed in order to use `mlx_lm` models.
>>> model = models.transformers("meta-llama/Llama-3.2-1B-Instruct")
>>> class User(BaseModel):
...     name:str
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'BaseModel' is not defined
>>> from pydantic import BaseModel
>>> class User(BaseModel):
...     name:str
...
>>> generator = generate.json(model,User)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/sigjhl/miniforge3/envs/ol/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/sigjhl/miniforge3/envs/ol/lib/python3.11/site-packages/outlines/generate/json.py", line 49, in json
    generator = regex(model, regex_str, sampler)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1504, opened 2025-03-18
Restore index caching

RobinPicard:
Addresses issue #1466 

Caching for the creation of the states mapping in the `RegexGuide` was removed in PR #1278 to address issue #1274 highlighting that the cache is causing an error.

As I cannot reproduce the error with the v1.0 branch of Outlines using `outlines_core==0.1.26`, I propose we simply restore the cache on the function `cached_create_states_mapping` (and add tests for it).

### DIFF
--- a/outlines/processors/guide.py
+++ b/outlines/processors/guide.py
@@ -15,6 +15,7 @@ from outlines_core.fsm.guide import (
 )
 
 from outlines import grammars
+from outlines.caching import cache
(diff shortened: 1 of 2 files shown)


torymur:
Yep, caching was failing due to the absence of serialization support in the very early versions of `outlines-core`, which was later fixed.

torymur (approved):

rlouf (approved):

PR #1515, opened 2025-03-25
Update Model JSON Validation to Pydantic V2 Style

akaashdash:
Currently when parsing JSON responses from LLMs into the Pydantic BaseModel, `parse_raw` is being used. This is [deprecated behavior as of Pydantic v2](https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel) and is slated to be removed in the future. The new behavior is to use `model_validate_json` which behaves the same way.

No warnings are observed as all migration warnings are [suppressed](https://github.com/dottxt-ai/outlines/blob/main/pyproject.toml#L106).

It is best to migrate now before this causes issues. Outlines already [depends on Pydantic v2](https://github.com/dottxt-ai/outlines/blob/main/pyproject.toml#L34) so there is no need to support any behavior from Pydantic v1.

### DIFF
--- a/outlines/generate/json.py
+++ b/outlines/generate/json.py
@@ -50,7 +50,7 @@ def json(
         schema = pyjson.dumps(schema_object.model_json_schema())
         regex_str = build_regex_from_schema(schema, whitespace_pattern)
         generator = regex(model, regex_str, sampler)
-        generator.format_sequence = lambda x: schema_object.parse_raw(x)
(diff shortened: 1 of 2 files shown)


RobinPicard:
Hi @akaashdash! Thanks for opening a PR, good catch. Unfortunately we are about to merge the v1.0 branch into main, so I would rather not merge this PR as it would lead to merge conflicts down the road. Instead I've applied the change you've proposed to v1.0 in a different PR #1539

PR #1524, opened 2025-04-06
Update examples to use v1 interface

RobinPicard:
The objective of this PR is to update the content of the `examples` directory to make sure all files in it are using the v1 interface of Outlines and are running without errors. This does not imply that the examples are still considered relevant, this matter is left for a future PR.


cpfiffer:
We have a few cases in here where we use `Template.from_file` to load a prompt from disk, though these currently do not exist. We may need to add these back in. I do like having a mix showing off how to do inline prompts as well as on-disk prompts.

I'm still poking around, these are the big things for now.

RobinPicard:
I've added the `.txt` files in a `prompts` directory such that all the examples can be directly run without having to add anything. I've also made sure there's an about equal proportion of `Template.from_string` and `Template.from_file`.

rlouf (approved):

PR #1527, opened 2025-04-08
fix: removing prompt from text output in outlines serve

Martins6:


Martins6:
It says that the tests are failing but I don't see how they are connected at all. It seems it is a bug?


This the function failing:
```python
@pytest.mark.parametrize("pattern", REGEX_PATTERNS)
@pytest.mark.parametrize("model_fixture", ALL_MODEL_FIXTURES)
def test_generate_regex(request, model_fixture, pattern):
    model = request.getfixturevalue(model_fixture)
    generator = generate.regex(model, pattern)
    res = generator(**get_inputs(model_fixture), max_tokens=20)
    assert re.fullmatch(pattern, res) is not None, res
```

I don't see how they are connected to serve module or serve main function. So...I'm not sure what to do next 😅 

cc: @rlouf @cpfiffer

Issue #1528, opened 2025-04-08
Outlines v1 documentation


cpfiffer:
Outlines v1 has changed the interface quite a bit, and so it's a good opportunity to re-write the documentation. The documentation we have at the moment works fairly well, but it could be more intentionally designed to guide the user from basics through to more advanced topics. 

We should also overhaul the general theme and style. Currently, we use a hacked-together mkdocs theme that doesn't flow well visually. We should standardize on a common theme. At .TXT, we use [Hextra](https://imfing.github.io/hextra/) and [Hugo](https://gohugo.io/) for our commercial documentation, and 

Goals for the new documentation:

- Organized documentation flow. A common practice for documentation is the following pages:
  - Main page. Quick start, library introduction, links to various resources.
  - Getting started. A detailed guide on installation and use for our various backends.
  - Guides. Comprehensive guides to 
    - Pydantic use
    - The regex DSL
    - Available inference backends
    - Using `Application` and `Template`
    - JSON schema features supported
    - Raw JSON schema development
    - Debugging + troubleshooting
    - Best practices
    - Performance tips
    - Remote serving with vLLM
    - Migration guide to v1
  - Tutorials and cookbooks. We can basically port over what we have now and change the code.
  - Advanced usage. How to use Outlines core for direct logit modification, building custom logit processors, etc.
- Move to a new URL, probably something like docs.dottxt.co/outlines/.

I'll start with generally getting the pages and general structure set up. The main page + quick start are not too hard to set up. We can mostly re-purpose what we have with some editing and polish.

rlouf:
I think we should dedicate a section to how to use Outlines to connect to a  vLLM server (via the OpenAI client for now), one very similar on connecting to Ollama etc. It's a very common pattern, and we want to position Outlines as the SDK for structured generation.

PR #1529, opened 2025-04-08
V1 documentation

cpfiffer:
Draft PR to implement the general structure of the v1 documentation, following issue #1528.


rlouf:
Are we switching to Hextra for it to look similar to the product doc? While I like the idea of having a similar design, mkdocs-material really makes a lot of things easy, such as automatically generating API reference documentation or integrating with Google Analytics. We could tweak the theme to have something similar looking, default to dark mode, etc. but I think we might be shooting ourselves in the foot by switching to Hextra.

cpfiffer:
It's a good point about mkdocs. Partly the goal was to have a visual refresh and to match the product documentation and present a cohesive brand image, as the Outlines docs are visually kind of messy IMO.

The advantage of sticking with mkdocs is that we have an established flow for generating the docs. It's also designed for Python and tends to work quite well.

As to Google Analytics, this is [supported in Hugo](https://gideonwolfe.com/posts/sysadmin/hugo/hugogoogleanalytics/) easily. Automatic API reference is simple as well by adding a single step before the hugo site render, where we use something like [lazydocs](https://github.com/ml-tooling/lazydocs) to generate markdown files from Python libraries.

cpfiffer:
I'm going to switch this back to mkdocs to limit the amount of work we'll have to do. I still want to make some visual changes though, but we may be able to do this in a series of PRs.

cpfiffer:
I reverted this back to the original mkdocs format, and am still updating the skeleton/moving things around.

docs/index.md
```diff
@@ -1,8 +1,154 @@
...
+[GitHub](https://github.com/yourusername/outlines){ .md-button }
+
+## Features
+
+<div class="grid cards" markdown>
+
+- :material-shield-check: __Reliable__ - Guaranteed schema compliance -- always valid JSON.
+- :material-puzzle: __Feature-rich__ - Supports a large proportion of the JSON Schema spec, along with regex and context-free grammars.
+- :material-lightning-bolt: __Fast__ - Outlines has negligible runtime overhead, and fast compilation times.
+- :material-cog: __Universal__ - Outlines is a powered by Rust, and can be easily bound to other languages.
```
rlouf, line 22: This is more for `outlines-core` documentation imo

docs/index.md
```diff
@@ -1,8 +1,154 @@
...
+- :material-lightbulb: __Simple__ - Outlines is a low-abstraction library. Write code the way you normally do with LLMs. No agent frameworks needed.
+- :material-magnify: __Powerful__ - Manage prompt complexity with prompt templating.
```
rlouf, line 24: I would put the model integrations front and center, especially with the inference servers

.gitignore
```diff
@@ -15,3 +15,4 @@ benchmarks/results
 build
 docs/build
 logs
+.cursor/
```
rlouf, line 18: Can you remove this and add it to `.gitignore`?

docs/index.md
```diff
@@ -1,8 +1,154 @@
...
+# Define a Pydantic model describing the output format
+class Person(BaseModel):
+    # Set name and age fields to be string/int respectively
+    name: str
+    age: int
+
+    # This field uses a regex pattern to constrain the output
+    # to only valid email addresses
+    email: str = Field(
+        pattern=r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$',
```
rlouf, line 65: This will fail if people change the model for OpenAI, is that what we want?

docs/index.md
cpfiffer: I'll change this to say that it is extensible through a simple API, since I'd like to keep six totaly cards (and it is also true that Outlines is extensible)

docs/index.md
cpfiffer: Removed the "extensible" thing and added:

> - :material-earth: __Universal__ - work with your favorite LLM inference backend, like [vLLM](https://docs.vllm.ai/en/stable/), [transformers](https://huggingface.co/docs/transformers/en/index), [OpenAI](https://platform.openai.com/docs/guides/structured-outputs), [llama.cpp](https://github.com/ggml-org/llama.cpp), [mlx-lm](https://github.com/ml-explore/mlx-lm), and more.

docs/index.md
cpfiffer: Hm, you're right -- probably not. Let's omit the email field entirely.

cpfiffer:
Note: I'm changing the structure of the docs. Is it worth it to build all the redirects we'd need for the current sitemap?

.gitignore
cpfiffer: I'm confused here, it's in the gitignore and I don't think I've  checked .cursor in.

cpfiffer:
I have introduced many changes to the documentation, including:

- Complete reorganization, centered around the user guide, the API reference, and examples (renamed from cookbooks)
- Rewritten welcome page
- Added skeleton files for all expected documentation pages
- Simplified theming + partial light/dark mode support (it is missing syntax highlighting theme changes)

The goal of this PR is to get the basic shape down, and then to add PRs later for individual documents. I've made a few changes here and there, such as rewrite the welcome page, but for the most part this PR lays out a general structure for the docs.

I'll flag this for review.

cpfiffer:
Some screenshots to help visualize:
image
image
image
image

There's a _lot_ of writing to be done here as well.

cpfiffer:
Actually, hold off on the reviews for now. I want to do some more clean up tomorrow morning.

cpfiffer:
Let's merge this in early so we can iterate on it more later, @RobinPicard @rlouf

RobinPicard:
Awesome, I love the clean look of the website! Thanks so much for the hard work @cpfiffer. I'm in favor of merging it right away as we will update some of the sections in other PRs anyway.

RobinPicard (approved):
Awesome, I love the clean look of the website! Thanks so much for the hard work @cpfiffer. I'm in favor of merging it right away as we will update some of the sections in other PRs anyway.

rlouf:
Just run pre-commit to remove the trailing spaces, and let's merge this!

cpfiffer:
Ran pre-commit, should be good.

PR #1532, opened 2025-04-09
Restore deprecated regex DSL functions with a warning

RobinPicard:
PR #1463 improved the DSL and, while doing so, modified some of the functions.

The aim of this PR is to restore support for the functions that have been removed to make sure there's no breaking change. It includes deprecation warnings to tell users to stop using those functions and use the new instead.


cpfiffer (approved):
Works for me!

PR #1534, opened 2025-04-11
Add optional tokenizer paramers to generation

jkuntzer:
The title and the changes speak for themselves, I hope. In my case, this change was necessary to [handle a bug with the Gemma 3 models](https://github.com/google-deepmind/gemma/issues/169) whose fix involves passing additional parameters to the tokenizer when encoding the prompt. However, I think my changes allow for overall greater flexibility.

I didn't run the tests locally because it's a minimal change and I couldn't be bothered to install all of the additional libraries. I did run the pre-commit though and got no errors.

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -224,6 +224,8 @@ class Transformers:
         The generated text
         """
         prompts = prompts if isinstance(prompts, list) else [prompts]
+        tokenizer_parameters = tokenizer_parameters or {}
+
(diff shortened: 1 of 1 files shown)


jkuntzer:
Hi @RobinPicard, sure I can give it a try. However, the bug is not related to outlines; it's an inherent model problem in the transformers library, I believe. The problem was that I couldn't implement the fix in outlines.

PR #1538, opened 2025-04-12
Update the README

rlouf:
This PR is a big overhaul of the README for the release of v1. I greatly simplified the flow and added a few "real-world" examples (hidden with `<details>`). It it rendered [here](https://github.com/rlouf/outlines/blob/v1-readme/README.md).

All of the examples have been tested.


RobinPicard:
I really like it!

It feels very natural and goes straight to the point of what users care the most about instead of listing out all available features. The real world examples in the `<details>` tag are a great idea.

2 comments:
- I find the "Core Features" part is a bit confusing as it lists a lot of different things. Maybe it would be clearer if we had feature groups?
- It displays the light-mode image even in dark mode (because of that we can't read the NVIDIA name)

rlouf:
> - I find the "Core Features" part is a bit confusing as it lists a lot of different things. Maybe it would be clearer if we had feature groups?

We can separate between model integrations and features maybe?

> - It displays the light-mode image even in dark mode (because of that we can't read the NVIDIA name)

Strange, it shows the dark mode image on both machines and my phone.

RobinPicard:
Yes I think it would make to separate those. We could have the following in the Core Features part:

1. Models:
- local
- server
- API-based

2. Output types
- multiple choices
- pydantic
...

3. Other features
- Prompt templating
- Regex DSL
- Applications

I think that could be a good way of including the other features (with a link so interested people know where to get more details) without cluttering the most important parts about models and output types .

For the dark-mode picture:
Screenshot From 2025-04-15 21-50-29

cpfiffer (commented):
Only minor comment. Overall this is _excellent_. Really practical, easy to jump into. 👍🏼

rlouf:
I changed the logos, this should be good for review.

PR #1547, opened 2025-04-19
Add ensure_ascii parameter to json function

lm-cyber:
Add ensure_ascii parameter to json function in generate


rlouf:
Thank you for contributing! What was the issue prior to this PR?

rlouf:
But this is only used to dump the schema definition (dict), I'm not sure how this is solving the problem you mention.

Issue #1548, opened 2025-04-22
Create a model to generate text through vLLM


RobinPicard:
Some users, especially those using models in productions, have their LLM run on a server using vLLM.

At the moment, we can only use vLLM through the offline inference mode, meaning that the client makes the model run directly. We should create a different model that would instead make requests to a remote vLLM server. 

As this model would mostly be used for production, we need to propose an async interface.

PR #1549, opened 2025-04-22
Rename VLLM to VLLMOffline and create server-based VLLM/AsyncVLLM

RobinPicard:
Addresses #1548

For the moment, this PR only creates the vllm server models and modifies the generator


outlines/generator.py
```diff
@@ -80,22 +74,43 @@ def __init__(self, model, output_type: Optional[Any]):
...
+        return super()._from_processor(cls, model, processor)
+
     def __call__(self, prompt, **inference_kwargs):
         return self.model.generate(prompt, self.logits_processor, **inference_kwargs)
 
     def stream(self, prompt, **inference_kwargs):
         return self.model.generate_stream(prompt, self.logits_processor, **inference_kwargs)
 
 
+class AsyncSteerableGenerator(BaseSteerableGenerator):
```
rlouf, line 100: Which integration will use  `AsyncSteerableGenerator`?

outlines/models/vllm_server.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        if output_type is None:
+            return {}
+
+        term = python_types_to_terms(output_type)
+        if isinstance(term, CFG):
+            return {"guided_grammar": term.definition}
+        elif isinstance(term, JsonSchema):
+            return {"guided_json": term.schema, "whitespace_pattern": term.whitespace_pattern}
+        else:
```
rlouf, line 45: Shouldn't there be a `try` / `else` here?

cpfiffer:
Do you have an example of use here?

outlines/models/vllm_server.py
RobinPicard: The functions `python_types_to_terms` and `to_regex` already include error messages in case of unsupported types. What do you want to put in the `except` block?

outlines/generator.py
RobinPicard: None actually. I changed it back

RobinPicard:
> Do you have an example of use here?

I was using the script below to test it @cpfiffer 

<details>

<summary>Script</summary>

import asyncio
from openai import AsyncOpenAI, OpenAI
from outlines.models import from_vllm_server
from outlines.generator import Generator

class Message(BaseModel):
    role: str
    content: str


def sync_f():
    client = OpenAI(
        base_url="http://0.0.0.0:8000/v1",
    )
    model = from_vllm_server(client)

    generator = Generator(model, output_type=Message)

    result = generator(
        "Hello, world!",
        model="facebook/opt-125m",
        extra_body={"chat_template": "{% for message in messages %}{% if message['role'] == 'user' %}{{ message['content'] }}{% endif %}{% endfor %}"},
        max_tokens=10,
    )
    print("--------------- sync -----------------")
    print(result)
    print("--------------- sync streaming -----------------")
    for result in generator.stream(
        "Hello, world!",
        model="facebook/opt-125m",
        extra_body={"chat_template": "{% for message in messages %}{% if message['role'] == 'user' %}{{ message['content'] }}{% endif %}{% endfor %}"},
        max_tokens=10,
    ):
        print(result)


def async_f():
    client = AsyncOpenAI(
        base_url="http://0.0.0.0:8000/v1",
    )
    model = from_vllm_server(client)

    generator = Generator(model, output_type=Message)

    async def generate():
        result = await generator(
            "Hello, world!",
            model="facebook/opt-125m",
            extra_body={"chat_template": "{% for message in messages %}{% if message['role'] == 'user' %}{{ message['content'] }}{% endif %}{% endfor %}"},
            max_tokens=10,
        )
        print(result)
    
    print("--------------- async -----------------")
    asyncio.run(generate())

    async def generate_stream():
        # Use async for directly on the generator.stream() method
        async for result in generator.stream(
            "Hello, world!",
            model="facebook/opt-125m",
            extra_body={"chat_template": "{% for message in messages %}{% if message['role'] == 'user' %}{{ message['content'] }}{% endif %}{% endfor %}"},
            max_tokens=5,
        ):
            await asyncio.sleep(0.1)
            print(result)

    print("--------------- async streaming -----------------")
    asyncio.run(generate_stream())


if __name__ == "__main__":
    sync_f()
    async_f()
</details>

rlouf:
The integration of sync/async is very elegant. The only (minor) thing that bothers me is that, because we also support the [offline inference](https://docs.vllm.ai/en/latest/serving/offline_inference.html) we have to call the function `from_vllm_server`. I'm unsure how many people will be using the offline version, and I'm wondering if we should drop it if anything to not have to maintain something people are not going to be using.

RobinPicard:
I agree that it's annoying to have to call it `from_vllm_server`. I can't tell you how popular vLLM offline is, but I think we also have the option of calling vLLM server just `VLLM` and rename the current `VLLM` into `VLLMOffline`. That way what is the most popular and what we want to put forward has the most straightforward name (I also find it a bit counter-intuitive to call "server" what's actually the client to a remote server)

rlouf:
> but I think we also have the option of calling vLLM server just `VLLM` and rename the current `VLLM` into `VLLMOffline`.

Let's do that.

docs/reference/models/vllm_offline.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Load the model
+
+Outlines supports models available via vLLM's offline batched inference interface. You can load a model using:
+
+```python
+import outlines
+from vllm import LLM
+
+model = outlines.from_vllm_offline(LLM("microsoft/Phi-3-mini-4k-instruct"))
```
cpfiffer, line 18: Unrelated to this PR, but how does `LLM` work here?

rlouf:
The PR looks good, but we have a new unrelated error in the test workflow.

rlouf:
What's up with the tests coverage?

rlouf (approved):

Issue #1553, opened 2025-04-23
Create a model to generate text through SgLang


rlouf:
Some users, especially those using models in productions, have their LLM run on a server using [SGLang](https://github.com/sgl-project/sglang). We can have users pass the [openai client](https://docs.sglang.ai/backend/openai_api_completions.html) to a `from_sglang` function.

As this model would mostly be used for production, we need to propose an async interface.

PR #1560, opened 2025-04-28
Create the SgLang model

RobinPicard:
Addresses #1553

I've ran the tests with a real SgLang server running locally.

### DIFF
--- a/outlines/models/sglang.py
+++ b/outlines/models/sglang.py
@@ -1,0 +1,277 @@
+import json
+import warnings
+from typing import TYPE_CHECKING, Union, Optional
+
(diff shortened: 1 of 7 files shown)


outlines/models/sglang.py
```diff
@@ -0,0 +1,279 @@
...
+        The generated text.
+
+        """
+        client_args = self._build_client_args(
+            model_input,
+            output_type,
+            **inference_kwargs,
+        )
+
+        print(client_args)
```
rlouf, line 99: You forgot to remove this

rlouf (approved):

PR #1575, opened 2025-05-12
Proposition output type documentation

RobinPicard:


cpfiffer:
did we want to roll this in to #1577?

rlouf:
> did we want to roll this in to #1577?

As in include it in the other PR?

RobinPicard:
That was before the other PR, the content here is included and expanded there

Issue #1576, opened 2025-05-14
Applications do not accept inference kwargs


RobinPicard:
After having created an `Application` instance, you can call with a model and keyword arguments. However, the keywords arguments are only used to render the prompt, so you cannot provide any inference kwargs. That's an issue as they are often important (especially in the case of models with a low default max_tokens).

The only solution I can think of is adding another parameter, a dict, that would contain the kwargs specific to either the templating or the inference.

PR #1578, opened 2025-05-15
Modify Application's call method to include inference kwargs

RobinPicard:
Addresses #1576 

I propose to have users put their template args in a dict instead of their inference kwargs because the list of the template variables is fixed while they may want to add/remove optional inference parameters.

A consequence of this change is that users cannot provide template variables as positional arguments as they used to be able to if the template was a callable, but I think it's for the best.


rlouf:
I somewhat agree with your choice. However I think this reveals a design issue that we will need to address in the near future. Could you open a new issue to that effect?

Issue #1583, opened 2025-05-16
Standardize the use of the model_name argument in model loading functions


RobinPicard:
Our functions `from_` to load models have inconsistent behavior regarding providing model names. Sometimes there's no argument for it while in other cases it's either or mandatory to pass it. We should standardize the behavior.

PR #1584, opened 2025-05-16
Standardize the use of the model_name argument

RobinPicard:
Addresses #1583 

We standardize the behavior by including an optional `model_name` argument in the loading function of all models for which a model name can be provided (so all models except those of type `SteerableModel` and `TGI`).

In the text generation methods, we give precedence to the inference keyword argument for the model name, otherwise we fall back to the `model_name` value provided at initialization (if there's one)


cpfiffer (approved):
Looks great to me, thank you!

PR #1586, opened 2025-05-22
Use mkdocs script to autogenerate API reference

RobinPicard:
Edited 2025-05-22

This PR creates the script to autogenerate API reference with mkdocs and make the config change to use it and display it in the doc website. It creates a new "API Reference" section in the documentation website and rename the previous one with that name to "Features".


cpfiffer (approved):
No obvious issues to me!

PR #1592, opened 2025-05-23
Update the utility features section of the documentation

RobinPicard:
In the "Features" section of our documentation, we want to include an "Utility" subsection that contains Outlines features that are convenient, but that are not mandatory. There are 3 of those:
- applications
- templates
- the regex DSL

### DIFF
--- a/docs_new/features/utility/application.md
+++ b/docs_new/features/utility/application.md
@@ -1,5 +1,5 @@
 ---
-title: Application API
+title: Application
 ---
 
 # Application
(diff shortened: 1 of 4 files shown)


docs_new/features/utility/application.md
```diff
@@ -8,69 +8,36 @@ The `Application` class enables you to encapsulate a prompt template and an outp
 
 ## Overview
 
-An `Application` combines a prompt template with an output type, creating a reusable component that can be applied to different models.
+To create an `Application` instance, initiliaze the class with a prompt template and an output type. You can then call the application with a model and the variables defined in your template in a dictionary.
```
cpfiffer, line 11:
```suggestion
An `Application` combines a prompt template with an output type, creating a reusable component that can be applied to different models. 

Applications are useful for simplifying repeated tasks where you have a well-defined `Template` and a fixed output type, such as classification tasks or data extraction.

To create an `Application` instance, initialize the class with a prompt template and an output type. You can then call the application with a model and the variables defined in your template in a dictionary.
```

I actually liked the deleted first paragraph, let's leave it in.

docs_new/features/utility/application.md
```diff
@@ -8,69 +8,36 @@ The `Application` class enables you to encapsulate a prompt template and an outp
...
-result = movie_reviewer(model, movie_title="The Matrix")
+For instance, the following function would be equivalent to the template defined above when used in an application:
 
-# Parse the result into a Pydantic model
-review = MovieReview.model_validate_json(result)
-print(f"{review.title} ({review.year}) - {review.rating}/10")
+```python
+def template(name: str) -> str:
+    return f"Is {name} a boy or a girl name?"
 ```
```
cpfiffer, line 43: Can you give more of an example of what using this template function would look like? It's a little hard for me to visualize.

cpfiffer (approved):
Only minor comments here. I'll approve for now so you can merge if you like my edits/comments.

PR #1615, opened 2025-06-16
Modify VLLMOffline to rely on vLLM's GuidedDecodingParams argument

RobinPicard:
vLLM has modified the way through which it supports structured generation. In its V1, providing a logits processor is no longer supported and we should instead provide the desired output type in a `GuidedDecodingParams` object.

In this PR, we implement this new interface and therefore turn `VLLMOffline` into a `BlackBoxModel`.

We also update the documentation and remove the tests for batched constrained generation as the currently available backends do not support it (outlines is not yet available with vLLM v1)


rlouf (approved):
Approving, conditional to fixing the code style

PR #1655, opened 2025-06-25
Add IPv4, UUID4, and Hex String Types

jacobmarks:
Adds these three new types to the regex DSL, along with test cases and documentation.


jacobmarks:
> Looks good, but it seems that you need to rebase for tests to run.

Does this work?

RobinPicard (approved):

RobinPicard:
> > Looks good, but it seems that you need to rebase for tests to run.
> 
> Does this work?

Yes! Sorry it was a bit of trouble. Thanks a lot for this contribution!

PR #1664, opened 2025-06-27
Add support for the chat interface

RobinPicard:
Description:
- Create an `inputs` module in which we put the `Vision` and `Chat` classes
- Users can initialize a `Chat` instance by providing a list of dicts and/or `Vision` instances and can later add/remove elements from it
- Dicts provided to `Chat` are turned into instances of `Message` for format validation and easier internal handling
- When the user calls a model with a string or a list of strings, we turn them into `Chat` instances with the function `prompt_string_to_chat` if they include tags such as `{% user %}`
- The type adapters of each model is responsible for handling the `Chat` input and turning it into the appropriate format (same as with `Vision`)

Should we tell the users about the `Message` class (or let it be just internal)?
Pro: some may find it more convenient to use `Message("user", "Hello there")` rather than `{"role": "user", "content": "Hello there"}`
Con: one more entity

### DIFF
--- a/example_chat.py
+++ b/example_chat.py
@@ -1,0 +1,51 @@
+# user provides Chat object
+import io
+import requests
+import PIL
+import openai
+import outlines
+
+from outlines.inputs import Chat, Vision
+
+model = outlines.from_openai(
+    openai.OpenAI(),
+    "gpt-4o"
+)
+
+def get_image(url):
+    r = requests.get(url)
+    return PIL.Image.open(io.BytesIO(r.content))
+
+# user provides Chat object
+prompt = Chat([
+    ("system", "You are a helpful assistant. Answer in French."),
+    ("user", "Just a random message, ignore it."),
+    ("user", Vision(get_image("https://picsum.photos/id/237/400/300"), "Describe this image"))
+])
+print(model(prompt, max_tokens=20))
+
+# user provides a string with chat tags/messages
+prompt = """
+{%system%} You are a helpful assistant. Answer in French.{%endsystem%} 
+{%user%} Just a random message, ignore it.{%enduser%} 
+{%user%} How are you doing?{%enduser%}
+"""
+print(model(prompt, max_tokens=20))
+
+# direct use of Vision (no chat)
+print(model(Vision(get_image("https://picsum.photos/id/237/400/300"), "Describe this image"), max_tokens=20))
+
+# example of how it would work with tools (not implemented yet)
+def get_weather(city):
+    return {"temperature": 25, "condition": "sunny"}
+
+def get_news(topic):
+    return {"news": "The weather in Paris is sunny and 25 degrees Celsius."}
+
+prompt = Chat([
+    ("system", "You are a helpful assistant. Answer in French. Use the tools provided to answer the user's question."),
+    ("user", "Tell me about the weather in Paris."),
+    ("assistant", "{'tool': 'get_weather', 'args': {'city': 'Paris'}}"),
+    ("function", "{'temperature': 25, 'condition': 'sunny'}")
+])
+print(model(prompt, max_tokens=20, tools=[get_weather, get_news]))
--- a/example_chat.py
+++ b/example_chat.py
@@ -5,7 +5,7 @@ import PIL
 import openai
 import outlines
 
-from outlines.inputs import Chat, Vision
+from outlines.inputs import Chat, Image
 
 model = outlines.from_openai(
     openai.OpenAI(),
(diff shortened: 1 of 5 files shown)


rlouf:
We need to think about users calling the model several times in a row, so it needs to be easy to add a new message to the chat history. For instance:

```python
model = ....

chat = Chat()
chat.add("user", "Prompt")
result = model(chat)
chat.add("assistant", result)
result = model(chat)
```

`Message` seems unnecessarily cumbersome here.

outlines/inputs.py
```diff
@@ -0,0 +1,179 @@
...
+    Parameters
+    ----------
+    role
+        The role of the message.
+    content
+        The content of the message.
+
+    """
+    role: Literal["user", "assistant", "system"]
+    content: str
```
rlouf, line 59: Is `content` always a string? e.g. for vision models

outlines/inputs.py
```diff
@@ -0,0 +1,179 @@
...
+            message will be removed.
+
+        """
+        self.messages.pop(index)
+
+    def __str__(self):
+        return "\n".join(str(message) for message in self.messages)
+
+
+def prompt_string_to_chat(prompt: str) -> Optional[Chat]:
```
rlouf, line 155: Using Jinja's parser via [extensions](https://jinja.palletsprojects.com/en/stable/templates/#extensions) would be more robust.

RobinPicard: Good idea

outlines/inputs.py
```diff
@@ -0,0 +1,215 @@
...
+    video
+        The video to use to generate the response.
+    keyword: str = "video"
+        The keyword to use to pass the video to the model. Only required for
+        `TransformersMultiModal` models with non-standard processors.
+
+    """
+    prompt: str
+    video: Any
+    keyword: str = "video"
```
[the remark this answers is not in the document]
RobinPicard: Yes, it's just for TransformersMultiModal in some cases. Most users would not have to change it (and would not know about it)

outlines/inputs.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def pop(self) -> Tuple[str, Union[str, Vision, Video, Audio]]:
+        """Remove a message from the chat."""
+        return self.messages.pop()
+
+    def __str__(self):
+        return "\n".join(str(message) for message in self.messages)
+
+
+def prompt_string_to_chat(prompt: str) -> Optional[Chat]:
```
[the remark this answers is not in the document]
rlouf: The goal is also to be able to call `Template.from_file` on a prompt file.

RobinPicard:
I've put examples in `example_chat.py`

Rationale for the proposition:
- Have as few objects as possible: only `Chat` and those for assets (`Vision`, `Audio`...)
- `Chat` is needed because it must contain a list and the list type is already used for batching
- The asset objects are useful for convenient dispatching, type casting and to encapsulate both the asset and the associated prompt
- The elements in the `Chat` list are tuples of 2 elements with the 1st one being the role that allows to know how to use the 2nd element. The 2nd elements are strings or assets, but we could have a dictionary if needed in the case of tools use in the future
- The asset objects can be used both within a Chat message as the content and as a standalone input when calling a multimodal model without the chat format

example_chat.py
```diff
@@ -0,0 +1,51 @@
...
+# user provides a string with chat tags/messages
+prompt = """
+{%system%} You are a helpful assistant. Answer in French.{%endsystem%} 
+{%user%} Just a random message, ignore it.{%enduser%} 
+{%user%} How are you doing?{%enduser%}
+"""
+print(model(prompt, max_tokens=20))
+
+# direct use of Vision (no chat)
+print(model(Vision(get_image("https://picsum.photos/id/237/400/300"), "Describe this image"), max_tokens=20))
```
rlouf, line 36: what about the following alternative?

```python
model("Describe this image", items=[Image(get_image(....))], max_tokens=20)
```

example_chat.py
```diff
@@ -0,0 +1,52 @@
...
+    "gpt-4o"
+)
+
+def get_image(url):
+    r = requests.get(url)
+    return PIL.Image.open(io.BytesIO(r.content))
+
+# user provides Chat object
+prompt = Chat([
+    ({"role": "system", "content": "You are a helpful assistant. Answer in French."}),
```
rlouf, line 21: With between parentheses?

outlines/inputs.py
```diff
@@ -0,0 +1,130 @@
...
+    ])
+    ```
+
+    Parameters
+    ----------
+    messages
+        A list of messages.
+
+    """
+    messages: List[Dict[str, Any]]
```
rlouf, line 112: How about initializing with `Chat(system="")`?

outlines/inputs.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def append(self, message: Dict[str, Any]):
```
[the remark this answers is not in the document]
rlouf: `add_user_message("This is a message", Image(...))`. What else could there be?

rlouf:
I've taken some time to think about the design, and here is my current thinking. I'm not adding model calls for simplicity

## Simple workflow

```python
chat = Chat()
chat.add_system("You are a friendly and helpful assistant.")
chat.add_user("Hi! Can you explain what Python decorators are?")
chat.add_assistant("""Hello! I'd be happy to explain Python decorators.
```

## Multimodal

```python
red_img = Image(PILImage.new('RGB', (100, 100), color='red'))
blue_img = Image(PILImage.new('RGB', (100, 100), color='blue'))
green_img = Image(PILImage.new('RGB', (100, 100), color='green'))

image_chat = Chat()
image_chat.add_system("You are an AI assistant capable of analyzing and comparing images.")

image_chat.add_user("What color is this image?", red_img)
image_chat.add_assistant("This image is red.")

image_chat.add_user("Can you compare these three images?", red_img, blue_img, green_img)

image_chat.add_assistant("""I can see three images with different colors:
1. First image: Red
2. Second image: Blue  
3. Third image: Green""")
```

### Function calling

```python
def get_weather(location):
    weather = ...
    return weather
chat = Chat()
chat.add_system("You are a helpful assistant.")
chat.add_user("What's the weather in Tokyo?", tools=[weather_tool])

# If, in the future, the model call returns a `Response` object we should
# allow the user to do chat.add_assistant(response) to simplify
chat.add_assistant(
    content="I'll check the weather in Tokyo for you.",
    tool_calls=[{
        "id": "call_123",
        "type": "function",
        "function": {
            "name": "get_weather",
            "arguments": '{"location": "Tokyo"}'
        }
    }]
)
chat.add_tool_response("call_123", "get_weather", '{"temperature": 22, "condition": "sunny"}')
chat.add_assistant("The weather in Tokyo is sunny with a temperature of 22°C.")
```

PR #1668, opened 2025-06-30
Fix error with additionalProperties for the OpenAI model

### LINKED ISSUE #1666: Fix error with `additionalProperties` in the OpenAI backend
Hello! I've faced an error during usage OpenAI Compatible API of Open router

Error code: 400 - {'error': {'message': 'Provider returned error', 'code': 400, 'metadata': {'raw': '{\n  "error": {\n    "message": "Invalid schema for response_format \'default\': In context=(), \'additionalProperties\' is required to be supplied and to be false.",\n    "type": "invalid_request_error",\n    "param": "response_format",\n    "code": null\n  }\n}', 'provider_name': 'OpenAI'}}, 'user_id': 'user_xxxxxxxxxxxxxxxxx'}

Seems like there is required a custom client for that purpose.

RobinPicard:
Closes #1666

### DIFF
--- a/outlines/models/utils.py
+++ b/outlines/models/utils.py
@@ -1,0 +1,27 @@
+import jsonpath_ng
+
+
+def set_additional_properties_false_json_schema(schema: dict) -> dict:
+    """Set additionalProperties to False to all objects in the schema using jsonpath.
+
+    Parameters
+    ----------
+    schema
+        The JSON schema to modify
+
+    Returns
+    -------
+    dict
+        The modified schema with additionalProperties set to False
+    """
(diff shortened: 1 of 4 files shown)


outlines/models/utils.py
```diff
@@ -0,0 +1,58 @@
+def set_additional_properties_false_json_schema(schema: dict) -> dict:
```
rlouf, line 1: Would using `jsonpath` simplify things here?

RobinPicard: Ah I did not know about it, much simpler indeed!

PR #1674, opened 2025-07-02
Add a batch method to models

RobinPicard:


rlouf:
The interface looks good, I'll take a closer look at the code soon. In the meantime, we still need to update the documentation.

rlouf (approved):

PR #1675, opened 2025-07-03
Fix: JSON schema union types (type arrays) fail with 'type must be a string' error

### LINKED ISSUE #1383: Dynamic schema creation

### LINKED ISSUE #1676: JSON schema union types (type arrays) fail with 'type must be a string' error
## Description

When using JSON schemas with union types specified as arrays (e.g., `{"type": ["string", "null"]}`), outlines fails with a ValueError: "'type' must be a string". This prevents the use of optional fields and other union type patterns in dynamic JSON schema generation.

## Current Behavior

```python
from outlines.types import JsonSchema

schema = {
    "type": "object",
    "properties": {
        "age": {"type": ["integer", "null"]}  # Union type as array
    }
}

# This fails with: ValueError: 'type' must be a string
json_schema = JsonSchema(schema)
```

## Expected Behavior

The schema should be accepted and processed correctly, as union types using array notation are valid according to the JSON Schema specification.

## Root Cause

The current version of `outlines-core` (0.1.26) does not support type arrays. This was fixed in `outlines-core` PR dottxt-ai/outlines-core#138 which adds support for type arrays, but this fix is only available in `outlines-core` v0.2+.

However, upgrading to `outlines-core` v0.2+ requires significant changes due to breaking API changes (see #1380).

## Workaround

Until `outlines` can be updated to use `outlines-core` v0.2+, we need a Python-side preprocessing step that converts type arrays to the equivalent `anyOf` format.

## Examples of Affected Schemas

1. Simple optional field:
```json
{"type": ["string", "null"]}
```

2. Nested optional objects:
```json
{
    "type": ["object", "null"],
    "properties": {...}
}
```

3. Arrays with optional items:
```json
{
    "type": "array",
    "items": {"type": ["string", "null"]}
}
```

## Related Issues
- dottxt-ai/outlines-core#138 (the fix in outlines-core)
- #1380 (tracking outlines-core v0.2 migration)

brightlikethelight:
## Summary
- Fixes #1676 by adding a Python-side workaround for JSON schema union types (type arrays)
- Converts type arrays like `["string", "null"]` to `anyOf` format that outlines-core 0.1.26 supports
- Handles nested union types recursively

## Background

JSON schemas with union types specified as arrays (e.g., `{"type": ["string", "null"]}`) currently fail with a ValueError: "'type' must be a string". This prevents the use of optional fields and other union type patterns in dynamic JSON schema generation.

The root cause is that the current version of outlines-core (0.1.26) does not support type arrays. This was fixed in outlines-core PR dottxt-ai/outlines-core#138, but that fix is only available in outlines-core v0.2+. Upgrading to v0.2+ requires significant changes due to breaking API changes (see #1380).

## Solution

This PR implements a Python-side preprocessing step that:
1. Recursively traverses JSON schemas
2. Converts type arrays to the equivalent `anyOf` format
3. Preserves type-specific constraints (minLength, pattern, etc.)

This is a temporary workaround until outlines can be migrated to outlines-core v0.2+.

## Test plan
- Added comprehensive unit tests in `test_json_schema_union_types.py`
- Tests cover simple optionals, nested optionals, arrays with optional items, and constraint preservation
- All tests pass with the current implementation

## Example

Before preprocessing:
```json
{
  "type": "object",
  "properties": {
    "age": {"type": ["integer", "null"]}
  }
}
```

After preprocessing:
```json
{
  "type": "object", 
  "properties": {
    "age": {"anyOf": [{"type": "integer"}, {"type": "null"}]}
  }
}
```

## Related Issues
- #1676 (this issue)
- dottxt-ai/outlines-core#138 (the fix in outlines-core)
- #1380 (tracking outlines-core v0.2 migration)


RobinPicard (changes requested):
Thanks a lot for both the detailed issue and this excellent contribution! I have just a little comment, but looks good to me otherwise. Upgrading `outlines-core` to v0.2+ is something we want to do in the coming weeks fyi.

RobinPicard:
I don't think we need all that for such a small change (especially considering it'll be removed in a few weeks). I think what you had in your 1st commit was sufficient. There's still the issue of the `ensure_ascii` argument.

RobinPicard:
Merging is blocked by failing coverage check

RobinPicard (approved):

PR #1677, opened 2025-07-03
Create the inputs module and modify the treatment of multimodal inputs

### LINKED ISSUE #1649: Create module for model inputs
We current have the `Vision` class, used to provide a vision input as the prompt given to a model to generate text in the `template` module. This feels a bit awkward as prompt templates are a different type of object. We should consider creating a new dedicated module and should have a more systematic approach to prompt classes in general.

RobinPicard:
Closes #1649

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -223,7 +223,6 @@ class Transformers(Model):
 
         tokenizer.padding_side = "left"
         self.model = model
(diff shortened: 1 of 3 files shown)


RobinPicard:
Nightly tests are running without errors. Ah yes, good idea for the manual trigger.

rlouf:
Yeah let's open an issue for that. Definitely better than nightly

PR #1680, opened 2025-07-07
Create Chat input and add it to the models

### LINKED ISSUE #1629: Better support instruct models
We currently only support simple prompts, which requires users to manually add the tokens for conversations. We could allow them to pass a dictionary and convert it automatically to a prompt using HF tokenizers.

RobinPicard:
Closes #1629

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -1,4 +1,4 @@
-"""Integration with the `transformers` library."""
+"""Integration with the `transformers` library. """
 
 import warnings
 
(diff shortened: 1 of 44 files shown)


outlines/models/transformers.py
```diff
@@ -456,33 +549,34 @@ def __init__(self, model: "PreTrainedModel", processor):
 
         super().__init__(model, tokenizer)
 
-        self.type_adapter = TransformersMultiModalTypeAdapter()
+        self.type_adapter = TransformersMultiModalTypeAdapter(
```
rlouf, line 552: Is there a reason to prefer instantiating the type adapter with a tokenizer instead of using `apply_chat_template` from within this class?

rlouf:
I only have one question that's non blocking. Great work!

outlines/models/transformers.py
RobinPicard: The reason is just to try to contain as much user input formatting logic as possible in the type adapter to follow the general pattern.

rlouf: That's a good reason

PR #1700, opened 2025-07-29
Upgrade outlines-core to 0.2.11 and remove experimental CFG code

### LINKED ISSUE #1402: Bump `outlines-core` version to 0.2.3
**Dependency**: `outlines-core`
  **Version**: 0.2.3
  **Release Notes**: https://api.github.com/repos/dottxt-ai/outlines-core/releases/198329713

RobinPicard:
Closes #1402 #1380

Upgrading `outlines-core` to 0.2.11 means that we need to remove the `CFGGuide` and `CFGLogitsProcessor` that relied on the earlier version of the package. As a result, the `OutlinesCore` backend does not support the CFG output type anymore. We also remove the whole `fsm` module and the the `FSM` output type as those were related to the old logit processors. Finally, we delete the benchmarks as the operations for which we would want to check performance are happening in backend packages instead of in `outlines`.


pyproject.toml
```diff
@@ -24,7 +24,6 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
 dependencies = [
-   "interegular",
    "jinja2",
    "lark",
    "nest_asyncio",
```
rlouf, line 29: I also think we could have dropped this in a previous PR

pyproject.toml
```diff
@@ -39,7 +38,7 @@ dependencies = [
    "typing_extensions",
    "iso3166",
    "airportsdata",
-   "outlines_core==0.1.26",
+   "outlines_core==0.2.11",
    "genson",
    "torch",
```
rlouf, line 43: Unrelated to this PR, bu why do we still need `torch` as a dependency?

outlines/fsm/parsing.py
```diff
@@ -1,1129 +0,0 @@
...
-    Sequence,
-    Set,
-    Tuple,
-    Union,
-)
-
-import interegular
-from interegular.fsm import FSM, Alphabet, OblivionError
-from interegular.patterns import Unsupported
-from lark import Lark, Token
```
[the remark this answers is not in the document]
RobinPicard: There's nothing related to interegular or lark left anymore.

pyproject.toml
```diff
@@ -24,7 +24,6 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
 dependencies = [
-   "interegular",
    "jinja2",
    "lark",
```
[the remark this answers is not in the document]
RobinPicard: Ah yes, I forgot that.

pyproject.toml
rlouf: Let's remove it, it's a really heavy dependency.

rlouf (approved):

PR #1703, opened 2025-07-30
import fix in base.py

Oni-giri:
Fix for issue referenced in https://github.com/dottxt-ai/outlines/issues/1702


rlouf:
Thank you! We will need to cut a new release if this bug is present on the version in PyPi

Issue #1627, opened 2025-06-19
Enable thinking models


rlouf:
A v0 would be allowing users to instantiate the logits processor `ThinkingLogitsProcessor` with the end of thinking token and pass it to the generator.

```python
model = ...
logits_processor = ThinkingLogitsProcessor("</think>", output_type)
generator = Generator(model, processor=logits_processor)
```

The implementation is simple:

1. While `</think>` has not been observed, `__call__` is a pass-through.
2. After `<think>` has been generated, mask the logits to generate the type passed by the user.

TheooJ:
Agree that this would be very useful. For now, the best I've found is to add ``reasoning: str`` to the top of the type, which significantly improves performance (both for thinking and non-thinking models).

Could there also be a ``start_processing_logits=</think>`` directly in the Generator ? In any case I like the proposed API too.

RobinPicard:
Adding `reasoning: str` to the top of the type is when you're using a Pydantic class to represent a JSON schema, right?

rlouf:
A quick note, unrelated to the discussion. During internal evaluations, we've come across a couple of instances where the model "thinks" beyond the total context length. This makes me think we could add an option to limit the number of thinking tokens.

PR #1706, opened 2025-08-01
Create our own logits processor for Xgrammar, add support for mlx

### LINKED ISSUE #1698: Add support for the MLXLM models for the Xgrammar backend
The `Xgrammar` backend currently relies on a logits precessor directly provided by the library. An issue is that this logits processor only works with `torch` tensors (what's used by the `Transformers` model). As a result, it cannot be used with the `MLXLM` models that use `mlx` tensors while `xgrammar` does provide support for `mlx`.

To fix that issue, we should use the lower-level API of the `xgrammar` library and create a logits processor for that backend ourselves. That way, we would be able to handle both `torch` and `mlx` tensors. This is already what we do for the `LLGuidance` backend, so it should be possible to draw inspiration from it.

Additionally, creating our own logits processor is needed to add extra features such as a thinking mode.

RobinPicard:
Closes #1698 

This is a dependency for adding support for thinking models (#1627) as we will need to add some custom logic to the logits prorcessor, so we must create our own.

The PR adds supports for `mlx` but not for `numpy` (so the `LlamaCpp` model is not supported) as it does not seem that Xgrammar currently supports `numpy` tensors. If that changes in the future or if we find a way to make it work, it should be easy to add as the logits processor already has the right architecture for it.

PS: coverage check is still broken, there's an issue for that we should work on at some point

### DIFF
--- a/docs/features/advanced/backends.md
+++ b/docs/features/advanced/backends.md
@@ -47,7 +47,7 @@ As mentioned previously, selecting the structured generation backend is only app
 | **Models** | | | |
 | Transformers | ✅ | ✅ | ✅ |
 | LlamaCpp | ✅ | ✅ | ❌ |
-| MLXLM | ✅ | ✅ | ❌ |
+| MLXLM | ✅ | ✅ | ✅ |
 | **Output Types** | | | |
 | JSON Schema | ✅ | ✅ | ✅ |
 | Regex | ✅ | ✅ | ✅ |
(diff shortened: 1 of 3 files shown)


RobinPicard:
The PR to add support for thinking models and forcing them to stop thinking after a given number of tokens have been used is coming soon. It will be available for all steerable models/tensor types.

PR #1723, opened 2025-08-11
Alternative implementation of thinking mode

RobinPicard:


outlines/backends/__init__.py
```diff
@@ -96,13 +125,20 @@ def get_regex_logits_processor(
...
     )
-    return backend.get_regex_logits_processor(regex)
+    backend_logits_processor = backend.get_regex_logits_processor(regex)
+    if end_thinking_tag is not None:
+        end_thinking_token_id = _get_end_thinking_token_id(end_thinking_tag, model)
+        return ThinkingLogitsProcessor(end_thinking_token_id, thinking_max_tokens, backend_logits_processor)
+    return backend_logits_processor
 
 
 def get_cfg_logits_processor(
```
rlouf, line 135: Idem

outlines/backends/outlines_core.py
```diff
@@ -90,7 +90,7 @@ def _setup(self, batch_size: int, vocab_size: int) -> None:
         ]
 
     def _bias_logits_mlx( # pragma: no cover
-        self, batch_size: int, logits: TensorType
+        self, batch_size: int, logits: TensorType, skip: list[bool]
```
rlouf, line 93: If we go with this design, I would consider a different name like `passthrough`

outlines/processors/thinking_logits_processor.py
```diff
@@ -0,0 +1,42 @@
...
+                latest_token_id = self.tensor_adapter.to_scalar(input_ids[i][-1])
+                if latest_token_id == self.end_thinking_token_id:
+                    self._is_thinking[i] = False
+                elif self._num_tokens_generated >= self.thinking_max_tokens:
+                    logits[i][self.end_thinking_token_id] = float("inf")
+
+        if all(self._is_thinking):
+            return logits
+
+        return self.logits_processor.process_logits(input_ids, logits)
```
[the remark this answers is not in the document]
RobinPicard: That would be the best although it means the downstream logits processor needs to be able to handle tensors of different batch sizes and not always in the same order. I'm going to look into how constraining it is.

PR #1726, opened 2025-08-13
Jax/tensorflow deprecation

### LINKED ISSUE #1725: Deprecate jax and tensorflow
The tensor adapters module has support for those 2 tensor libraries, but the backend libraries (`outlines-core`, `xgrammar` and `llguidance`) do not support them. As a result, the tensor adapters for jax and tensorflow are never used and the `Transformers` models using them cannot be used without logits processors (so only unstructured generation is supported).

We should delete the unused code in the tensor adapters and add a deprecation warning in the `Transformers` model.

RobinPicard:
Closes #1725 

On top of the jax/tensorflow depreciation, this PR does 2 other things:

- Update the coverage check in the tests GitHub actions. Those were failing because of a config error. This PR fixes it and additionally require 100% for all files instead of doing a diff with the base branch and requiring 100% coverage only for the files modified by the PR
- Improve testing of the logits processors and thus fix all associated missing test coverage issues

### DIFF
--- a/outlines/backends/llguidance.py
+++ b/outlines/backends/llguidance.py
@@ -37,9 +37,6 @@ class LLGuidanceLogitsProcessor(OutlinesLogitsProcessor):
             The name of the tensor library used by the model
 
         """
-        if tensor_library_name not in SUPPORTED_TENSOR_LIBRARIES:
-            raise TypeError(f"Unsupported tensor library: {tensor_library_name}")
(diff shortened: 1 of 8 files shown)


rlouf (approved):

rlouf:
Please rebase on `main` (instead of merging `main`) before merging.

PR #1728, opened 2025-08-14
Fix correct handling of chat multimodal inputs in TransformersMM class

laitifranz:
## Description

This PR improves the handling of **multimodal** chat inputs for TransformersMultiModal models. It addresses the challenge of correctly applying chat templates to text while integrating diverse media assets (images, audio, video).

## Changes

-   **`outlines/models/transformers.py`**:
    *   Refactored `format_chat_input` to explicitly separate multimodal assets (images, audio, video) from text messages within chat inputs;
    *   Ensured that the complete `Chat` messages are passed to `self.tokenizer.apply_chat_template` to generate text prompts with correct placeholders for multimodal content;
-   **`docs/features/models/transformers_multimodal.md`**:
    *   Updated documentation to reflect the new multimodal chat input handling capabilities;
    *   Added a new usage example demonstrating how to utilize these features.

## Call for Feedback

Looking for feedback about docs and testers regarding the Chat multimodal input handling using different models/media!

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -472,20 +472,29 @@ class TransformersMultiModalTypeAdapter(ModelTypeAdapter):
         else:
             raise ValueError(
                 f"Invalid content type: {type(content)}. "
-                "The content must be a string or a list containing text and assets "
-                "or a list of dict items with explicit types."
+                + "The content must be a string or a list containing text and assets "
+                + "or a list of dict items with explicit types."
             )
 
     def _extract_assets_from_content(self, content: list) -> list:
         """Process a list of dict items."""
         assets = []
+
         for item in content:
+            if len(item) > 2:
+                raise ValueError(
+                    f"Found item with multiple keys: {item}. "
+                    + "Each item in the content list must be a dictionary with a 'type' key and a single asset key. "
+                    + "To include multiple assets, use separate dictionary items. "
+                    + "For example: [{{'type': 'image', 'image': image1}}, {{'type': 'image', 'image': image2}}]. "
+                )
+
             if "type" not in item:
                 raise ValueError(
                     "Each item in the content list must be a dictionary with a 'type' key. "
-                    "Valid types are 'text', 'image', 'video', or 'audio'. "
-                    f"For instance {{'type': 'text', 'text': 'your message'}}. "
-                    f"Found item without 'type' key: {item}"
+                    + "Valid types are 'text', 'image', 'video', or 'audio'. "
+                    + "For instance {{'type': 'text', 'text': 'your message'}}. "
+                    + f"Found item without 'type' key: {item}"
                 )
             if item["type"] == "text":
                 continue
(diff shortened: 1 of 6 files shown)


laitifranz:
## TL;DR

Added support for [multimodal model chat templates](https://huggingface.co/docs/transformers/chat_templating_multimodal#template-configuration) to `@format_input.register(Chat)` function in HF Transformers MultiModal. 

## Major Changes

-  **`outlines/models/transformers.py`**: 
   * Removed deprecated function `@format_input.register(dict)` and updated relative test since Outlines is now a version `> 1.2.0`;
   * In function `@format_input.register(Chat)`, the input is now processed by collecting assets and applying the processor's chat templates to correctly set placeholders for assets without doing it manually.
-  **`outlines/inputs.py`**: 
    * Updated `Chat` class docstring to include info regarding new multimodal chat format supported only by HF transformer MM models.
- **`docs/features/models/transformers_multimodal.md`**: 
     * Updated the guide with Chat and Batching examples.
- **`tests/models/test_transformers_multimodal*`** and **`tests/test_inputs.py`**:
     * Updated chat prompt to the new format;
     * Added test for unsupported contents and invalid type input.

## Feedback

- All checks have successfully passed;
- Personally tested in my projects for a week using Qwen/Qwen2.5-VL-7B-Instruct with complex chat messages;
- Let me know if any changes should be made or if I am missing smth from my implementation. I tried to leave the implementation as clean as possible to avoid backward incompatibility. I know that multimodal chat templates are exclusive for HF transformers multimodal models right now, but it simplifies a lot the management of asset tags and multiple assets concatenated with a complex chat structure.

Thanks @RobinPicard @rlouf for the support!

outlines/models/transformers.py
```diff
@@ -430,61 +430,47 @@ def format_input(self, model_input):
...
             if isinstance(message["content"], list):
-                images.extend(message["content"][1:])
-                messages_without_images.append({
-                    "role": message["role"],
-                    "content": message["content"][0],
-                })
-            else:
-                messages_without_images.append(message)
+                for item in message["content"]:
+                    if item["type"] == "text":
```
RobinPicard, line 442: Maybe an explicit error message explaining what should be done if the "type" key is not present would make sense

RobinPicard (commented):
Thanks a lot for the PR! On top of the tiny comments, something I think we should consider is taking a list of string + assets as the value of the `content` key on top of a list of dicts. The keys expected in those dicts seem mostly standardized so we could create the dicts for the user and hand them to the tokenizer chat templating method.

So the user would provide for instance:
```python
{
    "role": "user",
    "content": ["What's on this image?", Image],
},
```

And we would turn it for them into:
```python
{
    "role": "user",
    "content": [
        {"type": "text", "text": "What's on this image?"},
        {"type": "image", "image": Image},
    ],
},
```

What do you think? (providing directly the expected list of dicts would still be supported)

PS: you need to rebase your branch

PR #1755, opened 2025-09-17
Feature/mistral ai integration

yasteven:
This PR adds Mistral AI integration via the `mistralai` package, including:
- `Mistral` wrapper and `MistralTypeAdapter` for formatting prompts and structured outputs (Pydantic, JSON schemas, Literals).
- Support for text generation, streaming, and structured outputs (no regex/CFG due to API limitations).
- Unit tests for input/output formatting, generation, and streaming.
- Integration tests with real API calls using nested Pydantic models.
- Documentation (`mistral_doc.md`) with usage examples and limitations (e.g., no batch inference, `codestral-mamba` restrictions).
- Testing the dottxt-ai demo project for extracting earnings-reports failed to execute with the current version of outlines, so I forked that project separately and got the demo script extracting financial data from 10-K filings (`FinancialData`, `FinancialRow`) with the new mistral ai model (https://github.com/yasteven/demos/blob/testing_mistralai/earnings-reports/extraction.py). Along the way I fixed the issues in the current demo, but I don't have any GPUs to test it locally.

Please review the changes and let me know if additional tests or docs are needed.

Cheers!

-SEE


yasteven:
Robin,

In order to further ensure consistency with other models, I have added support for async, completed streaming functionality, fixed multi-modal inputs (allowing for image processing), and addressed the issues you identified for the prior PR commits above.

I have also nearly doubled the number of tests to account for these updates, including adding a real image processing test; all of them are passing (though occasionally the mistral image model makes a mistake, thinking 🟥 is the same as 🐈 ) 

I also updated the mistral.md to reflect the changes, and ran pycodestyle to catch any formatting issues that the pre-commit checks miss.

There may still be lingering comments around to eliminate, and there is always room for improvement (especially in the tests); but this is now very close to being the same quality as the other models. 

Enjoy!

R/ Steven E

yasteven:
I identified and fixed a bug where the new token streaming async implementation for the mistral model was not forwarding the structured output constraints, and since this bug wasn't captured by the tests I created a test for coverage over it. I switched from Mistrals' pixtral model to its small 3.1 for better image test reliability. 

The rest of the effort for this pull request on my end is:

- Get up to date information on Mistral into the documentation ᵉ (e.g, for available and suggested models)
- remove any superfluous comments
- ensure the async implementation is behaving correctly in an async manner in the tests... because:

the async context isn't managed by the model class; I assume the users does it (since there isn't a dedicated async api client unlike chatgpt ). 

image (2)

I'll look at the openai tests to see what the canonical calling conventions should be, reference the mistral async documentation, and resolve the differences.

Steven E 

ᵉ - I may make a demo of outlines in the future where it retrieves the model list with capabilities enumerated, and generates that section of the documentation.

yasteven:
My apologies Robin, I thought the API based models were supposed to emulate native enum support when I reviewed other implementations. I believe my most recent commit addresses those lingering issues, and I pulled in this comment thread and the diff of daf8fdb and 3e4a7dd into some LLMs who inform me that the mistral.py file issues you have raised are in fact resolved. My most recent commit also reduces the tests to a core set of integration tests. I have to move onto another task at the moment so I'll force push in the next day.

yasteven:
@RobinPicard 

```code
For this pass of mine addressing the comments since the last commit, 
I built a python tool which replicates performing a LLM chat process 
that I used to make the current changes to the relevant files to fix 
the issues you raised since the last commit. I did a quick glance to
see how well it's working, and I was able to use it to educate me on 
some of the parts I didn't understand. Below is it's commit summary:
```

Thanks for the thoughtful review—it's been invaluable in aligning this with the project's clean, maintainable style.  I drew inspiration from patterns in other integrations (like OpenAI and Anthropic) for consistency, but your points on grouping,  removals, and test simplicity make total sense. I've incorporated them below in a single commit for easy review.

### How Issues Are Addressed
Here's a quick rundown of the changes—let me know if anything needs tweaking!

**outlines/models/mistral.py**:
- **Line 82/116 (Missing line breaks)**: Added blank lines after all major import blocks and methods 
  (e.g., after `format_str_model_input`, `format_list_model_input`, etc.) for PEP 8 readability.
- **Line 85 (Reject multimodal lists)**: Early `ValueError` in `format_list_model_input` if `len(model_input) > 1`, 
  with updated docstring noting "single string prompt only."
- **Line 207 (Thematic grouping + None at top)**: Reordered `format_output_type`: `if output_type is None` first; 
  then JSON group (Pydantic/dataclass/TypedDict/Genson/JsonSchema/dict); unsupported (Regex/CFG/literal) last. 
  Added a catch-all error with the type name for better UX.
- **Line 258 (Unused `is_native_dict`)**: Removed import; simplified to `elif isinstance(output_type, dict):`.
- **Line 409 (Remove model description)**: Dropped `_get_model_description` entirely—no maintenance needed.
- **Line 584/599 (Dupe/confusing `supports_structured_output`)**: Removed from mistral.py 
  simplified in `Mistral` to `return True` (all models support per latest docs—no confusing checks or lists).

**tests/models/test_mistral.py**:
- **Line 21 (Single stable model)**: Switched to single constants: `MODEL_NAME = "magistral-large-latest"`, `VISION_MODEL = "pixtral-large-latest"`. 
  Inline 429-fallback per test via `client.models.list()` picks the first non-codestral chat model 
- **Line 89 (`@pytest.mark.api_call`)**: Added at class level for both sync/async classes 
  (covers all methods/fixtures consistently with other models).

**tests/models/test_mistral_type_adapter.py**:
- **Line 59 (Why patch?)**: Removed all `@patch` and mock assertions from input tests; 
  now assert directly on return structure (e.g., `result[0].content == "Hello world"` with minimal mistralai imports for types).
- **Line 433 (Only test type adapter)**: Stripped all integration tests (e.g., batch/streaming/vision); 
  kept pure unit tests for adapter methods only. (Moved any leftovers to `test_mistral.py` if needed.)


```code
Human Comments:

 Line 21: @RobinPicard - I would rather test with a single stable model, and we'll change it if it gets depreciated...

On the model fallback in tests: That was mainly to avoid flakiness from free-tier token limits 
(some models eat through quotas faster during CI runs). I swapped it to dynamically fetch available models 
via client.list_models() and made it have a stable one in the source.

Also I used the specific dated version due to the Mistral API documentation so that we do need to change it upon depreciation. Those kinds of failing tests drive those changes:

"Mistral AI API are versions with specific release dates. To prevent any disruptions due to model updates and breaking changes, it is recommended to use the dated versions of the Mistral AI API. Additionally, be prepared for the deprecation of certain endpoints in the coming months."

I read that as we might not even be able to rely on any of the endpoints existing in the future (mistral-large-latest).

Pong:

Line 207: @RobinPicard - Ping for this comment. We could have `if output_type is None` at the top, then the JSON schema equiv...
        
I modeled a lot of this after the patterns in other model integrations (e.g., OpenAI, Gemini) to keep things consistent, 
but I see how we can tighten it up to better match the project's style. I'm totally on board with the thematic grouping in 
format_output_type, removing the model description/maintenance, and simplifying the tests—no problem at all.

A lot of things like this go under the radar when LLMs are doing most of the work.

The parts of this function which throw errors due to an assertion that Mistral doesn't support them needs another
pass to determine that accuracy (like how I added multi-modal inputs), but I believe that would be more appropriate for a future PR.

```

Looking forward to your thoughts!

RobinPicard (approved):

Issue #1722, opened 2025-08-11
Provide type conversion among json schema and multiple-choice output types


RobinPicard:
Some models only support some types within the json schema and multiple-choice output types (for instance pydantic, but not json schema string for Gemini). It means that we need to handle all different types in the type adapter and that users need to know model-specific things.

Ideally, we would have be able to cast all types into each other, but if we can only do some that would already be good!

Examples of possible interfaces:

1. Gemini model

The model only accepts dataclasses, types dicts and pydantic models. A JsonSchema or Genson schema builder will lead to an error.

```python
elif is_dataclass(output_type):
    return self.format_json_output_type(output_type)
elif is_typed_dict(output_type):
    return self.format_json_output_type(output_type)
elif is_pydantic_model(output_type):
    return self.format_json_output_type(output_type)
# else error
```

We would have instead:

```python
if json_schema(output_type): # Evaluates to yes for all json schema output types
  return JsonSchema.convert_to(output_type, ["dataclass", "typed_dict", "pydantic"])
```

2. Ollama

The model needs the output type to be a json schema string. This is already implemented, but it's a bit wordy and we need to repeat it in all models with the same requirements.

```python
elif isinstance(output_type, JsonSchema):
    return json.loads(output_type.schema)
elif is_dataclass(output_type):
    schema = TypeAdapter(output_type).json_schema()
    return schema
elif is_typed_dict(output_type):
    schema = TypeAdapter(output_type).json_schema()
    return schema
elif is_pydantic_model(output_type):
    schema = output_type.model_json_schema()
    return schema
elif is_genson_schema_builder(output_type):
    return output_type.to_json()
```

We would have instead:

```python
if json_schema(output_type):
  return JsonSchema.convert_to(output_type, "schema_string")
```

PR #1761, opened 2025-09-25
Add json type conversion

RobinPicard:
Addresses #1722 

This PR adds 2 class methods to the `JsonSchema` class to check whether an object is a JSON schema type and to convert such an object to a different JSON schema type

This change has 2 main objectives:

- Provide intercompatibility between all different supported JSON schema types such that the user could provide any of them even if the model they use support only one in the list. This can be useful for some API-based models (Gemini)
- Simplify the code in the `format_output_type` methods of the model type adapters. We want to push the complexity upstream such that the models only have to deal with the `JsonSchema` class instead of having to treat different JSON schema types itself


outlines/types/dsl.py
```diff
@@ -324,13 +332,115 @@ def __init__(
...
+            SchemaBuilder,
+        ],
+        target_types: List[Literal[
+            "str",
+            "dict",
+            "pydantic",
+            "typeddict",
+            "dataclass",
+            "genson",
+        ]],
```
rlouf, line 381: Why not have different methods?

rlouf (approved):

Issue #1659, opened 2025-06-25
Support LMStudio backend


rlouf:
LMStudio supports structured outputs with `outlines-core`. We can use their [Python SDK](https://github.com/lmstudio-ai/lmstudio-python) to add it as a backend in Outlines.

okaditya84:
Looks like the code side of this landed in #1787 — `outlines.from_lmstudio` with sync + async models, the type adapter, and mock-based tests are all on `main`. But it never got documented: there's no `docs/features/models/lmstudio.md`, no entry under Models in the `mkdocs.yml` nav, and the capability table in `docs/features/models/index.md` doesn't list it. So the integration exists today but is essentially undiscoverable.

I'll put up a PR adding the docs page (following the structure of `ollama.md`, since the integration mirrors it — JSON Schema output, sync/async, streaming, vision via `lms.prepare_image`) plus the nav and capability-table entries. That should make this closable.

PR #1787, opened 2025-11-29
Initial lm studio integration

Anri-Lombard:
Addresses #1659.

I haven't solved the TODOs left by the ollama implementation, for example "`int`, `float` and other Python types could be supported via JSON Schema." Also have not built in image support yet.

lmstudio requires context to be entered, so deviates from current integrations since we need to enter the context at least once and currently I'm not closing - but I don't think it's needed? Correct me if I'm wrong 🙏

### DIFF
--- a/tests/models/test_lmstudio.py
+++ b/tests/models/test_lmstudio.py
@@ -20,22 +20,25 @@ from tests.test_utils.mock_lmstudio_client import (
 )
 
 
-# Image for testing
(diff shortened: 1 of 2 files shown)


RobinPicard:
Great! I'll take care of the uv.lock, I'll just add a commit to your PR with it when it's approved

RobinPicard (approved):

PR #1774, opened 2025-11-01
Add parametrized steerable model tests (Transformers backend)

kudos07:
### Summary
This draft PR introduces a parametrized test suite for steerable models
covering multiple Hugging Face models under the Transformers backend.

It focuses on one aspect of issue #1717:
- Expands initialization and inference coverage to multiple models.
- Runs each inference several times to surface flaky behavior.
- Adds a shared Hugging Face cache fixture to prevent rate-limit errors.
- Includes a backend parameter to allow future vLLM/llama_cpp coverage.

### Notes
- vLLM tests are skipped on Windows; CI can re-enable them.
- This is a draft PR as discussed with @RobinPicard to align on test organization before expanding further.

PR #1814, opened 2026-01-26
Add parametrized Transformers smoke test for tokenizer robustness

kudos07:
This PR replays the parametrized Transformers smoke test cleanly on top of `main`.

It keeps the test in `test_transformers.py`, avoids any `conftest.py` changes, and exercises tokenizer differences using `TEST_MODEL` plus a small secondary model.

Supersedes #1774.


RobinPicard (commented):
Almost there! Thanks for your perseverance!

PR #1820, opened 2026-02-28
fix(llamacpp): correct EOS attention mask and vocab truncation bugs

### LINKED ISSUE #1819: LlamaCppTokenizer bugs: EOS token incorrectly masked as padding and fallback vocab truncation collisions

I found two likely bugs in `outlines.models.llamacpp.LlamaCppTokenizer`:

1. `encode()` attention mask marks EOS tokens as padding.
2. Fallback vocab extraction (when `hf_tokenizer` is unavailable) can truncate token pieces and collapse distinct tokens to one vocab key.

I’m sharing a self-contained repro script below (no branch changes required).

```python
# save as: repro_llamacpp_tokenizer_bugs.py
from outlines.models.llamacpp import LlamaCppTokenizer
import llama_cpp


def repro_attention_mask_bug():
    class FakeHFTokenizer:
        eos_token_id = 2
        eos_token = "</s>"
        def get_vocab(self):
            return {"a": 1, "</s>": 2, "b": 3}

    class FakeTokenizerWrapper:
        def __init__(self):
            self.hf_tokenizer = FakeHFTokenizer()

    class FakeInnerTokenizer:
        def tokenize(self, _prompt_bytes, add_bos=True, special=True):
            return [1, 2, 3]  # EOS in middle of real prompt tokens
        def detokenize(self, token_ids):
            return b""

    class FakeModel:
        def __init__(self):
            self.tokenizer_ = FakeTokenizerWrapper()
        def tokenizer(self):
            return FakeInnerTokenizer()

    tok = LlamaCppTokenizer(FakeModel())
    token_ids, attention_mask = tok.encode("anything")
    print("token_ids:", token_ids)
    print("attention_mask:", attention_mask)

    # Expected for non-padded single prompt: all attended
    assert attention_mask == [1, 1, 1], attention_mask


def repro_fallback_vocab_collision_bug():
    # Two pieces share first 32 bytes; differ after that
    prefix = b"a" * 32
    pieces = {
        0: prefix + b"X",
        1: prefix + b"Y",
    }

    def fake_llama_model_get_vocab(_model):
        return object()

    def fake_llama_token_to_piece(_vocab, token_id, buffer, size, _lstrip, _special):
        piece = pieces[token_id]
        for i, byte in enumerate(piece[:size]):
            buffer[i] = byte
        # Simulate piece length > fixed buffer
        return len(piece)

    llama_cpp.llama_model_get_vocab = fake_llama_model_get_vocab
    llama_cpp.llama_token_to_piece = fake_llama_token_to_piece

    class FakeInnerTokenizer:
        def tokenize(self, _prompt_bytes, add_bos=True, special=True):
            return [0]
        def detokenize(self, token_ids):
            return b""

    class FakeModel:
        model = object()
        def tokenizer(self):
            return FakeInnerTokenizer()
        def token_eos(self):
            return 1
        def n_vocab(self):
            return 2

    tok = LlamaCppTokenizer(FakeModel())
    print("vocab:", tok.vocabulary)

    # Expected: 2 unique entries preserved
    assert len(tok.vocabulary) == 2, tok.vocabulary


if __name__ == "__main__":
    repro_attention_mask_bug()
    repro_fallback_vocab_collision_bug()


#Run as python repro_llamacpp_tokenizer_bugs.py
```

### Expected result:

```shell
- `repro_attention_mask_bug()` should keep all tokens attended for a non-padded single prompt:
  expected `attention_mask == [1, 1, 1]`.
- `repro_fallback_vocab_collision_bug()` should preserve both distinct token pieces:
  expected `len(vocabulary) == 2`.
```

### Error message:

```shell
AssertionError: [1, 0, 1]
AssertionError: {'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa': 1}
```

<details>

```bash
python -c "from outlines import _version; print(_version.version)"
python -c "import sys; print('Python', sys.version)"
pip freeze


### Context for the issue:

These look like correctness issues in tokenizer behavior and can surface as subtle constrained-decoding errors.

I may be wrong on fix details, but the repros above seem to show real behavioral problems.  
If this direction is valid, I can open a focused PR with fixes split by issue.

cc @robinpicard

### PR DESCRIPTION
**Pull request by giulio-leone**
## Summary

Fixes two bugs in `LlamaCppTokenizer` reported in #1819.

### Bug 1 – EOS token incorrectly masked in attention mask

`encode()` used `pad_token_id` (which equals `eos_token_id`) to build the attention mask. Any EOS token appearing inside a real prompt was masked as padding (`attention=0`), corrupting downstream processing.

Since batch encoding is not supported (`NotImplementedError` for lists), there is never any actual padding in the output. The fix sets `attention_mask = [1] * len(token_ids)` so every token is always attended.

### Bug 2 – Fallback vocab truncates long token pieces

The fallback path (no HF tokenizer) used a fixed 32-byte `ctypes` buffer for `llama_token_to_piece`. When a token piece exceeds 32 bytes:
- The piece is silently truncated to 32 bytes
- Two distinct tokens sharing the first 32 bytes collide in the vocabulary dict, losing an entry

The fix detects when `n > size` (the API signals truncation) and retries with a correctly-sized buffer.

### Bonus

Hoists `llama_model_get_vocab(model.model)` out of the per-token loop to avoid redundant FFI calls.

Fixes #1819

### DIFF
--- a/tests/models/test_llamacpp_tokenizer.py
+++ b/tests/models/test_llamacpp_tokenizer.py
@@ -269,8 +269,6 @@ def test_negative_n_skips_invalid_token():
         tok.__init__(model)
 
     # Token 1 (error) must not appear in the vocabulary
-    assert 1 not in tok.vocabulary.values() or all(
-        v != 1 for k, v in tok.vocabulary.items() if k != ""
-    )
+    assert 1 not in tok.vocabulary.values()
     assert tok.vocabulary["ok"] == 0
     assert tok.eos_token == eos_piece.decode()

### CODE REVIEW

File: outlines/models/llamacpp.py [hunk 1]
```diff
@@ -51,16 +51,24 @@ def __init__(self, model: "Llama"):
...
                     buffer,
                     size,
                     0,
                     True
                 )
-                token_piece = buffer[:n].decode("utf-8", errors="replace") # type: ignore
+                # n > size means the piece was truncated; retry with a
+                # larger buffer so distinct tokens are not collapsed.
+                if n > size:
+                    big = (ctypes.c_char * n)()
```
kudos07: Nice fix for truncation collisions. Could we also guard `n < 0` from `llama_token_to_piece` before decoding, so error returns don’t produce invalid token pieces?

Issue #1658, opened 2025-06-25
Make exceptions uniform across model providers


rlouf:
API calls can fail because of timeouts, rate limiting, etc. and each model provider has its specific exception classes for these. It would be good to create exceptions common to every model provider.

PR #1823, opened 2026-03-04
Implement uniform exceptions classes across model providers

plpxsk:
Implements #1658 

You can review how some providers implement these, in a comment below.

The updated **docs** show a summary of the implementation.  
If interested, also below, I include one plan that helped with research and implementation.

Overall, I tried to be comprehensive, but without creating too many classes.

Questions:
- [ ] Do we prefer a flat-er hierarchy, or more nesting, such as by Category in the summary table
For example, GenerationError is a result of a successful API call, not a failure. 
- [ ] Do we want to handle retries ourselves? Could be complicated, so I prefer letting consumers choose.
- [ ] How do we like the key API names, especially `normalize_provider_exception()`? Alternatives: `translate_`, `convert_*`, *_external_exception, `*_client_exception_` `*_sdk_exception` etc


outlines/exceptions.py
```diff
@@ -0,0 +1,346 @@
...
+    code = _extract_status_code(exc)
+    if code is not None:
+        if code in _STATUS_CODE_MAP:
+            return _STATUS_CODE_MAP[code](provider=provider, original_exception=exc)
+        if code >= 500:
+            return ServerError(provider=provider, original_exception=exc)
+        if 400 <= code < 500:
+            return BadRequestError(provider=provider, original_exception=exc)
+
+    return APIError(provider=provider, original_exception=exc)
```
RobinPicard: I think we shouldn't wrap all exceptions in `APIError` as there can be non model related errors and also it does not make too much sense for local models

RobinPicard (changes requested):
It looks mostly good, thanks a lot for the PR! My main concern is whether it makes sense to have local model exceptions go through `normalize_provider_exception` if there's no mapping for them.

For the 3 specific questions you asked:
- I prefer flatter
- Let's not handle that ourselves
- Your naming is good imo

plpxsk:
Got it. Will check limiting and test coverage

RobinPicard:
Are you still interested by working on this topic @plpxsk?

plpxsk:
Yes. I've been reworking the exceptions shape a bit based on improved tests. Will push new commits end of this week.

plpxsk:
Ready for review. 

Key changes:
- Streamlined exceptions wrapping with context manager 
- Invalid schema notifications folded under BadRequestError with hint
- Test coverage increased

Minor note:
- MistralAI version 1.9.11 (per `uv.lock`) is behind 2.x which has a different SDK exceptions setup. I prioritized v2.

Let me know any and all feedback/edits, and I will be happy to integrate.

RobinPicard (approved):
Excellent PR, thanks a lot for the hard work!

RobinPicard:
I pushed to your branch to resolve rebase conflicts @plpxsk

PR #1843, opened 2026-04-10
docs(examples): add semantix-ai semantic validation example

labrat-akhona:
## Summary

Adds a runnable example to `examples/` showing how `outlines` (structural decoding) and `semantix-ai` (semantic intent validation via local NLI) compose:

- **`outlines`** guarantees the response parses against a JSON schema.
- **`semantix-ai`** checks that the generated string actually *means* what was asked for, using a local cross-encoder NLI model (no API key required).

The script generates a structured customer-support reply with `outlines`, then scores the `reply` field for tone with `semantix-ai`. It also scores a deliberately rude counterexample with the same judge — same schema, different semantics — to make the failure mode `outlines` cannot catch on its own concrete.

## What changed since the previous revision

Per @RobinPicard's earlier feedback (`semantix-ai` doesn't depend on `outlines`, so `docs/community/examples.md` wasn't the right home), this revision:

- Drops the `docs/community/examples.md` entry.
- Adds a single self-contained worked example at `examples/semantix_semantic_validation.py`.
- Rebases onto current `main` so the diff is just the new file.

## Run

```
pip install outlines openai "semantix-ai[nli]"
python examples/semantix_semantic_validation.py
```

semantix-ai: https://github.com/labrat-akhona/semantix-ai

### DIFF
--- a/examples/semantix_semantic_validation.py
+++ b/examples/semantix_semantic_validation.py
@@ -1,0 +1,94 @@
+"""Compose outlines structural decoding with semantix-ai semantic validation.
+
+``outlines`` constrains the *shape* of an LLM output — the generated text is
+guaranteed to parse against a JSON schema. ``semantix-ai`` validates the
+*meaning* of any string field against a natural-language intent, scored
+locally by a small NLI cross-encoder (no API key required).
+
+The two layers are complementary:
+
+* ``outlines`` guarantees the response parses;
+* ``semantix`` guarantees it means what we asked for.
+
+A schema check cannot tell the difference between a polite reply and a rude
+one — they both parse. This example generates a structured customer-support
+reply with outlines, then checks the tone of the ``reply`` field with
+semantix. We then score a deliberately rude counterexample with the same
+judge to show how semantix catches a semantic violation a schema check
+cannot see.
+
+Install
+-------
+``pip install outlines openai "semantix-ai[nli]"``
+
+References
+----------
+* semantix-ai: https://github.com/labrat-akhona/semantix-ai
+"""
+
+from enum import Enum
+
+import openai
+from pydantic import BaseModel
+from semantix import NLIJudge
+
+import outlines
+from outlines import Generator
+from outlines.types import JsonSchema
+
+
+class Category(str, Enum):
+    BILLING = "billing"
+    TECHNICAL = "technical"
+    OTHER = "other"
+
+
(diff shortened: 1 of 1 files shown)


RobinPicard:
I took a look at the repo but could not see `outlines` in the dependencies. How are you using it?

labrat-akhona:
Hey @RobinPicard — fair catch, thanks for taking the time to look.

You're right: semantix-ai doesn't depend on `outlines`. It operates at a different layer — `outlines` constrains *generation* (schema/grammar), while semantix validates the *semantics* of an already-generated string against a natural-language intent via local NLI. They're complementary rather than one building on the other.

Given that, I don't think `examples.md` is the right home for this link — that page is for projects actually using `outlines`. Two options, happy with either:

1. I close this PR, build a real worked example (`outlines` for structured decoding → semantix for post-generation intent check) in a notebook, and resubmit once it genuinely uses the library.
2. If you'd rather not carry integration examples for complementary validators at all, I'll just close it.

Let me know which you prefer and I'll action it.

RobinPicard:
Hi @labrat-akhona, sorry for the delay. 1 sounds good, it could in the `examples` directory, we already have scripts showing integration with other products there

labrat-akhona:
Hey @RobinPicard — thanks for the steer. Pushed a fresh revision:

- Dropped the `docs/community/examples.md` entry.
- Added a single runnable example at [`examples/semantix_semantic_validation.py`](https://github.com/dottxt-ai/outlines/blob/main/examples/semantix_semantic_validation.py).
- Rebased onto current `main` so the PR diff is just the new file (`+94 / -0`).

The example uses `outlines.Generator` with a `JsonSchema` to structurally constrain a customer-support reply (matching the `Generator(model, JsonSchema(...))` pattern from `examples/pick_odd_one_out.py`), then runs the `reply` field through a local NLI cross-encoder from `semantix-ai` to validate the tone. It also scores a hand-written rude counterexample with the same judge to make the schema-valid-but-semantically-wrong failure mode explicit.

Happy to adjust naming / docstring style / move it into a subdirectory if you'd prefer — just let me know.

RobinPicard:
Looks good. I just changed my mind on the location, could you put it in `examples/community` please? Also put in the `__init__.py` of the `community` directory a little comment on how it contains examples provided by members of the community on how to use `outlines` with other libraries?

PR #1849, opened 2026-04-20
fix: return Pydantic model instance from SteerableGenerator for BaseModel output types

### LINKED ISSUE #1526: Outlines v1 response model validation
In Outlines v1, we specify an output format with `model(prompt, OutputClass, ...)`.

The current behavior of this is to provide a JSON string, rather than the validated model class `OutputClass`.

Example code:

```json
import json
from outlines import models

class Person(BaseModel):
    name: str
    age: int
    email: str = Field(pattern=r'^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$')


model_id = "HuggingFaceTB/SmolLM2-135M-Instruct"
model = models.from_transformers(
    AutoModelForCausalLM.from_pretrained(model_id),
    AutoTokenizer.from_pretrained(model_id)
)

person_text = """
John Doe
30
john.doe@example.com
"""

result = model(
    f"Extract the person information from this text:\n{person_text}", 
    Person,
    max_new_tokens=100
)
print(json.dumps(result, indent=2))
```

The output type is a JSON string:

```json
{ "name": "John Doe", "age": 30, "email": "john.doe@example.com" }
```

My expectation here would be that `model` return `Person` object, rather than the raw string. This would look like

```python
Person(name='John Doe', age=30, email='john.doe@example.com')
```

To fix this, I currently have to do the (very simple) extra line

```python
person = Person.model_validate_json(result)
```

Is this intended behavior? Outlines < 1.0 would typically return a Pydantic object back.

octo-patch:
Fixes #1526

## Problem

When a Pydantic `BaseModel` subclass is passed as the `output_type` to a `SteerableGenerator`, the generated text is constrained to be valid JSON matching the model's schema — but the return value is still a raw JSON string. Callers have to manually run `MyModel.model_validate_json(result)` after every call, which is a footgun especially for users migrating from Outlines < 1.0 where the Pydantic object was returned directly.

As noted in the issue, the maintainer consensus is to treat Pydantic `BaseModel` subclasses as a special case and auto-parse the result, while leaving all other output types (regex, `Literal`, `CFG`, raw `JsonSchema`, etc.) unchanged as strings.

## Solution

- Store `output_type` on `SteerableGenerator` instances (was previously discarded after building the logits processor). The `from_processor` class-method sets it to `None`.
- Add a `_parse_result` helper that calls `output_type.model_validate_json(result)` when `output_type` is a Pydantic `BaseModel` subclass, and returns the string unchanged for all other types.
- Apply `_parse_result` in `__call__` and `batch`. Streaming is intentionally left unchanged — partial token sequences are not valid JSON and cannot be validated mid-stream.
- Import the existing `is_pydantic_model` utility from `outlines.types.utils` to keep the check consistent with the rest of the codebase.

## Example (before / after)

```python
from outlines import from_transformers, Generator

class Person(BaseModel):
    name: str
    age: int

model = from_transformers(...)
generator = Generator(model, Person)

# Before this fix:
result = generator("Extract: John, 30")
# result == '{"name": "John", "age": 30}'  (str)
person = Person.model_validate_json(result)  # extra step required

# After this fix:
result = generator("Extract: John, 30")
# result == Person(name='John', age=30)  (Person instance)
```

## Testing

- `test_steerable_generator_parse_result_pydantic` — verifies `_parse_result` converts a JSON string to the expected Pydantic model instance.
- `test_steerable_generator_parse_result_non_pydantic` — verifies non-Pydantic output types are returned unchanged.
- `test_steerable_generator_call_pydantic` — end-to-end: calls the generator with a Pydantic output type and checks the return type.
- `test_steerable_generator_init_valid_processor` — updated to assert `output_type is None` after `from_processor`.
- `test_steerable_generator_init_other_output_type` — updated to assert `output_type` is stored correctly.

### DIFF
--- a/outlines/generator.py
+++ b/outlines/generator.py
@@ -23,6 +23,7 @@ from outlines.backends import (
 from outlines.backends.base import LogitsProcessorType
 from outlines.types import CFG, JsonSchema
 from outlines.types.dsl import python_types_to_terms, to_regex
+from outlines.types.utils import is_pydantic_model
 
 
 class BlackBoxGenerator:
@@ -231,6 +232,7 @@ class SteerableGenerator:
 
         """
         self.model = model
+        self.output_type = output_type
         if output_type is None:
             self.logits_processor = None
         else:
@@ -272,9 +274,34 @@ class SteerableGenerator:
         """
         instance = cls.__new__(cls)
         instance.model = model
+        instance.output_type = None
         instance.logits_processor = processor
 
         return instance
+
+    def _parse_result(self, result: Any) -> Any:
+        """Parse a generated string into a Pydantic model instance if applicable.
+
+        When the output type is a Pydantic ``BaseModel`` subclass, the
+        generated JSON string is automatically validated and converted into a
+        model instance.  For all other output types the result is returned
+        unchanged.
+
+        Parameters
+        ----------
+        result
+            The raw string returned by the model.
+
+        Returns
+        -------
+        Any
+            A Pydantic model instance when ``output_type`` is a Pydantic
+            ``BaseModel`` subclass, otherwise the original string.
+
+        """
+        if is_pydantic_model(self.output_type):
+            return self.output_type.model_validate_json(result)
+        return result
 
     def __call__(self, prompt: Any, **inference_kwargs) -> Any:
         """Generate a response from the model.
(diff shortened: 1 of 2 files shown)


RobinPicard:
Hi @octo-patch, thanks for opening a PR but the consensus we reached in the issue was not to include this behavior. I'll you know if that changes in the future.

PR #1862, opened 2026-05-11
Fix error message for model type determination

Reslan-Tinawi:
The `from_transformers` method check the type of the passed `tokenizer_or_processor`, and based on this type, it returns either a `Transformers` or `TransformersMultiModal`.

The code implementation is correct, but in the final else statement, there's a missing negation in the error message presented to the user.

Could be confusing.


RobinPicard (approved):
Good catch, thanks for opening a PR!

Issue #1303, opened 2024-11-30
Add more custom types


rlouf:
Custom types are a very convenient way to abstract regular expressions for very frequent patterns such as phone numbers, zip codes, etc. and we should expand the current collection.

Yeuvoir:
Hi, I’d like to work on this by adding a small set of common regex-based custom types first, starting with semantic version strings and MAC addresses, with tests in tests/types/test_custom_types.py.

sinhasagar507:
Hi @rlouf – I'm Sagar, a student getting into open source. I'd like to add three currently-missing custom types: hex_color (#1a2b3c), slug (my-post-title), and credit_card. (Skipping ipv6 — already in #1867.)

Plan: follow #1863 — add each as a Regex(...) in outlines/types/init.py, with tests in tests/types/test_custom_types.py and a docs entry.

Thanks!

pepepeboom:
Hi! Since the June comment above claiming `hex_color` and `slug` hasn't resulted in a PR in the past five-plus weeks, I've prepared a small PR adding just those two types (following the structure of #1863, with tests). @sinhasagar507 if you're still working on this, happy to defer or coordinate — otherwise I'll submit it shortly.

uditmahato:
Hi! The types claimed earlier in this thread (semver, MAC address, hex_color, slug, credit_card) have all been merged, so I picked three that are still missing: `e164` for canonical ITU-T international phone numbers (complementing the per-country `locale` patterns), and `latitude` / `longitude` for signed decimal-degree coordinates with proper bounds. Same structure and test style as #1863 / #1934 / #1945. PR: #1966

PR #1863, opened 2026-05-17
Add semver and MAC address custom types

Yeuvoir:
## Summary

- Add `semver` as a regex-based custom type for semantic version strings
- Add `mac_address` as a regex-based custom type for colon-separated MAC addresses
- Add tests for valid and invalid examples of both custom types

Refs #1303

## Tests

- `uv run --extra airports --extra countries --with pytest python -m pytest tests/types/test_custom_types.py`


RobinPicard:
Looking good, thanks for opening a PR! Would you mind just adding a comment on top of each type definition to explain the choices made in terms of departure from official specs + a link to them please?

RobinPicard (approved):

PR #1691, opened 2025-07-18
Update docs with OpenRouter guide

Oni-giri:
The goal of this small PR is to provide some awareness and context in the documentation regarding OpenRouter, which is fairly seamless to use with Outlines, given the compatibility with the OpenAI API. It was mentionned in https://github.com/dottxt-ai/outlines/issues/1687

PR #1865, opened 2026-05-29
Update docs with OrcaRouter guide

xilema2:
The goal of this small PR is to add documentation for using OrcaRouter with
Outlines. OrcaRouter is an OpenAI-compatible model router, so it works
seamlessly with the existing `from_openai` loader by simply pointing the
`openai` client at OrcaRouter's base URL — no new model class is required.

This mirrors the existing OpenRouter guide added in #1691: the PR only adds a
`docs/features/models/orcarouter.md` page and a single navigation entry in
`mkdocs.yml`. No code or tests are changed, since OrcaRouter relies entirely on
the already-supported `from_openai` integration.

Disclosure: I'm an engineer on the OrcaRouter team.

### DIFF
--- a/docs/features/models/orcarouter.md
+++ b/docs/features/models/orcarouter.md
@@ -1,0 +1,45 @@
+# OrcaRouter
+
+!!! Installation
+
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):

PR #1870, opened 2026-06-03
Fix unformatted f-string in List type error message

tan7vir:
## What

`_handle_list` in `outlines/types/dsl.py` raises a `TypeError` when a `List`
type is given the wrong number of arguments. The message was built from two
adjacent string literals, and the second one was a plain string rather than an
f-string:

```python
raise TypeError(
    "Only homogeneous lists are supported. You should provide exactly "
    + "one argument to `List`, got {args}."
)
```

So `{args}` was printed literally instead of the actual arguments. For example
`List[int, str]` produced:

```
Only homogeneous lists are supported. You should provide exactly one argument to `List`, got {args}.
```

The sibling `_handle_dict` a few lines down already uses an f-string
(`f"Dict must have exactly two type arguments. Got {args}."`), so this just
brings `_handle_list` in line with it.

## After

```
Only homogeneous lists are supported. You should provide exactly one argument to `List`, got (<class 'int'>, <class 'str'>).
```

## Tests

`test_dsl_handle_list` already exercised this path with a bare
`pytest.raises(TypeError)`, which passed even with the broken message. I
tightened that assertion to `match` the interpolated args so it fails on the
old string and passes on the fix.


RobinPicard:
Thanks for the fix!

PR #1871, opened 2026-06-05
Emit {0,n} for at_most so the regex compiles on the outlines-core backend

lollinng:
### Problem

`Term.at_most(n)` / `at_most(n, term)` (the `QuantifyMaximum` DSL term) compiles to a regex of the form `(...){,n}`:

```python
return f"({to_regex(term.term)}){{,{term.max_count}}}"   # -> "(a){,5}"
```

Python's `re` accepts the lower-bound-omitted `{,n}` form, but the **default** generation backend — `outlines-core` (Rust `regex` / regex-automata) — does **not** (it only supports `{n}`, `{n,}`, `{n,m}`). So any structured generation using an `at_most` quantifier raises when the `Index`/`Guide` is built:

```python
>>> from outlines_core import Index, Vocabulary
>>> v = Vocabulary.from_pretrained("gpt2")
>>> Index("(a){,3}", v)
ValueError: Failed to build DFA error building NFA ...
>>> Index("(a){0,3}", v)   # OK
```

The whole `at_most` quantifier is effectively unusable with the default backend.

### Fix

Emit the equivalent `(...){0,n}`. Semantically identical (0 to n repetitions), but valid for both `re` and `outlines-core`. A unit test asserted the old `(a){,5}` output; updated to `(a){0,5}`.

### Verification

```
to_regex(at_most 5)            -> "(a){0,5}"
outlines_core.Index("(a){0,5}") -> builds OK   (previously "(a){,5}" failed)
pytest tests/types/test_dsl.py -k "to_regex or at_most" -> 2 passed
```

### DIFF
--- a/outlines/types/dsl.py
+++ b/outlines/types/dsl.py
@@ -955,7 +955,7 @@ def to_regex(term: Term) -> str:
     elif isinstance(term, QuantifyMinimum):
         return f"({to_regex(term.term)}){{{term.min_count},}}"
     elif isinstance(term, QuantifyMaximum):
-        return f"({to_regex(term.term)}){{,{term.max_count}}}"
+        return f"({to_regex(term.term)}){{0,{term.max_count}}}"
     elif isinstance(term, QuantifyBetween):
         return f"({to_regex(term.term)}){{{term.min_count},{term.max_count}}}"
     else:
--- a/tests/types/test_dsl.py
+++ b/tests/types/test_dsl.py
@@ -1118,7 +1118,7 @@ def test_to_regex():
     assert to_regex(min_term) == r"(a){2,}"
 
     max_term = QuantifyMaximum(String("a"), 5)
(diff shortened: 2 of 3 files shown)


RobinPicard (approved):
Good catch, thanks!

RobinPicard:
A test in `test_to_regex.py` is failing actually. Can you fix it please?

lollinng:
Fixed in 802114b. The stale assertion was in `tests/types/test_to_regex.py::test_to_regex_simple` (line 76) — it still expected the old `(a){,2}` form. Updated it to `(a){0,2}` to match the new output. (The earlier PR only updated the equivalent assertion in `test_dsl.py`.)

```
- assert to_regex(a) == "(a){,2}"
+ assert to_regex(a) == "(a){0,2}"
```

The other CI failures (ollama multimodal 400s) are unrelated to this change.

PR #1875, opened 2026-06-12
Remove unused Llama tokenizer detection

### LINKED ISSUE #1874: Remove dead code: is_llama and get_llama_tokenizer_types() in TransformerTokenizer

`get_llama_tokenizer_types()` and `self.is_llama` in `TransformerTokenizer` look like dead code - they're set but never read at runtime.

After `1ea2b18`, `convert_token_to_string()` dropped the `if self.is_llama:` guard and always applies the space workaround. The detection machinery was left behind.

```python
import inspect
import transformers

wrapper = TransformerTokenizer(transformers.AutoTokenizer.from_pretrained("gpt2"))
assert "is_llama" not in inspect.getsource(wrapper.convert_token_to_string)

wrapper.is_llama = False
a = wrapper.convert_token_to_string("<0x20>")
wrapper.is_llama = True
b = wrapper.convert_token_to_string("<0x20>")
assert a == b  # flag has no effect
```

### Expected result:

```shell
get_llama_tokenizer_types(), self.is_llama, and related tests. No behavior change.
```

### Error message:

```shell

```

Python 3.11.0

### Context for the issue:

_No response_

palios-taey:
Closes #1874.

`TransformerTokenizer.is_llama` is write-only: `convert_token_to_string()` now always applies the `<0x20>` / SPIECE space workaround and no longer branches on the flag.

This removes the unused Llama-tokenizer detection helper, the unused attribute assignment, and the tests that only exercised that dead path. `get_llama_tokenizer_types()` is not exported in `__all__`, but it is currently included in the generated API reference; happy to keep it as a deprecated compatibility shim if maintainers prefer preserving that import for one release.

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -28,48 +28,6 @@ __all__ = ["Transformers", "TransformersMultiModal", "from_transformers"]
 SPIECE_UNDERLINE = "\u2581"
 
 
-def get_llama_tokenizer_types():
-    """Get all the Llama tokenizer types/classes that need work-arounds.
-
-    When they can't be imported, a dummy class is created.
-
-    """
-    try:
-        from transformers.models.llama import LlamaTokenizer
-    except ImportError:  # pragma: no cover
-
-        class LlamaTokenizer:  # type: ignore
-            pass
-
-    try:
-        from transformers.models.llama import LlamaTokenizerFast
-    except ImportError:  # pragma: no cover
-
-        class LlamaTokenizerFast:  # type: ignore
-            pass
-
-    try:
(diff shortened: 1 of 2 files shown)


palios-taey:
The CI failures here look like infra, not this change. The test jobs fail on `listen tcp 127.0.0.1:11434: bind: address already in use` (the Ollama port) followed by `The operation was canceled`, and the same test jobs are red on `main` right now too. This PR only removes the unused `get_llama_tokenizer_types` workaround and its test, so it shouldn't be touching that path. Happy to rebase once the runners settle.

RobinPicard (approved):
Thanks for the PR! Yes, the failing actions are unrelated

PR #1876, opened 2026-06-15
Remove dead is_llama detection code in TransformerTokenizer

### LINKED ISSUE #1874: Remove dead code: is_llama and get_llama_tokenizer_types() in TransformerTokenizer

joaquinhuigomez:
`convert_token_to_string` applies the `SPIECE_UNDERLINE` / `<0x20>` space workaround unconditionally, so the `is_llama` flag and the `get_llama_tokenizer_types()` helper that computed it are no longer read anywhere. This removes both, deletes the test that exercised the helper, and drops the now-meaningless `is_llama` assignments in `test_transformer_tokenizer_convert_token_to_string` (which still passes). Fixes #1874.

### DIFF
--- a/outlines/models/transformers.py
+++ b/outlines/models/transformers.py
@@ -28,48 +28,6 @@ __all__ = ["Transformers", "TransformersMultiModal", "from_transformers"]
 SPIECE_UNDERLINE = "\u2581"
 
 
-def get_llama_tokenizer_types():
-    """Get all the Llama tokenizer types/classes that need work-arounds.
-
-    When they can't be imported, a dummy class is created.
-
-    """
-    try:
-        from transformers.models.llama import LlamaTokenizer
-    except ImportError:  # pragma: no cover
-
-        class LlamaTokenizer:  # type: ignore
-            pass
-
-    try:
-        from transformers.models.llama import LlamaTokenizerFast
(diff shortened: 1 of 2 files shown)


RobinPicard:
Thanks for opening a PR, but someone opened one (#1875) for the same issue right before, sorry.

PR #1879, opened 2026-06-25
Migrate to src layout

### LINKED ISSUE #1080: Use `src` project layout
Having `outlines` as a top-level directory can create problems with `.so` among other things. We should therefore adopt the `src` layout.

Youssefares:
Moves the `outlines` package from the repo root into `src/outlines/` to prevent import shadowing issues with compiled extensions.

###  Updates
- `pyproject.toml`: add `where = ["src"]` to packages.find, update setuptools_scm write_to path, update coverage omit path
- `.github/workflows/release_pypi.yaml`: update egg-info path

### 🧪 Soundness Checks 
-  Built a wheel and inspected its contents: outlines/ still appears at the root of the built distribution (not src/outlines/), so downstream installs are unaffected:

```
  zipinfo dist/outlines-*.whl | head -3
  # outlines/__init__.py
  # outlines/_version.py
  # outlines/applications.py
```

-   Installed the wheel in a clean venv and verified imports work from outside the repo root:

```
  cd /tmp
  python -c "import outlines.types; print(outlines.types.email)"
```

fix #1080


pyproject.toml
```diff
@@ -189,7 +190,7 @@ ignore_missing_imports = true
 # we omit the files that require a GPU or Apple Silicon
 # as well as the models that make API calls
 omit = [
-    "outlines/_version.py",
+    "src/outlines/_version.py",
     "outlines/models/anthropic.py",
```
RobinPicard: You forgot to update the other paths in this file

RobinPicard (changes requested):
Great, thanks for opening a PR! Just one little thing to change

pyproject.toml
Youssefares: All done 🙌

Youssefares:
> Great, thanks for opening a PR! Just one little thing to change

@RobinPicard Happy to help! Feel free to point me towards any high priority or highly upvoted work. Looking to learn more about the inner workings of this (& outlines-core) 😄.

RobinPicard (approved):
That's great to hear! Thanks for this contribution and don't hesitate to open new PRs if you see issues that need to be tackled.

PR #1880, opened 2026-06-26
fix(dsl): keep None unquoted for Optional nested in containers

sarathfrancis90:
While using `list[Optional[int]]` as an output type I noticed the generated regex matched `[5, "None"]` but rejected `[5, None]`. A standalone `Optional[int]` matches the bare `None` keyword, so the two disagree on how the absent value is written, and the quoted `"None"` parses back to the 4-character string rather than `None`.

The cause is in `_handle_union`: the `None` arm of an `Optional`/`Union` is emitted as `String("None")`. The JSON-quoting helper added for string literals inside container types (`_ensure_json_quoted`) then wraps every `String` in double quotes, this one included. `True`/`False` escape it because they're already `Regex` terms, which the helper leaves untouched.

I made `None` a `Regex("None")` keyword token like the booleans, so it stays unquoted inside `List`/`Dict`/`Tuple` while a genuine `Literal["None"]` string is still quoted. Added e2e regex tests for `Optional` nested in each container and updated the one structural assertion in `test_dsl_handle_union`.

`pytest tests/types` and `pre-commit run` pass locally.

### DIFF
--- a/outlines/types/dsl.py
+++ b/outlines/types/dsl.py
@@ -837,7 +837,10 @@ def _handle_union(args: tuple, recursion_depth: int) -> Alternatives:
         return Alternatives(
             [
                 python_types_to_terms(other_ptype, recursion_depth + 1),
-                String("None"),
+                # ``None`` is a keyword, not a string value: keep it a ``Regex``
+                # (like ``True``/``False``) so it is not JSON-quoted when the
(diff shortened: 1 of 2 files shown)


RobinPicard:
Thanks for fixing this problem!

PR #1881, opened 2026-06-26
Fix crash on a union of >=2 non-None members plus None

iamsharduld:
## What

`python_types_to_terms` (via `_handle_union`) only special-cases the **2-arg** `Optional[T]` form
(one non-None member + `None`). Any union with **two or more non-None members plus `None`** falls into
the general branch, which recurses into `NoneType` — which has no handler — and crashes:

```python
from typing import Union, Optional

python_types_to_terms(Optional[int])              # OK
python_types_to_terms(Union[int, str, None])      # TypeError: Type NoneType is currently not supported
python_types_to_terms(Optional[Union[int, str]])  # TypeError
# also Dict[str, Union[int, str, None]], List[Optional[Union[int, str]]], ...
```

These are common annotations and the path is reachable from the public `Generator` entry point
(`generator.py` calls `python_types_to_terms(output_type)` on the user's type).

## Fix

Generalize `_handle_union` to peel off the `None` member(s) for **any** arity and map them to
`String("None")`, preserving the existing 2-arg `Optional[T]` semantics:

```python
has_none = type(None) in args or None in args
terms = [python_types_to_terms(a, d+1) for a in args if a not in (type(None), None)]
if has_none:
    terms.append(String("None"))
return Alternatives(terms)
```

So `Union[int, str, None]` → `Alternatives([integer, string, String("None")])`.

> Note: PR #1880 separately changes the 2-arg `None` token from `String("None")` to `Regex("None")`
> (container-quoting fix). It doesn't touch this >=2-member case, so there's no overlap; if it lands
> first, the `None` token here should be switched to match.

## Tests

`tests/types/test_dsl.py::test_dsl_handle_union_multiple_members_with_none` — `Union[int, str, None]`
and `Optional[Union[int, str]]` no longer crash and include the `None` arm. Existing union/dsl/to_regex
tests still pass.

### DIFF
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -835,21 +835,22 @@ def _ensure_json_quoted(term: Term) -> Term:
 
 
 def _handle_union(args: tuple, recursion_depth: int) -> Alternatives:
-    # Handle the Optional[T] type
-    if len(args) == 2 and (type(None) in args or None in args):
-        other_ptype = next(arg for arg in args if arg not in (type(None), None))
-        return Alternatives(
-            [
-                python_types_to_terms(other_ptype, recursion_depth + 1),
-                # ``None`` is a keyword, not a string value: keep it a ``Regex``
-                # (like ``True``/``False``) so it is not JSON-quoted when the
-                # ``Optional`` ends up nested inside a container type.
-                Regex("None"),
-            ]
-        )
-    return Alternatives(
-        [python_types_to_terms(arg, recursion_depth + 1) for arg in args]
-    )
+    # Peel off the None member(s) for any arity (not just the 2-arg Optional[T]
+    # case): a union such as `Union[int, str, None]` / `Optional[Union[int, str]]`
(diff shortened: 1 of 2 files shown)


RobinPicard:
Thanks for opening a PR! Could you take care of resolving the conflicts with `main` please?

iamsharduld:
Thanks for the review — sorry about the conflict! Merged `main` and resolved it.

The clash was in `_handle_union`, overlapping the recent "keep None unquoted in containers" fix. I composed the two: still peel `None` off a union of any arity (fixes the `Union[int, str, None]` crash), but emit it as `Regex("None")` instead of `String("None")` so it stays unquoted inside containers, like the 2-arg `Optional[T]` path.

Updated the test to match and added a small container check; full `tests/types/test_dsl.py` passes locally. Happy to adjust if you'd prefer another approach.

RobinPicard (approved):
I took care of the rebase, things are looking good. Thanks a lot for the PR!

PR #1883, opened 2026-06-26
Mark non-required JSON-schema properties NotRequired in TypedDict conversion

iamsharduld:
## What

`json_schema_dict_to_typeddict` wraps non-`required` properties in `Optional[T]` on a `total=True`
TypedDict. But `Optional` only makes the **value** nullable — the **key** stays required. So a schema
where `age` is optional round-trips back as required, unlike the pydantic/dataclass converters:

```python
schema = {"type":"object","properties":{"name":{"type":"string"},"age":{"type":"integer"}},"required":["name"]}
td = json_schema_dict_to_typeddict(schema, "Person")
td.__required_keys__   # {'name', 'age'}   -- age should be optional
# json_schema_dict_to_pydantic / _to_dataclass on the same schema -> required == ['name']
```

## Fix

Use `NotRequired[T]` (PEP 655) for non-required properties so the **key** is optional, matching the
input schema and the sibling converters.

## Tests

`tests/types/test_json_schema_utils.py::test_json_schema_dict_to_typeddict_basic` updated to assert
`age` is `NotRequired[int]` and lands in `__optional_keys__` (it previously codified the buggy
`Optional[int]`/required behavior). Full `test_json_schema_utils.py` passes.

### DIFF
--- a/outlines/types/json_schema_utils.py
+++ b/outlines/types/json_schema_utils.py
@@ -7,9 +7,9 @@ from typing import Any, Dict, List, Literal, Optional
 from pydantic import BaseModel, create_model
 
 if sys.version_info >= (3, 12):  # pragma: no cover
-    from typing import _TypedDictMeta, TypedDict  # type: ignore
+    from typing import _TypedDictMeta, NotRequired, TypedDict  # type: ignore
 else:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):
Great, thanks!

PR #1886, opened 2026-06-29
Handle enum refs in prompt schema rendering

### LINKED ISSUE #229: Prompting: response_model | schema does not work with Enum
```
class Armor(str, Enum):
    leather = "leather"
    chainmail = "chainmail"
    plate = "plate"

class Character(BaseModel):
    """
    A character in a fantasy world.
    """
    name: constr(max_length=12)
    armor: Armor
    strength: int

@text.prompt
def ppt(response_model):
    """Make A character in a fantasy world.

    RESPONSE FORMAT:
    {{ response_model | schema }}
    """


print(ppt(Character))
```

Get error `KeyError: 'properties'`

If remove `armor` from `Character` this works fine

lxingy3:
## Summary

- Render referenced enum schemas in the `schema` template filter instead of recursing into non-object schemas.
- Keep the existing prompt-example rendering for descriptions, nested models, and regular fields.
- Add regression coverage for Pydantic models with Enum fields.

Closes #229

## Tests

- `python -m pytest tests/test_templates.py -q`
- `python -m ruff check outlines/templates.py tests/test_templates.py --config=pyproject.toml`
- `git diff --check`

Note: `pre-commit run --files outlines/templates.py tests/test_templates.py` could not complete locally because fetching `pre-commit/pre-commit-hooks` failed with a TLS handshake error.

### DIFF
--- a/src/outlines/templates.py
+++ b/src/outlines/templates.py
@@ -306,16 +306,27 @@ def parse_pydantic_schema(raw_schema, definitions):
     key in the schema of the top-level model.
 
     """
+    if "properties" not in raw_schema:
+        return parse_pydantic_schema_value(
+            raw_schema.get("title", "value"), raw_schema, definitions
+        )
+
     simple_schema = {}
     for name, value in raw_schema["properties"].items():
-        if "description" in value:
-            simple_schema[name] = value["description"]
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):
Thanks for opening a PR!

PR #1889, opened 2026-06-30
fix(types): convert JSON Schema type arrays to a Union

sarathfrancis90:
JSON Schema lets `type` be an array of type names — the canonical way to express a nullable field is `{"type": ["string", "null"]}`. `schema_type_to_python` only matched `type` against a single string, so a type array fell through to the `Any` fallback. When a raw JSON schema is converted to a pydantic model / TypedDict / dataclass (the path the API model backends use, e.g. Gemini), this silently dropped the constraint: an `["integer", "null"]` field came out as `Any` instead of `Optional[int]`.

I map each member of a type array to its Python type and combine them with `Union`, and added handling for the `"null"` type so it resolves to `NoneType`. That lines up with the `anyOf` the regex backend already uses for type arrays.

Added tests covering nullable and multi-type arrays, the standalone `null` type, and a single-element array. `pytest tests/types/` and `pre-commit run --all-files` are green.

### DIFF
--- a/outlines/types/json_schema_utils.py
+++ b/outlines/types/json_schema_utils.py
@@ -2,7 +2,7 @@
 
 import sys
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from pydantic import BaseModel, create_model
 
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):

PR #1894, opened 2026-07-01
fix(types): map JSON Schema const to a Literal

sarathfrancis90:
`schema_type_to_python` maps `enum` to a `Literal`, but its singular sibling `const` isn't handled at all, so a `const` field silently falls through to `Any` — or, when a `type` keyword is present too, widens to that bare type.

This bites the schema round-trip. Pydantic serialises a one-element `Literal` as `{"const": <value>, "type": <name>}`, so converting such a model to a typeddict/dataclass drops the fixed value:

```python
from typing import Literal

class Event(BaseModel):
    kind: Literal["created"]
    id: int

JsonSchema.convert_to(Event, ["typeddict"]).__annotations__["kind"]
# str  -- expected Literal["created"]
```

I map `const` to `Literal[value]`, placed before the `type` handling so `const` wins when both are present (which is the shape Pydantic emits). Added unit + end-to-end tests.

`pre-commit` (mypy, ruff) clean and `pytest tests/types` passes.

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -35,6 +35,13 @@ def schema_type_to_python(
         values = schema["enum"]
         return Literal[tuple(values)]
 
+    if "const" in schema:
+        # ``const`` pins the field to a single value (the singular sibling of
+        # ``enum``). Pydantic emits it for a one-element ``Literal``, often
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):
Good fix, thanks!

PR #1897, opened 2026-07-01
fix: reject invalid day '00' in the date type regex

hoangsonww:
## What this fixes

`outlines.types.date` accepted an impossible day value of `00`. For example, `2022-03-00` matched the `date` regex and was treated as a valid date.

The day component was defined as `[0-2][0-9]|3[0-1]`, which matches `00` through `29` plus `30`/`31`. Since Outlines constrains generation to the given format, this allowed a model to emit invalid dates such as `2022-03-00`.

## The fix

Tighten the day group to `0[1-9]|[12][0-9]|3[01]`, which matches only `01` through `31`.

- Before: `(\d{4})-(0[1-9]|1[0-2])-([0-2][0-9]|3[0-1])`
- After: `(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])`

Every previously valid date still matches; only impossible days (`00`) are now rejected. Full calendar validation (per-month lengths and leap years) is intentionally left out of scope for this format regex, consistent with the existing behavior for values like `2022-02-31`.

## Tests

Added cases to `test_type_regex` in `tests/types/test_custom_types.py`:

- `2022-03-01` -> valid
- `2022-03-31` -> valid
- `2022-03-00` -> invalid (the bug)
- `2022-00-23` -> invalid (locks in the month lower bound)

Before the change, `2022-03-00` incorrectly passed. After the change the full `test_type_regex` suite passes (110 cases).


RobinPicard (approved):

ijatmamuhome-ux:
> ## What this fixes
> 
> `outlines.types.date` accepted an impossible day value of `00`. For example, `2022-03-00` matched the `date` regex and was treated as a valid date.
> 
> The day component was defined as `[0-2][0-9]|3[0-1]`, which matches `00` through `29` plus `30`/`31`. Since Outlines constrains generation to the given format, this allowed a model to emit invalid dates such as `2022-03-00`.
> 
> ## The fix
> 
> Tighten the day group to `0[1-9]|[12][0-9]|3[01]`, which matches only `01` through `31`.
> 
> - Before: `(\d{4})-(0[1-9]|1[0-2])-([0-2][0-9]|3[0-1])`
> - After: `(\d{4})-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01])`
> 
> Every previously valid date still matches; only impossible days (`00`) are now rejected. Full calendar validation (per-month lengths and leap years) is intentionally left out of scope for this format regex, consistent with the existing behavior for values like `2022-02-31`.
> 
> ## Tests
> 
> Added cases to `test_type_regex` in `tests/types/test_custom_types.py`:
> 
> - `2022-03-01` -> valid
> - `2022-03-31` -> valid
> - `2022-03-00` -> invalid (the bug)
> - `2022-00-23` -> invalid (locks in the month lower bound)
> 
> Before the change, `2022-03-00` incorrectly passed. After the change the full `test_type_regex` suite passes (110 cases).
>

PR #1900, opened 2026-07-03
Normalize LM Studio exceptions

Dikshj:
# Thank you for opening a PR!

A few important guidelines and requirements before we can merge your PR:

- [x] We should be able to understand what the PR does from its title only;
- [x] There is a high-level description of the changes;
- [x] *If I add a new feature*, there is an issue discussing it already;
- [x] There are links to *all* the relevant issues, discussions and PRs;
- [x] The branch is rebased on the latest `main` commit;
- [x] **Commit messages** follow these guidelines;
- [x] One commit per logical change;
- [x] The code respects the current **naming conventions**;
- [x] Docstrings follow the numpy style guide;
- [x] `pre-commit` is installed and configured on your machine, and you ran it before opening the PR;
- [x] There are tests covering the changes;
- [x] The documentation is up-to-date;

## Description

Follow-up to #1823 / #1658.

#1823 added the shared exception hierarchy and provider error normalization for most client-backed model wrappers. After that merge, LM Studio still appears to be missing from the provider normalization path: runtime transport failures can escape as raw exceptions from the LM Studio client.

This PR adds LM Studio to the shared provider exception mapping and wraps LM Studio sync and async client calls with `normalize_provider_errors("lmstudio")`.

## Changes

- Map LM Studio `ConnectionError` failures to `APIConnectionError`.
- Map LM Studio `TimeoutError` failures to `APITimeoutError`.
- Continue using status-code fallback for provider errors that expose HTTP-like status codes, including rate limits.
- Wrap sync and async LM Studio model lookup, generation, and streaming calls.
- Preserve the original provider exception through `original_exception` and Python exception chaining.
- Keep local validation and programmer errors, such as `ValueError`, passing through unchanged.

## Compatibility

This keeps the existing LM Studio public API unchanged. It only standardizes provider/runtime failures under the existing Outlines exception hierarchy, while preserving the underlying exception for debugging.

Streaming is covered for both provider failures and early consumer cancellation. Since `normalize_provider_errors` catches `Exception`, `GeneratorExit` and other `BaseException` subclasses are not wrapped when a caller stops consuming a stream early.

## Verification

- `pre-commit run --all-files`
- `python -m pytest tests/test_exceptions.py -k lmstudio`
- `python -m pytest tests/models/test_lmstudio.py`

### DIFF
--- a/src/outlines/exceptions.py
+++ b/src/outlines/exceptions.py
@@ -358,6 +358,12 @@ def _provider_exception_map(provider: str) -> dict[type, type[APIError]]:
             urllib3_exc.MaxRetryError: APIConnectionError,
         }
 
+    if provider == "lmstudio":
+        return {
+            ConnectionError: APIConnectionError,
+            TimeoutError: APITimeoutError,
+        }
+
     # Unknown provider: no SDK-specific mapping available.
     # normalize_provider_exception will fall back to status-code inspection.
     return {}
--- a/src/outlines/models/lmstudio.py
+++ b/src/outlines/models/lmstudio.py
@@ -1,6 +1,6 @@
 """Integration with the `lmstudio` library.
 
-Local runtime calls intentionally bypass
+Provider/runtime calls are normalized via
 outlines.exceptions.normalize_provider_errors().
 """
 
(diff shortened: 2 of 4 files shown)


Dikshj:
This is a follow-up to #1823.

#1823 added the shared exception hierarchy and provider error normalization across Outlines, resolving the broader #1658 work. After that merge, LM Studio still appears to be missing from the provider normalization path: runtime transport failures can escape as raw ConnectionError / TimeoutError from the LM Studio client.

This PR does not replace #1823. It fills that LM Studio-specific gap by adding lmstudio to the existing provider exception map and wrapping LM Studio sync/async generation and streaming calls with normalize_provider_errors(" lmstudio\).

RobinPicard (changes requested):
That's an interesting idea, but we need actual sdk error mapping to make it work and be useful. Right now the `lmstudio` handling in `src/outlines/models/lmstudio.py` is too light as it only covers generic Python errors.

PR #1903, opened 2026-07-05
feat(mlxlm): support vision models via mlx-vlm processors

### LINKED ISSUE #1188: Consideration of support for mlx-vlm
### Presentation of the new feature

Similarly to mlx-lm, [mlx-vlm](https://github.com/Blaizzy/mlx-vlm) allows local Metal execution of MLX vision models.  The project is used quite extensively and is growing quickly.

Just like with mlx-vlm, just using the Torch 'mps' backend is impractical due to hugely inefficient execution, so everyone running a vision model locally on Metal uses this library.

### Where does it fit in Outlines?

This fits as a new integration, just like mlx-lm.  I believe due to the high degree of similarity to the mlx-lm implementation, mlx-vlm will be less difficult to integrate than a new framework!

pablopupo:
Closes #1188.

This adds vision-model support through `mlx-vlm` while preserving the existing text-only `MLXLM` path.

`from_mlxlm` now dispatches on its second argument:

- an MLX or Transformers tokenizer returns `MLXLM`;
- a Transformers processor returns the new `MLXLMMultiModal` wrapper.

The multimodal wrapper follows the existing `TransformersMultiModal` split. It accepts `Chat` inputs with `Image` assets, formats them with the processor's chat template, and uses `mlx_vlm.generate` or `mlx_vlm.stream_generate`. Plain string inputs, constrained generation, and streaming are supported. Vision batching raises `NotImplementedError` because the current `mlx-vlm` batching API does not fit Outlines' existing `generate_batch` contract.

Both wrappers expose the underlying Hugging Face tokenizer through `hf_tokenizer`, so the outlines-core, XGrammar, and llguidance backends can build their vocabularies without depending on whether the loader returned a tokenizer or processor.

`mlx-vlm` is included in the existing `mlxlm` extra so the loader has one installation path for text and vision models.

Verification:

- 22 MLX-LM and MLX-VLM tests pass.
- An Apple Silicon smoke test with `mlx-community/SmolVLM-256M-Instruct-4bit` covered unconstrained, regex-constrained, JSON-schema, and streaming generation on a real image.
- Pre-commit passes.

### DIFF
--- a/docs/features/models/mlxlm.md
+++ b/docs/features/models/mlxlm.md
@@ -10,14 +10,14 @@ Outlines provides an integration with [mlx-lm](https://github.com/ml-explore/mlx
 
     You need a device that [supports Metal](https://support.apple.com/en-us/102894) to use the mlx-lm integration.
 
-    You need to install the `mlx` and `mlx-lm` libraries to be able to use mlx in Outlines. Install all optional dependencies of the `MLXLM` model with: `pip install "outlines[mlxlm]"`.
+    You need to install the `mlx` and `mlx-lm` libraries to be able to use mlx in Outlines. Vision models additionally require `mlx-vlm`. Install all optional dependencies of the `MLXLM` model with: `pip install "outlines[mlxlm]"`.
 
 ## Model Initialization
 
 To create a MLXLM model instance, you can use the `from_mlxlm` function. It takes 2 arguments:
 
 - `model`: an `mlx.nn.Module` instance
-- `tokenizer`: a `transformers.PreTrainedTokenizer` instance
+- `tokenizer`: a `transformers.PreTrainedTokenizer` instance, or a `transformers` processor for vision models (see [Vision Models](#vision-models))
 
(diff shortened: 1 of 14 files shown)


RobinPicard:
Thanks a lot for picking up this old issue! I know it mentioned having a single class, but thinking more about I believe it would be better to follow the same approach as for the `Transformers` model with 2 classes and a dispatch in the loader function. Sorry for the change, would you mind implementing it? Otherwise things look really good!

pablopupo:
Done, no worries. It now mirrors the Transformers setup exactly. `MLXLMMultiModal` inherits from `MLXLM` with its own type adapter, and `from_mlxlm` dispatches on tokenizer vs processor the same way `from_transformers` does. The backends read the tokenizer from a shared `hf_tokenizer` attribute that both classes populate, so they didn't need any special casing there anymore. Re-ran the mlxlm tests and the SmolVLM smoke test on the new structure, all green.

pablopupo:
The requested text/multimodal split is in place, and the focused MLX-LM tests are green. @RobinPicard, could you take another look when you have a chance?

PR #1906, opened 2026-07-06
Stop omitting lmstudio.py from the coverage report

### LINKED ISSUE #1506: Handle skipped tests with coverage
We use the library `coverage` to check our test coverage in the CI. An issue we encountered is that skipped tests (for instance those related to `mlx`) are not taken into account such that the test fails even though we actually have a test for a given set of lines. A temporary solution is to exclude the files concerned from what `coverage` checks, but it's not the best. We should find a way of having `coverage` take into account those skipped tests.

ErenAta16:
Fixes #1506

`src/outlines/models/lmstudio.py` was in the coverage `omit` list because its image-handling paths need a live LM Studio server (`lms.prepare_image` uploads bytes to a running instance), and `from_lmstudio()`'s success branches / `AsyncLMStudio.close()` were never reached by any test that runs without one.

None of that actually requires real hardware or an API key, unlike the GPU/Apple Silicon files also in that list, so this closes the gaps with mocked tests instead:

- Un-skip the two image-input tests in `test_lmstudio_type_adapter.py`by stubbing `lms.prepare_image`, since everything else in those tests only checks the adapter builds the right `lmstudio.Chat` structure. Also added the matching "invalid image asset" test for chat-formatted input, mirroring the one that already existed for list-formatted input.

- Added tests for `AsyncLMStudio.close()`, the context re-entry skip in `generate_stream`, and `from_lmstudio()`'s dispatch for a plain `Client`/`AsyncClient` (both of which can be constructed against a bogus host without connecting, so no real server or mock is needed).
- Removed `src/outlines/models/lmstudio.py` from `[tool.coverage.run] omit` now that it hits 100% branch coverage on its own.
Verified locally with a coverage config that has no omit list: `src/outlines/models/lmstudio.py: 145 stmts, 54 branches, 100% cover`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -194,7 +194,6 @@ omit = [
     "src/outlines/models/anthropic.py",
     "src/outlines/models/dottxt.py",
     "src/outlines/models/gemini.py",
-    "src/outlines/models/lmstudio.py",
     "src/outlines/models/mlxlm.py",
     "src/outlines/models/openai.py",
     "src/outlines/models/mistral.py",
--- a/tests/models/test_lmstudio.py
+++ b/tests/models/test_lmstudio.py
@@ -161,6 +161,15 @@ def test_lmstudio_init_from_client():
         assert model.client == client
         assert model.model_name is None
 
+    # `from_lmstudio` must dispatch to `LMStudio` for a sync client regardless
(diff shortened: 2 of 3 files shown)


RobinPicard (approved):
Nice addition, thanks!

PR #1907, opened 2026-07-06
Docs/fix reorg broken links

ritsth:
The models and guide reorg (pages moved into `features/models/` and `guide/`)
left a bunch of relative links pointing at the old paths. `mkdocs build`
doesn't run with `--strict`, so these just log warnings instead of failing CI.

Fixed ~40 links across 7 files — mostly wrong `../` depth after the move, plus
a couple of absolute links (`/installation`, `/user_guide/migration`) and one
filename mismatch (`templates.md` vs the actual `template.md`).

Verified by rebuilding the docs and checking the mkdocs warning output before
and after, the targeted warnings are gone. Also spot-checked a few of the
ambiguous ones against the live site to make sure I was pointing at the right
page.

Used an AI assistant to find and fix these; reviewed and tested every change myself.

### DIFF
--- a/docs/community/contribute.md
+++ b/docs/community/contribute.md
@@ -66,7 +66,7 @@ Outlines provides optional dependencies for different supported backends, which 
 pip install ".[vllm]"
 ```
 
-A list of supported optional dependencies can be found in the [installation guide](/installation).
+A list of supported optional dependencies can be found in the [installation guide](../guide/installation.md).
 
 ### Using VSCode DevContainer / GitHub Codespaces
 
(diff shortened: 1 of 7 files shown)


RobinPicard:
The documentation fix looks good, but you've added another unrelated commit to the branch!

ritsth:
Oh, my mistake, I will fix that.

ritsth:
Fixed. That was a stray commit from an earlier squash-merged PR. Rebased onto the current main and resolved a real conflict with the LM Studio docs that landed in the meantime; should just be the link fixes now.

RobinPicard (approved):
Awesome, thanks!

PR #1909, opened 2026-07-09
fix: only set additionalProperties on object schemas

vidigoat:
## What this fixes

`set_additional_properties_false_json_schema` (used by the OpenAI backend to make a schema `strict`-compatible) sometimes injects `"additionalProperties": false` into **non-object** schemas. The OpenAI structured-output API rejects `additionalProperties` on anything that isn't an object, so a legitimate schema fails with a `400 invalid_request_error`.

## Reproduction

```python
from typing import Literal
from outlines.models.openai import OpenAITypeAdapter

class Node(BaseModel):
    name: str
    kind: Literal["object"]   # pydantic emits {"type": "string", "const": "object"}

schema = OpenAITypeAdapter().format_output_type(Node)
kind = schema["response_format"]["json_schema"]["schema"]["properties"]["kind"]
print(kind)
# -> {'const': 'object', 'title': 'Kind', 'type': 'string', 'additionalProperties': False}
#                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ invalid on a string field
```

Any scalar value equal to the string `"object"` triggers it — a `Literal["object"]` field (emitted as `const`), or a field whose `default` / `title` / `description` is `"object"`.

## Root cause

```python
jsonpath_expr = jsonpath_ng.parse('$..*')
for match in jsonpath_expr.find(schema):
    if match.value == 'object':                       # matches ANY value == "object"
        if 'additionalProperties' not in match.context.value:
            match.context.value['additionalProperties'] = False  # mutates the parent
```

It keys off the string *value* `"object"` appearing anywhere in the schema and then mutates that node's parent, instead of keying off the JSON Schema keyword `{"type": "object"}`.

## Fix

Recurse over the schema and set `additionalProperties` only on dicts whose `type` is `"object"`. Nested object schemas are still handled; already-present `additionalProperties` values are preserved. This also drops the `jsonpath_ng` usage in this helper in favour of a plain recursive walk.

## Test evidence

Added two regression tests in `tests/models/test_utils.py`:
- `test_set_additional_properties_false_json_schema_string_value_object` — fails on `main`, passes with the fix (a string field with `const`/`title` `"object"` must not get `additionalProperties`).
- `test_set_additional_properties_false_json_schema_nested_objects` — documents that nested object schemas still get `additionalProperties`.

```
$ pytest tests/models/test_utils.py tests/models/test_openai_type_adapter.py -q
16 passed
```

`ruff check` and `mypy --allow-redefinition` both pass on the changed files.

### DIFF
--- a/src/outlines/models/utils.py
+++ b/src/outlines/models/utils.py
@@ -1,8 +1,10 @@
-import jsonpath_ng
-
-
 def set_additional_properties_false_json_schema(schema: dict) -> dict:
-    """Set additionalProperties to False to all objects in the schema using jsonpath.
+    """Set additionalProperties to False on all object schemas.
+
+    Recursively walks the JSON schema and, for every object schema, sets
+    ``additionalProperties`` to False unless it is already present. An object
+    schema is one whose ``type`` is ``"object"`` or, for a nullable object, a
+    list containing ``"object"`` (e.g. ``["object", "null"]``).
 
     Parameters
     ----------
@@ -14,20 +16,26 @@ def set_additional_properties_false_json_schema(schema: dict) -> dict:
     dict
         The modified schema with additionalProperties set to False
     """
-    # Get all nodes
-    jsonpath_expr = jsonpath_ng.parse('$..*')
-    matches = jsonpath_expr.find(schema)
 
-    # Go over all nodes and set additionalProperties to False if it's an
-    # object. `type` can either be the bare string "object" or, per the JSON
-    # Schema spec, a list of type names (e.g. ["object", "null"]) used to
-    # express nullable objects, so both forms must be checked.
-    for match in matches:
-        is_object_type = match.value == 'object' or (
-            isinstance(match.value, list) and 'object' in match.value
-        )
-        if is_object_type:
-            if 'additionalProperties' not in match.context.value:
-                match.context.value['additionalProperties'] = False
+    def _walk(node):
+        if isinstance(node, dict):
+            # Only an object *schema* should get ``additionalProperties``. Keying
+            # off the string ``"object"`` appearing anywhere would wrongly add the
+            # keyword to non-object schemas, e.g. a string field with
+            # ``const``/``default``/``title`` equal to ``"object"``, which the
+            # OpenAI API then rejects. ``type`` may be the bare string or a list
+            # such as ``["object", "null"]`` for a nullable object.
+            node_type = node.get("type")
+            is_object = node_type == "object" or (
+                isinstance(node_type, list) and "object" in node_type
+            )
+            if is_object and "additionalProperties" not in node:
+                node["additionalProperties"] = False
+            for value in node.values():
+                _walk(value)
+        elif isinstance(node, list):
+            for item in node:
+                _walk(item)
 
+    _walk(schema)
     return schema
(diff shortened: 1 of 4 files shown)


RobinPicard (commented):
Thanks for opening a PR! There's a conflict to solve with the main `main` as the file you are modifying has been updated. Let's remove the `jsonpath_ng` dependency too as you are removing the only use of this library.

src/outlines/models/utils.py
```diff
@@ -14,14 +15,21 @@ def set_additional_properties_false_json_schema(schema: dict) -> dict:
...
-            if 'additionalProperties' not in match.context.value:
-                match.context.value['additionalProperties'] = False
+    def _walk(node):
+        if isinstance(node, dict):
+            # Only an object *schema* (``"type": "object"``) should get
+            # ``additionalProperties``. Keying off the string value ``"object"``
+            # appearing anywhere would wrongly add the keyword to non-object
+            # schemas, e.g. a string field with ``const``/``default``/``title``
+            # equal to ``"object"``, which the OpenAI API then rejects.
+            if node.get("type") == "object" and "additionalProperties" not in node:
```
Sanjays2402: This misses nullable object schemas whose `type` is `["object", "null"]`. Those nested objects keep no `additionalProperties`, so the strict OpenAI/Mistral schema is rejected; the equivalent check needs to handle lists containing `"object"` too.

vidigoat:
Thanks @RobinPicard! Rebased onto main so the conflict is resolved, and removed the `jsonpath_ng` dependency (from both `pyproject.toml` and `uv.lock`) since this was its only use — `uv lock --check` passes.

@Sanjays2402 good catch on nullable objects — the check now also matches `type` lists containing `"object"` (e.g. `["object", "null"]`), so nested nullable object schemas get `additionalProperties: false` too. Added a regression test for that case alongside the existing ones.

ErenAta16 (commented):
Reproduced the old bug directly before checking the fix, since the whole thing hinges on whether the old jsonpath walk really did hit non-object schemas.

It does. Running the old `$..*` version against the PR's own repro (`{"type": "string", "const": "object", "title": "object"}` nested under an object) adds `additionalProperties: False` straight onto the string schema, because `$..*` visits every value in the tree, `"object"` shows up as the value of `const` and `title`, and `match.context.value` is then the string schema itself. So the fix is addressing a real 400 from the OpenAI side, not a theoretical one.

The rewritten `_walk` keys off `node.get("type")` instead of any value that happens to equal `"object"`, which is the correct signal. Checked it against a few cases beyond the one in the test:

- The repro's string-with-`const: "object"` field: correctly left untouched.
- Plain nested object: still gets `additionalProperties: False`.
- Nullable object with `"type": ["object", "null"]`: still gets it, the list branch is preserved.
- A property literally named `object` with a string type: correctly left untouched (the key is walked as a value, has no matching `type`, so nothing is added).

One thing worth noting for whoever merges: `_walk` mutates `schema` in place and returns the same object, same as the old version did, so behavior for existing callers is unchanged there. Dropping `jsonpath_ng` entirely is a nice side benefit given this was its only use.

Looks correct and well-scoped to me.

RobinPicard (approved):
Thanks!

PR #1910, opened 2026-07-09
Add FunctionCallingLogitsProcessor for tool call support

### LINKED ISSUE #1626: Add function calling and MCP support
A v0 would be allowing users to instantiate the logits processor with the tool call tokens and pass it to the generator. Note that this would only work with models that use simple tokens for tool calling (Llama has a more complex pattern that we can handle later).

```python
model = ...
logits_processor = FunctionCallingLogitsProcessor("<tool_call>", "</tool_call>",  Union[User, Customer])
generator = Generator(model, processor=logits_processor)
```

The implementation is simple:
1. While `<tool_call>` has not been observed, `__call__` is a pass-through.
2. After `<tool_call>` has been observed, mask the logits to follow a given JSON Schema. 
3. Once the JSON generation has completed, force the generation of `</tool_call>`
4. Then let the model generate freely, `__call__` is a pass-through.

If there are several possible function calls, the type for (2) would be a `Union` of the different JSON Schemas.

Youssefares:
## Summary
Adds a `FunctionCallingLogitsProcessor` that switches between free generation and constrained JSON generation around tool call delimiters, implementing the v0 design from #1626.

Refactors get_logits_processor to reuse for compiling tools_type.

### How it works
Each sequence in the batch independently transitions through three phases:

1. **PASSTHROUGH** - logits returned unmodified. Scans for the open delimiter token.
2. **ARGUMENTS** - delegates to an inner `OutlinesCoreLogitsProcessor` that constrains generation to the JSON schema. Checks `guide.is_finished()` after each step.
3. **CLOSING** - masks all tokens to `-inf` except the close delimiter.

After closing, returns to PASSTHROUGH — supports multiple tool calls per generation.

### Implementation details

- Compiles the JSON schema into an FSM `Index` once at construction. Index 1:* Batch but Guide 1:1 Batch.
- Slices `input_ids` so the inner processor only sees tokens generated since entering ARGUMENTS, not the passthrough prefix or open delimiter.
- Cross-library token masking via `_force_single_token_row` handles numpy, torch, and MLX (MLX arrays are immutable, so standard item assignment doesn't work).

### Tests

- 20 unit tests parametrized across numpy, torch, and MLX 
- 1 integration test with a real `OutlinesCoreLogitsProcessor`, GPT-2 tokenizer, and simulated generation loop: verifies the output between delimiters is valid JSON matching the schema.

fix #1626


RobinPicard:
Hi @Youssefares! Thanks for looking into this topic. I think it's something for which we still need to do a bit more thinking to ensure we use an interface that would work for various tool calling format though. We want to avoid the situation in which we have to break the interface first introduced because it's not exhaustive enough.

Youssefares:
> Hi @Youssefares! Thanks for looking into this topic. I think it's something for which we still need to do a bit more thinking to ensure we use an interface that would work for various tool calling format though. We want to avoid the situation in which we have to break the interface first introduced because it's not exhaustive enough.

Thanks, Robin. Makes sense! Would it be useful to look into a few formats we're looking to support? and spec'ing out the different possible interfaces. If so, what formats are they? I can look into it and update the issue with details if you tell me 3-4 names that you believe will give us good breadth 🙌

RobinPicard:
Thanks for proposing! We ideally want an approach that could accommodate the best open source models such as those from `DeepSeek`, `Kimi`, `Z.ai` and `Qwen`. They are annoying to work with as some use strange formats with XML-like key-value tags system, multiple nested tags... But in general as a disclaimer I cannot tell you when we'll actually merge anything on that topic. Experimenting is good though, so if you feel like opening draft PRs to showcase different approaches, that would be valuable.

Youssefares:
Np! Appreciate the openness. Sounds good! As a way to leave this in a semi-mergeable state, I will have look at the open source model formats and will update the corresponding issue (if not the PR too). If simple enough I will work into this PR, then we can merge at leisure 🙌

PR #1914, opened 2026-07-11
fix: reject leading-zero octets in ipv6 IPv4-mapped forms

### LINKED ISSUE #1911: types.ipv6 accepts non-canonical, leading-zero IPv4-mapped addresses
### Describe the bug

`types.ipv6`'s two IPv4-mapped branches (`::ffff:a.b.c.d` and `x:x:x:x::a.b.c.d`) embed the same octet subpattern that `types.ipv4` used before #1904: `25[0-5]|2[0-4][0-9]|1?[0-9]{1,2}`. That pattern accepts 2-digit leading-zero octets (`01`, `00`, etc.) while already rejecting 3-digit leading-zero octets (`010`), so the embedded IPv4 suffix of an IPv6 address can match non-canonical, invalid addresses:

```python
import re
from outlines import types

pattern = re.compile(rf"^(?:{types.ipv6.pattern})$")

pattern.fullmatch("::ffff:01.1.1.1")    # matches -- should not
pattern.fullmatch("::ffff:00.0.0.0")    # matches -- should not
pattern.fullmatch("::ffff:192.168.1.1") # matches -- correct
```

Checked against the standard library for confirmation:

```python
>>> import ipaddress
>>> ipaddress.ip_address("::ffff:01.1.1.1")
    ...
ValueError: '::ffff:01.1.1.1' does not appear to be an IPv4 or IPv6 address
```

### Reproduction

```python
import re
from outlines import types

pattern = re.compile(rf"^(?:{types.ipv6.pattern})$")

bad = ["::ffff:01.1.1.1", "::ffff:00.0.0.0", "1:2:3:4::5.6.007.8"]
good = ["::ffff:192.168.1.1", "::ffff:127.0.0.1", "1:2:3:4::5.6.7.8", "::1", "2001:db8::1"]

for addr in bad:
    assert not pattern.fullmatch(addr), f"{addr} should not match"  # fails for the 2-digit forms

for addr in good:
    assert pattern.fullmatch(addr), f"{addr} should match"  # all pass
```

### Context

This is the same class of bug fixed for the standalone `ipv4` type in #1904, which that PR's own description flagged as present in `ipv6` too but intentionally left untouched to keep that PR surgical ("The `ipv6` regex embeds the same octet pattern for IPv4-mapped addresses; I've left it untouched to keep this surgical and can follow up separately."). Filing this so it's tracked as its own item -- the fix should be the same shape: replace both occurrences of `25[0-5]|2[0-4][0-9]|1?[0-9]{1,2}` in the `ipv6` definition with `25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9]`, the pattern #1904 introduced for `ipv4`.

### Environment

- outlines: current `main` (verified against `src/outlines/types/__init__.py`, `ipv6` definition, lines defining the `::ffff:` and `x:x:x:x::` IPv4-mapped branches)
- Pure regex issue, no GPU or model required to reproduce

joaquinhuigomez:
The two IPv4-mapped branches of `types.ipv6` (`::ffff:a.b.c.d` and `x:x:x:x::a.b.c.d`) embed the old octet subpattern that accepts 2-digit leading-zero octets, so invalid addresses like `::ffff:01.1.1.1` match even though `ipaddress.ip_address` rejects them. This switches those branches to the same canonical octet pattern used for the `ipv4` fix (#1904) so both types agree, and adds the leading-zero cases plus valid IPv4-mapped forms to the type tests (the new negative cases fail on main and pass with this change). Fixes #1911.


RobinPicard:
I have already merged #1936 doing the same thing as your PR, as I had not noticed yours yet (while it was opened first). Very sorry about that! Thanks for opening this PR in any case and looking forward to other contributions!

PR #1917, opened 2026-07-12
fix(types): quote non-string Dict keys so generated output is valid JSON

ErenAta16:
## Problem

`Dict[K, V]` where `K` is not `str` (e.g. `int`, `float`, `bool`, `date`, `time`, `datetime`) produces a regex that accepts **unquoted** keys, which is not valid JSON:

```python
from typing import Dict
import re, json

pattern = to_regex(python_types_to_terms(Dict[int, str]))
re.fullmatch(pattern, '{1:"a"}')   # matches
json.loads('{1:"a"}')              # json.decoder.JSONDecodeError:
# Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
```

JSON requires all object keys to be double-quoted strings regardless of the semantic key type. Any user constraining generation with a `Dict[int, ...]`-style annotation (directly, or nested inside a Pydantic/dataclass field) can get output that fails `json.loads`.

## Root cause

`_ensure_json_quoted` only wraps `String`/`Alternatives`-of-`String` terms (used for `Literal`/`Enum` string values) in quotes. Bare `Regex` terms such as `types.integer`, `types.number`, `types.boolean`, `types.date`, `types.time`, `types.datetime` are left unquoted when used as a dict key, since they're valid *unquoted* in other JSON contexts (e.g. as a value). `types.string` is the exception -- its pattern is already self-quoted (`"[^"]*"`).

## Fix

In `_handle_dict`, after computing the key term, wrap it in literal double quotes unless it's already self-quoted (`types.string`). `Dict[str, ...]` behavior is unchanged.

## Testing

- Updated `test_dsl_handle_dict` (it previously asserted the buggy unquoted behavior for `dict[int, str]`).
- Added `test_dsl_handle_dict_non_string_key_produces_valid_json`, which checks via `json.loads` that the generated regex only matches valid JSON for both the quoted and unquoted key forms.
- Full `tests/types/test_dsl.py` suite passes locally (59 passed).

### DIFF
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -838,19 +838,32 @@ def _handle_literal(args: tuple) -> Alternatives:
     return Alternatives([python_types_to_terms(arg) for arg in args])
 
 
-def _ensure_json_quoted(term: Term) -> Term:
-    """Wrap bare ``String`` terms in double quotes for JSON container contexts.
+def _ensure_json_quoted(term: Term, quote_regex: bool = False) -> Term:
+    """Wrap bare ``String`` terms (and, if ``quote_regex``, ``Regex`` terms) in
+    double quotes for JSON container contexts.
 
     When string literal values (from ``Literal`` or ``Enum``) appear inside
     container types (``List``, ``Tuple``, ``Dict``), they must be JSON-quoted
     so the generated regex matches valid JSON.  ``Regex``-based terms (e.g.
-    ``types.string``) already include their own quotes and are left untouched.
+    ``types.string``) already include their own quotes and are left untouched
+    by default.
+
+    ``quote_regex`` additionally quotes bare ``Regex`` terms (``types.integer``,
+    ``types.date``, numeric ``Literal``/``Enum`` members that resolve to
+    ``Regex``, etc.). Needed for ``Dict`` keys, where JSON requires every key
+    to be a quoted string regardless of the Python key type. ``types.string``
+    itself is never re-quoted, since its own pattern already includes the
+    quotes. This also correctly reaches ``Regex`` terms nested inside
+    ``Alternatives`` (e.g. ``Dict[Literal[1, 2], str]``), via the same
+    recursion ``String`` quoting already uses.
     """
     if isinstance(term, String):
         return String(f'"{term.value}"')
     if isinstance(term, Alternatives):
-        quoted = [_ensure_json_quoted(t) for t in term.terms]
+        quoted = [_ensure_json_quoted(t, quote_regex) for t in term.terms]
         return Alternatives(quoted)
+    if quote_regex and isinstance(term, Regex) and term.pattern != types.string.pattern:
+        return Sequence([String('"'), term, String('"')])
     return term
 
 
@@ -918,18 +931,15 @@ def _handle_dict(args: tuple, recursion_depth: int) -> Sequence:
     if args is None or len(args) != 2:
         raise TypeError(f"Dict must have exactly two type arguments. Got {args}.")
     # Add dict support with key:value pairs
-    key_type = _ensure_json_quoted(python_types_to_terms(args[0], recursion_depth + 1))
-    if isinstance(key_type, Regex) and key_type.pattern != types.string.pattern:
-        # JSON object keys must always be double-quoted strings, even when
-        # the Python key type is not `str` (e.g. `Dict[int, str]`,
-        # `Dict[date, str]`). `_ensure_json_quoted` only quotes `String`/
-        # `Alternatives` terms (from `Literal`/`Enum`); bare `Regex` terms
-        # like `types.integer`, `types.number`, `types.boolean`, `types.date`
-        # etc. are otherwise left unquoted, which produces a regex that
-        # matches invalid JSON (e.g. `{1:"a"}` instead of `{"1":"a"}`).
-        # `types.string` is excluded since its pattern is already
-        # self-quoted (`"[^"]*"`).
-        key_type = Sequence([String('"'), key_type, String('"')])
+    # JSON object keys must always be double-quoted strings, even when the
+    # Python key type is not `str` (e.g. `Dict[int, str]`, `Dict[date, str]`,
+    # or `Dict[Literal[1, 2], str]`, which resolves to an `Alternatives` of
+    # `Regex` terms). `quote_regex=True` covers all of those uniformly,
+    # including through `Alternatives`, via the same recursion `String`
+    # quoting already uses.
+    key_type = _ensure_json_quoted(
+        python_types_to_terms(args[0], recursion_depth + 1), quote_regex=True
+    )
     value_type = _ensure_json_quoted(python_types_to_terms(args[1], recursion_depth + 1))
     return Sequence(
         [
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -839,23 +839,15 @@ def _handle_literal(args: tuple) -> Alternatives:
 
 
 def _ensure_json_quoted(term: Term, quote_regex: bool = False) -> Term:
-    """Wrap bare ``String`` terms (and, if ``quote_regex``, ``Regex`` terms) in
-    double quotes for JSON container contexts.
+    """Wrap ``String`` terms in double quotes for JSON container contexts.
 
-    When string literal values (from ``Literal`` or ``Enum``) appear inside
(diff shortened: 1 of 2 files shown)


src/outlines/types/dsl.py
```diff
@@ -919,6 +919,17 @@ def _handle_dict(args: tuple, recursion_depth: int) -> Sequence:
         raise TypeError(f"Dict must have exactly two type arguments. Got {args}.")
     # Add dict support with key:value pairs
     key_type = _ensure_json_quoted(python_types_to_terms(args[0], recursion_depth + 1))
+    if isinstance(key_type, Regex) and key_type.pattern != types.string.pattern:
```
RobinPicard: I think a better solution would be to add a parameter to `_ensure_json_quoted` to indicate whether to add quotation marks to `Regex` terms on top of `String` terms (default would be `False`). That way we benefit from the recursive functioning of this function in case the term is an `Alternative`  and we have a better containment of the issue (everything related to adding quotation marks in a single function).

RobinPicard (commented):
Thanks for the PR, it's a good idea to fix that. I just have a suggestion to improve it.

src/outlines/types/dsl.py
ErenAta16: Good catch, and it's more than a style improvement: the isinstance(key_type, Regex) check I had only handles a bare Regex, it misses Dict[Literal[1, 2, 3], str] entirely, since int Literal members go through _handle_literal into Alternatives([Regex("1"), Regex("2"), Regex("3")]), not a top-level Regex. Confirmed that produces unquoted numeric keys with the old code, same bug through a different path.

Added quote_regex to _ensure_json_quoted (d5c58af) and dropped the separate check, so it goes through the same Alternatives recursion String quoting already uses. Added a regression test for the Literal[int] case specifically. Full test_dsl.py passes except two pre-existing, unrelated failures on this Windows box (temp-file PermissionError in the from_file tests, reproduces identically on main without my changes).

tests/types/test_dsl.py
```diff
@@ -839,21 +839,73 @@ def test_dsl_handle_dict():
...
+    result = _handle_dict(get_args(dict_type), recursion_depth=0)
+    quoted_int_key = Sequence([String('"'), types.integer, String('"')])
+    assert result.terms[1].term.terms[0] == quoted_int_key
+    assert result.terms[1].term.terms[2] == types.string
+
+
+def test_dsl_handle_dict_non_string_key_produces_valid_json():
+    """Regression test: Dict[int, str] (and other non-str key types) must
+    only match strings that are valid JSON, i.e. the key must be quoted."""
+    import json
```
RobinPicard: Let's keep all the imports at the top of the file when possible

tests/types/test_dsl.py
```diff
@@ -839,21 +839,73 @@ def test_dsl_handle_dict():
...
+
+    dict_type = dict[int, str]
```
RobinPicard: It's duplicating what's above

tests/types/test_dsl.py
```diff
@@ -839,21 +839,73 @@ def test_dsl_handle_dict():
...
+    result = _handle_dict(get_args(dict_type), recursion_depth=0)
+    pattern = to_regex(result)
+
+    # unquoted key: not valid JSON, must not match
+    assert _re.fullmatch(pattern, '{1:"a"}') is None
+    with pytest.raises(json.JSONDecodeError):
+        json.loads('{1:"a"}')
```
RobinPicard: That does not make sense, we don't want to test that the `json` library works well.

src/outlines/types/dsl.py
```diff
@@ -918,7 +931,15 @@ def _handle_dict(args: tuple, recursion_depth: int) -> Sequence:
     if args is None or len(args) != 2:
         raise TypeError(f"Dict must have exactly two type arguments. Got {args}.")
     # Add dict support with key:value pairs
-    key_type = _ensure_json_quoted(python_types_to_terms(args[0], recursion_depth + 1))
+    # JSON object keys must always be double-quoted strings, even when the
```
RobinPicard: I don't think we need such a long comment

src/outlines/types/dsl.py
```diff
@@ -838,19 +838,32 @@ def _handle_literal(args: tuple) -> Alternatives:
     return Alternatives([python_types_to_terms(arg) for arg in args])
 
 
-def _ensure_json_quoted(term: Term) -> Term:
-    """Wrap bare ``String`` terms in double quotes for JSON container contexts.
+def _ensure_json_quoted(term: Term, quote_regex: bool = False) -> Term:
+    """Wrap bare ``String`` terms (and, if ``quote_regex``, ``Regex`` terms) in
```
RobinPicard: Let's try to shorten it

RobinPicard (changes requested):
The code change itself looks fine, but there are some changes to make in terms of style and the tests need to be better focused

tests/types/test_dsl.py
ErenAta16: Moved to the module-level imports (json and Literal were both already imported at the top).

tests/types/test_dsl.py
ErenAta16: Right, dropped it. The dict[int, str] key-quoting is already covered by the structural assertion in test_dsl_handle_dict, so this one was redundant.

tests/types/test_dsl.py
ErenAta16: Agreed, removed the json.loads / JSONDecodeError lines. Kept just the assertion on our generated term structure.

src/outlines/types/dsl.py
ErenAta16: Trimmed it down to two lines.

src/outlines/types/dsl.py
ErenAta16: Shortened, kept the quote_regex explanation and the Alternatives note, dropped the rest.

ErenAta16:
Pushed 4191032 addressing the review: shortened the _ensure_json_quoted docstring and the _handle_dict comment, moved the test imports up, dropped the redundant dict[int, str] regex test (already covered structurally) and the json.loads checks, and refocused the remaining test on the Literal[int] Alternatives case as a structural assertion. The quote_regex refactor you suggested is in d5c58af. Full test_dsl.py passes (57, minus the 2 pre-existing from_file temp-file failures unrelated to this change).

ErenAta16:
Note on the red X: the 3.10 job failed inside the ollama integration tests (`llama-server process has terminated: signal: segmentation fault`, plus a `127.0.0.1:11434 address already in use` bind error), not on anything this PR touches. `tests/types/test_dsl.py` passes on both 3.10 and 3.13 in that same run. So the failure is infra flakiness in the provider tests, unrelated to this diff.

tests/types/test_dsl.py
```diff
@@ -839,21 +839,45 @@ def test_dsl_handle_dict():
...
+    # strings, so the bare `types.integer` regex must be wrapped in quotes
+    # even though the Python key type is `int`.
+    dict_type = dict[int, str]
+    result = _handle_dict(get_args(dict_type), recursion_depth=0)
+    quoted_int_key = Sequence([String('"'), types.integer, String('"')])
+    assert result.terms[1].term.terms[0] == quoted_int_key
+    assert result.terms[1].term.terms[2] == types.string
+
+
+def test_dsl_handle_dict_literal_int_key_quoted():
```
RobinPicard: Let's just include it in the test `test_dsl_handle_dict` as it's doing the same thing with a different value for `dict_type`

ErenAta16: Done in 13fe8dd. Folded it in as another `dict_type` block right after the `dict[int, str]` one, since the assertions are the same shape and it reads as a continuation rather than a separate concern.

Kept the two-line comment on it because the *why* isn't obvious from the code alone: `Literal[1, 2, 3]` reaches `_handle_dict` as `Alternatives([Regex("1"), ...])` rather than a bare `Regex`, which is exactly the path the original version of this fix missed. `tests/types/test_dsl.py`: 58 passed.

RobinPicard (approved):

ErenAta16:
Thanks for the review and the merge, @RobinPicard. The `quote_regex` suggestion in particular made this a better fix than what I opened with, since routing through `_ensure_json_quoted`'s existing recursion is what caught the `Dict[Literal[1, 2, 3], str]` case my original `isinstance(key_type, Regex)` check missed entirely.

PR #1920, opened 2026-07-13
docs: add DaoXE to OpenAI-compatible providers examples

seven7763:
## Summary
Documents [DaoXE](https://daoxe.com) as an OpenAI-compatible provider example in `docs/features/models/openai_compatible.md` (`base_url=https://daoxe.com/v1`, account model IDs).

I maintain DaoXE. Examples: https://github.com/seven7763/DaoXE-AI

### DIFF
--- a/docs/features/models/openai_compatible.md
+++ b/docs/features/models/openai_compatible.md
@@ -46,6 +46,7 @@ Many providers offer OpenAI-compatible endpoints:
 - **Fireworks AI**
 - **Perplexity**
 - **Local servers** (LocalAI, etc.)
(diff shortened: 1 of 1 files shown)


RobinPicard (commented):
I think we would rather restrict the list of OpenAI inference providers to those that serve a high-enough amount of traffic to avoid cluttering our documentation. We could add `DaoXE` in the future though!

seven7763:
That's a fair bar, and keeping the provider list short makes sense. Closing this rather than leaving it sitting in your queue — happy to revisit down the line if the traffic threshold is met.

Disclosure: I work on DaoXE. Thanks for the clear answer.

PR #1924, opened 2026-07-14
Fix the JsonSchema whitespace_pattern being ignored by the backend

dhanavanthesh:
## Problem

When you set `whitespace_pattern` on a `JsonSchema` output type, it has no effect during generation on the default `outlines_core` backend. The JSON is always produced with the default whitespace, whatever pattern you pass.

The cause is a dropped argument. When the generator builds the logits processor for a `JsonSchema`, it passes only `term.schema` to the backend and never passes `term.whitespace_pattern` (`src/outlines/generator.py`). The backend then calls `build_regex_from_schema(json_schema)` without the pattern (`src/outlines/backends/outlines_core.py`). Calling `to_regex(JsonSchema(...))` directly does apply the pattern, so the two paths disagree.

`whitespace_pattern` is a documented parameter (`docs/features/core/output_types.md`) and worked in v0, where it was added to control JSON whitespace (#916). It stopped being applied when JSON-schema handling moved to the backend layer in the v1 refactor.

## Fix

Pass `whitespace_pattern` through to the backend so it is actually used:

- The generator now passes `term.whitespace_pattern` when it builds the JSON-schema logits processor.
- `get_json_schema_logits_processor` (the backend dispatcher, `BaseBackend`, and `OutlinesCoreBackend`) takes an optional `whitespace_pattern` and forwards it to `build_regex_from_schema`.
- The `llguidance` and `xgrammar` backends build the grammar their own way and cannot use this pattern, so they raise `NotImplementedError` if one is set instead of ignoring it.

The default (`whitespace_pattern=None`) behaves exactly as before.

## Tests

Two new tests in `tests/backends/test_outlines_core.py`:

- one checks that a pattern you pass reaches `build_regex_from_schema`,
- one checks that the default passes `None`.

Both fail on `main` and pass with this change (verified by stashing the fix). The `tests/backends/test_outlines_core.py` suite passes (one MLX test is skipped). `ruff` and `mypy` are clean.

## Docs

`docs/features/core/output_types.md` now says `whitespace_pattern` is applied by the `outlines_core` backend, and that `llguidance` / `xgrammar` do not support it.

### DIFF
--- a/docs/features/core/output_types.md
+++ b/docs/features/core/output_types.md
@@ -178,7 +178,7 @@ output_type = JsonSchema(schema_dict)
 ```
 
 `JsonSchema` accepts two optional parameters:
-- `whitespace_pattern` (defaults to `None`): specifies the pattern to use for JSON syntactic whitespace. If none is provided, the default permissive JSON whitespace rules are used.
+- `whitespace_pattern` (defaults to `None`): specifies the pattern to use for JSON syntactic whitespace. If none is provided, the default permissive JSON whitespace rules are used. It is applied by the `outlines_core` backend; the `llguidance` and `xgrammar` backends do not support it and raise an error if one is set.
 - `ensure_ascii` (defaults to `True`): defines the value to use for the argument `ensure_ascii` of the `json.dumps` method. If false, non-ASCII characters will be turned into unicodes.
 
 ### Regex Patterns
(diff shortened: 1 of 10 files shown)


RobinPicard (approved):
Great PR, thanks! And good idea to raise an error when the parameter is used with `xgrammar`/`llguidance` instead of silently ignoring it.

PR #1930, opened 2026-07-15
Render Optional enum/nested fields in the schema filter

ritsth:
`parse_pydantic_schema_value` handled `enum` and `$ref` properties but not
`anyOf`, which is how pydantic renders `Optional[...]` and unions. So the
`schema` filter rendered enum values / nested structure for required fields
but fell back to a `<name>` placeholder for the `Optional` equivalents —
a gap in the enum-rendering support added in d0bbd703.

The fix adds an `anyOf` branch that recurses into the first non-null member,
so an `Optional[T]` field renders exactly like a required `T`. Extended
`test_get_schema` with `Optional[Enum]` and `Optional[NestedModel]` cases
(fail before, pass after). Full `test_templates.py` green; `ruff` clean.

Used an AI assistant to find and fix this; I reviewed, reproduced the issue,
and ran the tests myself before submitting.

### DIFF
--- a/src/outlines/templates.py
+++ b/src/outlines/templates.py
@@ -328,5 +328,17 @@ def parse_pydantic_schema_value(name, schema, definitions):
     elif "enum" in schema:
         values = " | ".join(str(value) for value in schema["enum"])
         return f"<{values}>"
+    elif "anyOf" in schema:
+        # `Optional[T]` is rendered as an `anyOf` of `T` and `null`. Recurse into
+        # the single non-null member so the field renders like its underlying type
+        # (e.g. `Optional[Enum]` shows the enum values, not a `<name>` placeholder).
+        # Genuine multi-member unions (e.g. `A | B`) fall through to the `<name>`
+        # placeholder rather than misleadingly advertising only their first branch.
+        non_null = [
+            member for member in schema["anyOf"] if member.get("type") != "null"
+        ]
+        if len(non_null) == 1:
+            return parse_pydantic_schema_value(name, non_null[0], definitions)
+        return f"<{name}>"
     else:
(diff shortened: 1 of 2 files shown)


src/outlines/templates.py
```diff
@@ -328,5 +328,17 @@ def parse_pydantic_schema_value(name, schema, definitions):
...
+        # `Optional[T]` and unions are rendered as `anyOf`. Recurse into the
+        # first non-null member so the field renders like its underlying type
+        # (e.g. `Optional[Enum]` shows the enum values, not a `<name>` placeholder).
+        # `anyOf` is only emitted when there are 2+ alternatives, so there is
+        # always at least one non-null member.
+        non_null = [
+            member for member in schema["anyOf"] if member.get("type") != "null"
+        ]
+        if non_null:
+            return parse_pydantic_schema_value(name, non_null[0], definitions)
```
Sanjays2402, line 341: This also catches non-optional unions and silently drops every branch after the first. For `Armor | Weapon`, the template will only advertise Armor values; can this path be limited to an `anyOf` with exactly one non-null member plus `null`?

ritsth: Good catch, you're right. As written, `Armor | Weapon` would advertise only Armor's fields and quietly drop the rest, which is misleading. I'll gate this on `len(non_null) == 1` so it only applies to the single-type case `(Optional[T])`, and genuine multi-member unions fall back to the `<name>` placeholder the way they did before this PR. That keeps the change scoped to Optional fields without pretending to represent a union it can't. I will push that shortly with a test for the union case.

RobinPicard (approved):
Thanks!

PR #1867, opened 2026-05-31
Add ipv6 custom type

u7k4rs6:
Closes #1866

## Summary

Adds an `ipv6` regex type to `outlines.types`, alongside the existing `ipv4`
type. Follows the same pattern as the recently merged `semver` and
`mac_address` types (#1863).

## Details

The pattern covers the RFC 4291 textual forms:
- full eight-group form (`2001:0db8:0000:...:7334`)
- `::` zero-compression, including `::` and `::1`
- IPv4-mapped / embedded forms (e.g. `::ffff:192.168.1.1`)

Consistent with the existing `ipv4` pattern, it validates textual shape only
and does not enforce canonical lowercase or zero-suppression.

## Tests

Added parametrized match/no-match cases to `tests/types/test_custom_types.py`
covering loopback, all-zeros, compressed, full, link-local, and IPv4-mapped
forms, plus invalid cases (triple colon, over-long group, nine groups, bare
IPv4). `pytest tests/types/test_custom_types.py` passes locally; pre-commit
(ruff, mypy) clean.

## Checklist
- [x] Title is self-explanatory
- [x] Linked to issue (#1866) discussing the feature
- [x] Branch based on latest `main`
- [x] One commit, imperative-mood message
- [x] Tests added
- [x] Pre-commit hooks pass (ruff, mypy)

PR #1934, opened 2026-07-18
Add hex_color and slug custom types

pepepeboom:
## Summary

Partially addresses #1303 by adding two more regex-based custom types, following the structure of #1863 (semver, mac_address):

- `hex_color`: CSS hexadecimal color notation, `#rgb` and `#rrggbb`
- `slug`: lowercase-hyphenated URL slugs, e.g. `my-post-title`

Of the types listed in #1303, `semver` and `mac_address` (#1863) and `ipv6` (#1867) are already merged, and `ipv4`/`uuid4`/`hex_str` predate the issue — this PR adds only the two missing ones above.

## Design notes

- `hex_color = Regex(r"#(?:[0-9A-Fa-f]{3}|[0-9A-Fa-f]{6})")` accepts only the three- and six-digit forms. The four- and eight-digit alpha forms from CSS Color Module Level 4 are intentionally excluded (and covered by reject tests), consistent with how the existing types stick to the single canonical textual form (e.g. `mac_address` accepts only the colon-separated form).
- `slug = Regex(r"[a-z0-9]+(?:-[a-z0-9]+)*")` structurally rejects uppercase, underscores, the empty string, and leading/trailing/consecutive hyphens, matching the output of common slugify helpers.
- Both patterns are unanchored like every other type in the file; full-match semantics are applied by `Term.validate`/`Term.matches` and the structured-generation backends.

Beyond the unit tests, I brute-force verified both patterns: hex_color against all payload lengths 0–9 and all printable ASCII characters in each digit position, and slug exhaustively against all strings of length ≤ 4 over the alphabet `{a, 1, -, _, A}` compared to a reference predicate.

## Tests

- `python -m pytest tests/types/test_custom_types.py` — 140 passed (18 new parametrized cases)

---

This PR was prepared with AI assistance (Claude Code); all changes and test results were verified locally.


RobinPicard (approved):
Good idea, thanks!

PR #1939, opened 2026-07-20
fix(types): make non-required dataclass fields Optional, matching the pydantic converter

### LINKED ISSUE #1938: json_schema_dict_to_dataclass drops optionality: non-required fields annotated as bare type with a None default, diverging from the pydantic converter
### What

`json_schema_dict_to_dataclass` (in `src/outlines/types/json_schema_utils.py`) drops the optionality of non-required fields. For a field that isn't in the schema's `required` list, it sets a `None` default but leaves the annotation as the bare type instead of wrapping it in `Optional[...]`, unlike the pydantic and typeddict converters right next to it. The result is a dataclass whose annotation and default contradict each other (`age: int = None`), and, more importantly, `JsonSchema.convert_to(schema, "dataclass")` and `convert_to(schema, "pydantic")` produce structurally different constraints for the same input schema.

### Where

`src/outlines/types/json_schema_utils.py`, `json_schema_dict_to_dataclass`:

```python
for property, details in properties.items():
    typ = schema_type_to_python(details, "dataclass")
    annotations[property] = typ          # <- not wrapped in Optional for non-required fields

    if property not in required:
        defaults[property] = None        # <- None default on a non-Optional annotation
```

Compare the two sibling converters in the same file, which both preserve optionality:

```python
# json_schema_dict_to_pydantic
if property not in required:
    field_definitions[property] = (Optional[typ], None)   # value made nullable

# json_schema_dict_to_typeddict
if property not in required:
    typ = NotRequired[typ]                                # key made optional
```

### Reproduction

```python
from dataclasses import fields
from typing import get_type_hints
from pydantic import TypeAdapter

from outlines.types.json_schema_utils import (
    json_schema_dict_to_dataclass,
    json_schema_dict_to_pydantic,
)

schema = {
    "type": "object",
    "title": "Person",
    "properties": {
        "name": {"type": "string"},
        "age": {"type": "integer"},   # optional: not in `required`
    },
    "required": ["name"],
}

dc = json_schema_dict_to_dataclass(schema)
hints = get_type_hints(dc)
print("dataclass age annotation:", hints["age"])          # <class 'int'>
print("dataclass age default:", fields(dc)[1].default)    # None

# What outlines actually feeds back into grammar generation for a dataclass
# output type is TypeAdapter(dc).json_schema():
print("dataclass -> schema:", TypeAdapter(dc).json_schema()["properties"]["age"])
# {'default': None, 'title': 'Age', 'type': 'integer'}

pm = json_schema_dict_to_pydantic(schema)
print("pydantic  -> schema:", pm.model_json_schema()["properties"]["age"])
# {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'default': None, 'title': 'Age'}
```

### Actual vs expected

For the same input schema, the two converters disagree on the optional `age` field:

| converter | `age` after round-trip | allows `null`? |
|---|---|---|
| pydantic  | `anyOf: [{integer}, {null}]`, `default: null` | yes |
| dataclass | `type: integer`, `default: null` | no |

The dataclass output has two problems:

1. It loses the optionality declared by the schema. The field is constrained to `integer` only, so structured generation can never emit `null` for it, even though the schema said it's optional. The pydantic path (via `Optional[int]`) keeps `null` in the allowed set.
2. Its `default: null` doesn't validate against its own `type: integer`. The annotation (`int`) and the default (`None`) contradict each other, which type checkers also flag.

Since `convert_to(schema, ["dataclass", "typeddict", "pydantic"])` exists specifically to let the caller pick a container for the *same* schema, the choice of container changing what values are accepted is surprising.

### Suggested fix

Wrap the annotation in `Optional[...]` for non-required fields, mirroring the pydantic converter:

```python
    if property not in required:
        typ = Optional[typ]
        defaults[property] = None
    annotations[property] = typ
```

This makes the dataclass path produce the same nullable field the pydantic path does, and removes the annotation/default contradiction. I can open a PR with this plus a regression test that asserts the three converters agree on optional-field nullability.

### Environment

Reproduced against `dottxt-ai/outlines` `main` (current), which already includes the `kw_only` dataclass fix from #1872 (a separate ordering bug in the same function). Pure Python, no model/backend needed.

ErenAta16:
Closes #1938.

## What

`json_schema_dict_to_dataclass` left a non-required field's annotation as the bare type and only added a `None` default. So a field that isn't in the schema's `required` list came out as `age: int = None`, where the annotation (`int`) and the default (`None`) contradict each other, and the optionality the schema declared was lost.

The two sibling converters in the same file already handle this correctly: `json_schema_dict_to_pydantic` wraps the value in `Optional[typ]`, and `json_schema_dict_to_typeddict` marks the key `NotRequired[typ]`. Only the dataclass path left it as a bare type.

## Why it matters

`JsonSchema.convert_to(schema, ["dataclass", "typeddict", "pydantic"])` is meant to let a caller pick a container for the *same* schema, so the containers shouldn't disagree on what's accepted. But they did. Round-tripping the generated type back through `TypeAdapter(...).json_schema()` (which is the path an outline dataclass output type takes into grammar generation) shows the divergence for an optional `age: integer` field:

| converter | `age` after round-trip | allows `null`? |
|---|---|---|
| pydantic  | `anyOf: [{integer}, {null}]`, `default: null` | yes |
| dataclass (before) | `type: integer`, `default: null` | no |

The dataclass path both loses the null option (so constrained generation can never emit `null` for a field the schema said is optional) and carries a `default: null` that doesn't validate against its own `type: integer`.

## Fix

Wrap the annotation in `Optional[...]` for non-required fields, mirroring the pydantic converter:

```python
if property not in required:
    typ = Optional[typ]
    defaults[property] = None
annotations[property] = typ
```

After this, the dataclass path produces the same `anyOf: [integer, null]` the pydantic path does.

## Tests

- Updated the existing tests that asserted the old bare-type behavior (`test_schema_type_to_python_object`, `test_json_schema_dict_to_dataclass_basic`, `_array_enum`, `_nested_object`). Notably `test_schema_type_to_python_object` was already documenting the divergence: it asserted `Optional[int]` for pydantic, `NotRequired[int]` for typeddict, and `int` for dataclass on the same optional field.
- Added `test_json_schema_dict_to_dataclass_optional_field_is_nullable`, which asserts the annotation is `Optional[int]` and that the dataclass and pydantic converters agree on optional-field nullability after a round-trip.

RED/GREEN verified: the new test fails on `main` (annotation comes out as bare `int`) and passes with the fix. Full `tests/types/test_json_schema_utils.py` passes (24 passed), and the wider `tests/types/` suite passes except for pre-existing unrelated skips (`test_custom_types.py` needs the optional `airportsdata` extra; some `from_file` tests hit a Windows temp-file permission issue that reproduces on `main` without this change).

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -180,10 +180,16 @@ def json_schema_dict_to_dataclass(
 
     for property, details in properties.items():
         typ = schema_type_to_python(details, "dataclass")
-        annotations[property] = typ
 
         if property not in required:
+            # Mirror the pydantic converter: a non-required field is nullable,
+            # so wrap the annotation in Optional rather than leaving a bare type
+            # with a None default (which contradicts its own annotation and
+            # loses the optionality the schema declared).
+            typ = Optional[typ]
             defaults[property] = None
+
+        annotations[property] = typ
 
     class_dict = {
         '__annotations__': annotations,
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -1,6 +1,6 @@
 import sys
 from dataclasses import is_dataclass
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, List, Literal, Optional, Union, get_type_hints
 
 from pydantic import BaseModel, TypeAdapter
 from pydantic_core import PydanticUndefined
@@ -81,7 +81,8 @@ def test_schema_type_to_python_object():
     assert hasattr(dataclass_result, "__dataclass_fields__")
     assert dataclass_result.__annotations__["name"] is str
     assert not hasattr(dataclass_result, "name")
-    assert dataclass_result.__annotations__["age"] is int
+    # Non-required field is nullable, matching the pydantic/typeddict callers above.
+    assert dataclass_result.__annotations__["age"] == Optional[int]
     assert dataclass_result.age is None
 
 
@@ -337,7 +338,7 @@ def test_json_schema_dict_to_dataclass_basic():
 
     annotations = result.__annotations__
     assert annotations["name"] is str
-    assert annotations["age"] is int
+    assert annotations["age"] == Optional[int]
     assert not hasattr(result, "name")
     assert result.age is None
 
@@ -362,7 +363,8 @@ def test_json_schema_dict_to_dataclass_array_enum():
     assert result.__name__ == "AnonymousDataclass"
 
     annotations = result.__annotations__
-    assert annotations["tags"] == List[str]
+    # `tags` is not required, so its annotation is nullable.
+    assert annotations["tags"] == Optional[List[str]]
     assert annotations["status"] == Literal[("active", "inactive", "pending")]
     assert not hasattr(result, "status")
     assert result.tags is None
@@ -394,7 +396,7 @@ def test_json_schema_dict_to_dataclass_nested_object():
 
     field = annotations["field"]
     assert field.__annotations__["name"] is str
-    assert field.__annotations__["age"] is int
+    assert field.__annotations__["age"] == Optional[int]
     assert not hasattr(field, "name")
     assert field.age is None
 
@@ -415,3 +417,39 @@ def test_json_schema_dict_to_dataclass_optional_before_required():
     instance = result(user_id=5)
     assert instance.user_id == 5
     assert instance.nickname is None
+
+
+def test_json_schema_dict_to_dataclass_optional_field_is_nullable():
+    # A non-required field must be annotated Optional[...], not a bare type with
+    # a None default. Otherwise the annotation and default contradict each other
+    # and the dataclass converter loses the optionality the pydantic converter
+    # keeps, so the two disagree for the same input schema.
+    schema = {
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "age": {"type": "integer"},
+        },
+        "required": ["name"],
+    }
+
+    dataclass_result = json_schema_dict_to_dataclass(schema)
+    hints = get_type_hints(dataclass_result)
+    assert hints["age"] == Optional[int]
+
+    # The three converters are offered for the same schema, so they must agree
+    # on whether the optional field accepts null. Compare against the pydantic
+    # converter, which already produces a nullable field.
+    def allows_null(field_schema: dict) -> bool:
+        if field_schema.get("type") == "null":
+            return True
+        if "anyOf" in field_schema:
+            return any(m.get("type") == "null" for m in field_schema["anyOf"])
+        return False
+
+    dataclass_age = TypeAdapter(dataclass_result).json_schema()["properties"]["age"]
+    pydantic_age = json_schema_dict_to_pydantic(schema).model_json_schema()[
+        "properties"
+    ]["age"]
+    assert allows_null(dataclass_age)
+    assert allows_null(pydantic_age)


src/outlines/types/json_schema_utils.py
```diff
@@ -180,11 +180,17 @@ def json_schema_dict_to_dataclass(
...
     for property, details in properties.items():
         typ = schema_type_to_python(details, "dataclass")
-        annotations[property] = typ
 
         if property not in required:
+            # Mirror the pydantic converter: a non-required field is nullable,
+            # so wrap the annotation in Optional rather than leaving a bare type
+            # with a None default (which contradicts its own annotation and
+            # loses the optionality the schema declared).
+            typ = Optional[typ]
```
Sanjays2402, line 189: `required` controls whether the key may be omitted; it does not make the value nullable. For an optional `age: {"type": "integer"}`, this changes the round-tripped schema to accept `"age": null`, which the input schema rejects. Mirroring the pydantic converter just copies the same conflation.

ErenAta16:
You're right, and I verified it rather than taking my own PR's premise for granted. Ran the three forms through `TypeAdapter(...).json_schema()` and validated documents against each:

| | `age` after round-trip | omitted | `3` | `null` |
|---|---|---|---|---|
| **input schema** (`age: {"type":"integer"}`, not in `required`) | - | accepted | accepted | **rejected** |
| main today (`age: int = None`) | `{"type":"integer","default":null}` | accepted | accepted | **rejected** |
| this PR (`age: Optional[int] = None`) | `{"anyOf":[{"type":"integer"},{"type":"null"}],"default":null}` | accepted | accepted | **accepted** |

So main round-trips the input schema's accepted set exactly, and my change widens it. That's a fidelity regression, not a fix, exactly as you described.

What makes this clearer is that the answer is already written down in this file. The typeddict converter twenty lines up carries the comment:

```python
# NotRequired (PEP 655) marks the KEY optional; Optional only makes the
# value nullable, leaving the key required on a total=True TypedDict.
```

I confirmed `NotRequired[int]` round-trips to a bare `{"type": "integer"}` that stays out of `required` and still rejects `null`, so the typeddict path is the one that gets this right. I read that comment as being about a `total=True` mechanic specific to TypedDict, when it's actually stating the general `required`-vs-nullable distinction you're making, and it applies just as much to the converter I was changing.

The consequence is bigger than this PR, though, and it inverts my original issue (#1938). I anchored on "the dataclass converter disagrees with pydantic, so make it match pydantic." But the disagreement is real and pydantic is the one that's wrong:

```python
# json_schema_dict_to_pydantic
if property not in required:
    field_definitions[property] = (Optional[typ], None)   # nullable value
```

That produces `anyOf: [integer, null]` for every non-required field, so `convert_to(schema, "pydantic")` already accepts `null` where the source schema doesn't. For constrained generation that means the grammar permits `"age": null` for a schema that forbids it. Three converters, and the two that use `Optional` both over-accept while the one using `NotRequired` is correct.

Pydantic can express the right thing (a bare type with a default stays non-nullable and drops out of `required`), so this is fixable there, not a limitation we're stuck with.

Given that, I'd rather not land this as-is. I'll close it and re-file against the pydantic converter, which is the one actually widening constraints today. The remaining wart on main is cosmetic by comparison: `age: int = None` type-checks badly and emits `"default": null` under a non-nullable type, but it does not change what's accepted. That's worth a separate, narrower fix and shouldn't have been bundled with a nullability change.

Thanks for catching this. It would have quietly loosened generation constraints for every optional field.

PR #1941, opened 2026-07-20
fix(types): allow the empty collection for List[T] and variadic Tuple[T, ...]

### LINKED ISSUE #1940: DSL regex for List[T] and variadic Tuple[T, ...] forbids the empty collection, diverging from Dict and the JSON-schema path
### What

The DSL regex for `List[T]` (and the variadic `Tuple[T, ...]`) requires at least one element, so a valid empty collection can never be generated. `_handle_list` / `_handle_tuple` build `[item(, item)*]`, with no way to match `[]`. This diverges both from `_handle_dict` in the same file (which correctly allows `{}`) and from the JSON-schema path, which allows an empty array for the same type.

### Where

`src/outlines/types/dsl.py`:

`_handle_list`:

```python
return Sequence(
    [
        String("["),
        item_type,
        KleeneStar(Sequence([String(", "), item_type])),
        String("]"),
    ]
)
```

`_handle_tuple` (variadic `Tuple[T, ...]` branch):

```python
return Sequence(
    [
        String("("),
        item_term,
        KleeneStar(Sequence([String(", "), item_term])),
        String(")"),
    ]
)
```

Both place `item` before the `KleeneStar`, so at least one element is mandatory. Compare `_handle_dict` right below them, which wraps the whole content in `Optional(...)` and therefore accepts `{}`:

```python
return Sequence(
    [
        String("{"),
        Optional(
            Sequence(
                [key_type, String(":"), value_type, KleeneStar(...)]
            )
        ),
        String("}"),
    ]
)
```

### Reproduction

```python
import re
from typing import Dict, List, Tuple

for tp, empty in [(List[int], "[]"), (Tuple[int, ...], "()"), (Dict[str, int], "{}")]:
    pattern = to_regex(python_types_to_terms(tp))
    print(f"{tp}: matches {empty!r} -> {re.fullmatch(pattern, empty) is not None}")
```

Output:

```
typing.List[int]: matches '[]' -> False
typing.Tuple[int, ...]: matches '()' -> False
typing.Dict[str, int]: matches '{}' -> True
```

So within the DSL itself, dicts allow empty but lists and variadic tuples don't.

The divergence from the JSON-schema path is just as clear. A pydantic model with a `List[int]` field goes through `JsonSchema` -> `build_regex_from_schema` instead, and that regex does allow the empty array:

```python
from outlines.types.dsl import JsonSchema, to_regex

class M(BaseModel):
    items: List[int]

pattern = to_regex(JsonSchema(M))
print(re.fullmatch(pattern, '{"items": []}') is not None)   # True
```

So `list[int]` as a bare output type forbids `[]`, but the same `List[int]` as a pydantic field allows it.

### Actual vs expected

`[]` is a valid value for `list[int]` (and `()` for `tuple[int, ...]`) in both Python and JSON, with no `min_items` constraint in play. Structured generation should be able to produce it. Right now, using a bare `list[int]`/`tuple[int, ...]` output type, the model is forced to emit at least one element, which can push it into inventing spurious items when the correct answer is an empty list.

### Suggested fix

Wrap the list/tuple content in `Optional`, exactly as `_handle_dict` already does:

```python
# _handle_list
return Sequence(
    [
        String("["),
        Optional(
            Sequence(
                [item_type, KleeneStar(Sequence([String(", "), item_type]))]
            )
        ),
        String("]"),
    ]
)
```

and the analogous change for the variadic-tuple branch of `_handle_tuple`. This makes both allow the empty collection while keeping the one-or-more forms unchanged, and brings them in line with `_handle_dict` and the JSON-schema path. Fixed-length tuples (`Tuple[int, str]`) are unaffected, since they don't go through this branch. I can open a PR with this plus a regression test.

### Environment

ErenAta16:
Closes #1940.

## What

`_handle_list` and `_handle_tuple` (in `src/outlines/types/dsl.py`) built `[item(, item)*]`, placing the item term before the `KleeneStar`, so at least one element was mandatory. The valid empty collection (`[]` for a list, `()` for a variadic tuple) could never be generated.

This diverged in two directions:

- From `_handle_dict` in the same file, which wraps its content in `Optional(...)` and therefore accepts `{}`.
- From the JSON-schema path: a pydantic model with a `List[int]` field goes through `JsonSchema` -> `build_regex_from_schema`, and that regex allows the empty array. So `list[int]` as a bare output type forbade `[]`, while the same `List[int]` as a pydantic field allowed it.

## Reproduction (before)

```python
import re
from typing import Dict, List, Tuple

for tp, empty in [(List[int], "[]"), (Tuple[int, ...], "()"), (Dict[str, int], "{}")]:
    pattern = to_regex(python_types_to_terms(tp))
    print(tp, "matches", empty, "->", re.fullmatch(pattern, empty) is not None)
# List[int]        matches [] -> False
# Tuple[int, ...]  matches () -> False
# Dict[str, int]   matches {} -> True
```

## Fix

Wrap the list/tuple contents in `Optional`, exactly as `_handle_dict` already does:

```python
Sequence([
    String("["),
    Optional(Sequence([item_type, KleeneStar(Sequence([String(", "), item_type]))])),
    String("]"),
])
```

and the same for the variadic branch of `_handle_tuple`. The empty collection now matches, the one-or-more forms are unchanged, and fixed-length tuples (`Tuple[int, str]`) are unaffected since they don't go through this branch.

## Tests

- Updated the structural tests in `test_dsl.py` (`test_dsl_handle_list`, `test_dsl_handle_tuple`, `test_list_of_literals_quoted`, `test_list_of_int_unchanged`, `test_list_of_bool_unchanged`, `test_tuple_ellipsis_literal_quoted`, `test_literal_with_special_characters`, `test_list_single_literal`) for the new `Optional(Sequence([...]))` wrapper.
- Added regex-level assertions that `[]`/`()` now match alongside the non-empty forms.

**One thing to call out explicitly**: this flips an assertion in `test_e2e_list_literal_empty_string`. That test previously asserted `not re.fullmatch(pattern, "[]")` for `list[Literal[""]]`. Its stated purpose (per its docstring) is that the empty *string* literal `""` gets quoted, i.e. `[""]` must have quotes; the `not match("[]")` line reads as an incidental consequence of the old empty-list-forbidden behavior rather than the point of the test. I changed it to assert `[]` matches (and kept an assertion that an unquoted item like `[x]` is still rejected). If the empty-list rejection was in fact intentional, this PR is the wrong call and #1940 is a "won't fix"; I filed the issue laying out the divergence either way so the decision sits with you. My reasoning for treating it as a bug is the dict-vs-list asymmetry plus the JSON-schema path already allowing empty arrays.

Ran the full `tests/types/test_dsl.py` (57 passed; the 2 failures are pre-existing `PermissionError`s in the `from_file` tests, a Windows temp-file quirk that reproduces on `main` without this change). `ruff check` clean on both files.


ErenAta16:
Rebased onto current `main` — this had fallen behind. Clean rebase, no conflicts, still a single commit touching `dsl.py` and its test.

Re-ran `tests/types/test_dsl.py` after the rebase: 59 passed.

@RobinPicard this is the same area as #1917 which you merged, so you've got the context. Happy to fold it in differently if you'd rather the empty-collection case be handled at a different level.

ErenAta16:
Re-checked this against current `main` (`be2cd151`). Still applies — 0 commits behind, `tests/types/test_dsl.py` at 59 passed, and the bug is unchanged:

```
main:            List[int] matches "[]" -> False
                 Tuple[int, ...] matches "()" -> False
                 Dict[str, int] matches "{}" -> True
```

The `Dict` row is the part that makes this look like an oversight rather than a decision — `_handle_dict` already wraps its content in `Optional(...)`, so the same emptiness is allowed there and only lists and tuples forbid it.

On blast radius, the generated pattern differs from `main` by exactly one optional group around the existing content, nothing else:

```
main    \(([+-]?(0|[1-9][0-9]*))(,\ ([+-]?(0|[1-9][0-9]*)))*\)
this PR \((([+-]?(0|[1-9][0-9]*))(,\ ([+-]?(0|[1-9][0-9]*)))*)?\)
```

so every non-empty string that matched before still matches — checked `[1]`, `[1, 2]`, `[1, 2, 3]`, `(1)`, `(1, 2)` against the new patterns, all still true, with `[]` and `()` newly accepted.

PR #1945, opened 2026-07-21
Add the credit_card custom type

dhanavanthesh:
Adds a `credit_card` custom type, in the same spirit as #1863 (semver, mac) and #1934 (hex_color, slug). It was proposed in a comment on #1303 but hadn't been picked up yet.

**What it does**

Matches payment card numbers for the common networks by issuer prefix and length: Visa (13/16/19), Mastercard (51-55 and 2221-2720), American Express, Diners Club, Discover (6011, 644-649, 65, with the 622126-622925 co-brand covered through the shared 62 prefix), JCB, Maestro and UnionPay. Since these prefixes are shared by credit, debit and prepaid cards, it validates the number format only, not the Luhn checksum or whether a card actually exists.

I kept it to the common network prefixes to match the lightweight style of the existing types; broader regional coverage could be a follow-up if you'd like it. Named `credit_card` per the #1303 proposal, but happy to rename to `card_number` if you prefer something more precise.

**Tests**

Added parametrized cases in `tests/types/test_custom_types.py`: accepted examples across the supported networks, the Mastercard 2-series and Discover 643/644 boundaries, and rejected inputs (separators, non-digit characters, unknown prefix, wrong length, empty). They fail on `main` and pass with the change. `ruff` is clean.

### DIFF
--- a/src/outlines/types/__init__.py
+++ b/src/outlines/types/__init__.py
@@ -88,6 +88,7 @@ __all__ = [
     "mac_address",
     "hex_color",
     "slug",
+    "credit_card",
     # Document-specific types
     "sentence",
     "paragraph",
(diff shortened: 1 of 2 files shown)


RobinPicard (approved):
Looks good, thanks!

PR #1949, opened 2026-07-22
fix(types): keep optional JSON Schema fields non-nullable

### LINKED ISSUE #1948: json_schema_dict_to_pydantic makes non-required fields nullable, widening generation constraints beyond the source schema
### What

`json_schema_dict_to_pydantic` makes every non-required field **nullable**, so the model it returns accepts `null` for fields the source schema does not allow to be `null`. For constrained generation this widens the grammar: the model can emit `"age": null` for a schema that forbids it.

`required` in JSON Schema controls whether a key may be **omitted**. It says nothing about whether the value may be `null`. The two get conflated here:

```python
# src/outlines/types/json_schema_utils.py, json_schema_dict_to_pydantic
if property not in required:
    field_definitions[property] = (Optional[typ], None)   # <- makes the VALUE nullable
else:
    field_definitions[property] = (typ, ...)
```

### Reproduction

```python
from outlines.types.json_schema_utils import json_schema_dict_to_pydantic

schema = {
    "type": "object",
    "properties": {
        "name": {"type": "string"},
        "age":  {"type": "integer"},
        "tags": {"type": "array", "items": {"type": "string"}},
    },
    "required": ["name"],
}

M = json_schema_dict_to_pydantic(schema)
print(M.model_json_schema()["properties"]["age"])
# {'anyOf': [{'type': 'integer'}, {'type': 'null'}], 'default': None, 'title': 'Age'}
```

Validating `{"name": "x", "age": null}`:

| | accepts `{"name":"x","age":null}` |
|---|---|
| source schema | **rejected** |
| schema regenerated from the converted model | **accepted** |

Same for `tags`, which becomes `anyOf: [array, null]`.

### Why this matters

The converted type is what feeds grammar generation, so this is not just a typing detail. Every optional field in a user's schema gains `null` as a legal generated value. A schema author who wrote `{"type": "integer"}` and left the field out of `required` was asking for "may be absent, otherwise an integer", and gets "may be absent, may be an integer, may be null".

### The file already documents the correct rule

The typeddict converter twenty lines above states the distinction, and gets it right:

```python
# NotRequired (PEP 655) marks the KEY optional; Optional only makes the
# value nullable, leaving the key required on a total=True TypedDict.
typ = NotRequired[typ]
```

Checked all three converters against the same input schema:

| converter | `age` after round-trip | accepts `null`? |
|---|---|---|
| typeddict (`NotRequired[int]`) | `{"type":"integer"}` | no, correct |
| dataclass (`int = None`) | `{"type":"integer","default":null}` | no, correct |
| **pydantic (`Optional[int]`)** | `{"anyOf":[{"type":"integer"},{"type":"null"}]}` | **yes, over-accepts** |

The pydantic path is the outlier.

### Suggested fix

Pydantic can express "omittable but not nullable" directly. A bare annotation with a default drops out of `required` and keeps its type:

```python
if property not in required:
    field_definitions[property] = (typ, None)
```

```
age -> {'default': None, 'title': 'Age', 'type': 'integer'}
required -> ['name']
  {"name":"x"}            accepted
  {"name":"x","age":3}    accepted
  {"name":"x","age":null} rejected      # matches the source schema
```

One caveat worth deciding deliberately rather than inheriting: this leaves `"default": null` sitting under a non-nullable type, which is the same cosmetic wart the dataclass converter has. It does not affect what is accepted, but if the schema is shown to users it reads oddly. `Field(default=None, exclude=...)` or omitting the default from the emitted schema would clean that up, if it is worth the extra machinery.

Genuinely nullable fields are unaffected, since a schema that means "may be null" says so with `{"type": ["integer","null"]}` or an explicit `anyOf`, and `schema_type_to_python` already handles that independently of `required`.

### Context

I originally reported this backwards in #1938, filing against the dataclass converter for *not* matching pydantic, and opened #1939 to make it match. @Sanjays2402 pointed out on that PR that `required` and nullability are different things, which is correct. Both are now closed. This issue is the real version of that report, filed against the converter that actually changes generation constraints.

Happy to open a PR if the suggested direction looks right, including whether to keep or suppress the `"default": null`.

Solaris-star:
## Summary
`json_schema_dict_to_pydantic` treated every non-`required` field as `Optional[T]`, which makes the **value** nullable. In JSON Schema, `required` only controls whether a key may be omitted — not whether `null` is legal.

That widened constrained generation: a schema with `"age": {"type": "integer"}` (not required) could emit `"age": null`.

The TypedDict converter already documents and implements the correct distinction (`NotRequired` vs value nullability). Align the Pydantic path.

## Fix
```python
# before
field_definitions[property] = (Optional[typ], None)
# after
field_definitions[property] = (typ, None)
```

Explicitly nullable schemas (`"type": ["string", "null"]` / `anyOf`) are unchanged — `schema_type_to_python` still produces `Optional[...]` independently of `required`.

## Test plan
- [x] `pytest tests/types/test_json_schema_utils.py`
- [x] New regression: optional int rejects `null`, allows omit; explicit nullable still accepts `null`

Fixes #1948

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -146,7 +146,10 @@ def json_schema_dict_to_pydantic(
     for property, details in properties.items():
         typ = schema_type_to_python(details, "pydantic")
         if property not in required:
-            field_definitions[property] = (Optional[typ], None)
+            # Omittable key, but keep the value type non-nullable unless the
+            # schema itself allows null (e.g. type: ["integer","null"]).
+            # Optional[typ] would widen generation to accept null.
+            field_definitions[property] = (typ, None)
         else:
             field_definitions[property] = (typ, ...)
 
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -2,7 +2,8 @@ import sys
 from dataclasses import is_dataclass
 from typing import Any, List, Literal, Optional, Union
 
-from pydantic import BaseModel, TypeAdapter
+import pytest
+from pydantic import BaseModel, TypeAdapter, ValidationError
 from pydantic_core import PydanticUndefined
 
 from outlines.types.json_schema_utils import (
@@ -447,10 +448,7 @@ def test_json_schema_dict_to_pydantic_optional_not_nullable():
     # omit ok, null rejected
     assert M(name="x").age is None
     assert M(name="x", age=3).age == 3
-    try:
+    with pytest.raises(ValidationError):
         M(name="x", age=None)
-        raise AssertionError("expected validation error for age=null")
-    except Exception:
-        pass
     # note may be null because schema allows it
     assert M(name="x", note=None).note is None
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -66,7 +66,7 @@ def test_schema_type_to_python_object():
     assert issubclass(pydantic_result, BaseModel)
     assert pydantic_result.__name__ == "TestObject"
     assert pydantic_result.model_fields["name"].annotation is str
-    assert pydantic_result.model_fields["age"].annotation == Optional[int]
+    assert pydantic_result.model_fields["age"].annotation is int
 
     # Typeddict caller
     typeddict_result = schema_type_to_python(schema, "typeddict")
@@ -228,9 +228,9 @@ def test_json_schema_dict_to_pydantic_basic():
     assert result.__name__ == "Person"
 
     assert result.model_fields["name"].annotation is str
-    assert result.model_fields["age"].annotation == Optional[int]
+    assert result.model_fields["age"].annotation is int
     assert result.model_fields["name"].default == PydanticUndefined
-    result.model_fields["age"].default is None
+    assert result.model_fields["age"].default is None
 
 
 def test_json_schema_dict_to_pydantic_nullable_type_array():
@@ -285,7 +285,7 @@ def test_json_schema_dict_to_pydantic_array_enum():
     assert issubclass(result, BaseModel)
     assert result.__name__ == "AnonymousPydanticModel"
 
-    assert result.model_fields["tags"].annotation == Optional[List[str]]
+    assert result.model_fields["tags"].annotation == List[str]
     assert result.model_fields["status"].annotation == Literal[("active", "inactive", "pending")]
     assert result.model_fields["tags"].default is None
     assert result.model_fields["status"].default == PydanticUndefined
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
The premise is right and I can corroborate it from the other side of this same file. Some context that is directly relevant here: I opened #1939 a few weeks ago arguing the opposite direction for the dataclass converter, wrapping non-required fields in `Optional[typ]`, and my justification was explicitly "the pydantic converter already does this, so match it". A reviewer pushed back, I ran the round-trip experiment, and it showed my change widened the accepted set rather than preserving it. I closed #1939 as a fidelity regression. So the precedent this PR is removing is the same precedent I leaned on, and it did not survive being measured. Worth knowing, because that `Optional` has been quietly acting as the reference implementation for the rest of the file.

I re-ran the measurement against this branch rather than reusing the old numbers. Ground truth is `jsonschema.validate` against the input schema, compared with `model_validate` on the converted model.

`{"age": {"type": "integer"}}`, not in `required`:

```
doc        source schema   main            this PR
omitted    accept          accept          accept
age = 3    accept          accept          accept
age = null reject          accept   <-- \  reject
```

That single disagreeing cell is the whole bug, and it closes here.

The part that matters more for outlines is that this is not only a validation-side difference, it reaches generation. Building the regex from the converted model:

```
                              main     this PR
{"name": "x", "age": 3}       True     True
{"name": "x", "age": null}    True     False
{"name": "x"}                 True     True
"null" appears in regex       True     False
```

So on `main` a constrained generation call against a schema that forbids null can still emit `"age": null`, and the key-omission behaviour that `required` actually governs is untouched by the fix. That is the argument I would put in the description, since it is the failure users would hit rather than a typing nicety.

One correction to the description. It says explicitly nullable schemas are preserved and gives both `"type": ["string","null"]` and `anyOf` as examples. The first is genuinely preserved:

```
"age": {"type": ["integer","null"]}   ->  Optional[int], null accepted, matches source
```

The `anyOf` form is not, or rather it looks preserved for the wrong reason:

```
"age": {"anyOf": [{"type":"integer"}, {"type":"null"}]}
  main:     Optional[Any]   round-trip {"anyOf": [{}, {"type":"null"}]}
  this PR:  Any             round-trip {"default": null}
```

Null is accepted in both cases, but only because the whole constraint has already collapsed to `Any` before this line runs, so the integer branch was lost too. That is pre-existing and looks like the territory #1963 is in, not something this PR introduces or needs to fix. I would just drop `anyOf` from the list of things the description claims are handled, otherwise it reads as tested behaviour when the passing result is incidental.

Second, a smaller thing worth a sentence in the comment. `(typ, None)` gives a round-trip schema of `{"type": "integer", "default": null}`, which is a default that its own type rejects. It does not affect generation, defaults are not constraints and I confirmed `null` is absent from the regex, but pydantic cannot express "key may be omitted, value not nullable" directly the way `NotRequired` does in the TypedDict path, and `(typ, None)` is the closest approximation rather than an exact encoding. Saying that in the comment would stop the next person from re-deriving it, and it is the honest version of "align the pydantic path with the TypedDict path".

Good catch on `result.model_fields["age"].default is None` missing its `assert`. That line was evaluating to a bare bool and asserting nothing, so the default was effectively untested. Unrelated to the fix but the right thing to sweep up while touching the file.

Tests look proportionate. `test_json_schema_dict_to_pydantic_optional_not_nullable` fails on `main` for the right reason, and updating the four existing annotation assertions rather than adding parallel ones keeps the file honest about what the converter now produces.

PR #1951, opened 2026-07-22
fix(types): preserve anyOf and oneOf constraints

### LINKED ISSUE #1950: schema_type_to_python drops anyOf/oneOf constraints to Any
### What

`schema_type_to_python` handles `type`, `enum`, `const`, and JSON Schema type arrays, but not `anyOf` or `oneOf`. A property expressed with either union keyword therefore falls through to `Any`.

That loses constraints when `json_schema_dict_to_pydantic` feeds the converted type back into structured generation.

### Reproduction

```python
from outlines.types.json_schema_utils import schema_type_to_python

print(schema_type_to_python({
    "anyOf": [{"type": "integer"}, {"type": "string"}],
}, "pydantic"))
# typing.Any (expected Union[int, str])

print(schema_type_to_python({
    "oneOf": [{"type": "boolean"}, {"type": "null"}],
}, "pydantic"))
# typing.Any (expected Optional[bool])
```

The resulting Pydantic schema has no value constraint at all:

```python
M = json_schema_dict_to_pydantic({
    "type": "object",
    "properties": {
        "value": {"anyOf": [{"type": "integer"}, {"type": "string"}]},
    },
    "required": ["value"],
})
print(M.model_json_schema()["properties"]["value"])
# {"title": "Value"}
```

### Expected

Map each branch recursively and combine them with `Union`, just as the existing JSON Schema type-array path does. For Python typing/Pydantic, `oneOf` cannot preserve exclusivity for overlapping branches, but retaining the union of allowed value types is still strictly better than widening to `Any`.

Solaris-star:
## Summary
`schema_type_to_python` handled `type`, `enum`, `const`, and type arrays, but silently fell through to `Any` for JSON Schema `anyOf` and `oneOf`.

That dropped all value constraints when `json_schema_dict_to_pydantic` converted a schema containing a union. Structured generation could then accept arbitrary values for a field whose schema specified a finite union.

## Fix
- Recursively convert each `anyOf` / `oneOf` branch
- Combine the resulting Python types with `Union`
- Preserve explicit nullable branches as `Optional[...]`

`oneOf` exclusivity cannot be represented by Python `Union` when branches overlap, but retaining the allowed type union is strictly safer than widening to `Any`.

## Verification
- [x] `pytest tests/types/test_json_schema_utils.py` — 24 passed
- [x] pre-commit: mypy, ruff, whitespace, merge checks all passed

Fixes #1950

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -42,6 +42,14 @@ def schema_type_to_python(
         # widening the value back to its bare type.
         return Literal[schema["const"]]
 
+    for keyword in ("anyOf", "oneOf"):
+        if keyword in schema:
+            members = tuple(
+                schema_type_to_python(member, caller_target_type)
+                for member in schema[keyword]
+            )
+            return Union[members] if members else Any  # type: ignore
+
     t = schema.get("type")
 
     if isinstance(t, list):
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -44,9 +44,17 @@ def schema_type_to_python(
 
     for keyword in ("anyOf", "oneOf"):
         if keyword in schema and "type" not in schema:
+            branches = schema[keyword]
+            # Boolean schemas are legal JSON Schema subschemas. ``true`` is
+            # always valid, so the best Python-type approximation is Any;
+            # ``false`` contributes no branch. Ignore both rather than
+            # passing a bool into the dict-oriented recursive converter.
+            if any(branch is True for branch in branches):
+                return Any
             members = tuple(
                 schema_type_to_python(member, caller_target_type)
-                for member in schema[keyword]
+                for member in branches
+                if isinstance(member, dict)
             )
             return Union[members] if members else Any  # type: ignore
 
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -43,7 +43,7 @@ def schema_type_to_python(
         return Literal[schema["const"]]
 
     for keyword in ("anyOf", "oneOf"):
-        if keyword in schema:
+        if keyword in schema and "type" not in schema:
             members = tuple(
                 schema_type_to_python(member, caller_target_type)
                 for member in schema[keyword]
--- a/src/outlines/exceptions.py
+++ b/src/outlines/exceptions.py
@@ -349,7 +349,7 @@ def _provider_exception_map(provider: str) -> dict[type, type[APIError]]:
         }
 
     if provider == "dottxt":
-        import urllib3.exceptions as urllib3_exc
+        urllib3_exc = importlib.import_module("urllib3.exceptions")
         return {
             # NewConnectionError subclasses ConnectTimeoutError, so it must come first
             urllib3_exc.NewConnectionError: APIConnectionError,
--- a/src/outlines/exceptions.py
+++ b/src/outlines/exceptions.py
@@ -349,6 +349,9 @@ def _provider_exception_map(provider: str) -> dict[type, type[APIError]]:
         }
 
     if provider == "dottxt":
+        # Import the package first so an unavailable/blocked optional SDK is
+        # detected even when its exceptions submodule is still cached.
+        importlib.import_module("urllib3")
         urllib3_exc = importlib.import_module("urllib3.exceptions")
         return {
             # NewConnectionError subclasses ConnectTimeoutError, so it must come first
--- a/tests/test_exceptions.py
+++ b/tests/test_exceptions.py
@@ -706,25 +706,21 @@ class TestExceptionMapTGI:
 
 
 class TestExceptionMapDottxt:
-    @pytest.fixture(autouse=True)
-    def _require_dottxt(self):
-        pytest.importorskip("urllib3")
+    @pytest.fixture
+    def urllib3_exceptions(self):
+        return pytest.importorskip("urllib3.exceptions")
 
-    def test_connect_timeout(self):
-        from urllib3.exceptions import ConnectTimeoutError
-        _check_mapping(ConnectTimeoutError, "dottxt", APITimeoutError)
+    def test_connect_timeout(self, urllib3_exceptions):
+        _check_mapping(urllib3_exceptions.ConnectTimeoutError, "dottxt", APITimeoutError)
 
-    def test_read_timeout(self):
-        from urllib3.exceptions import ReadTimeoutError
-        _check_mapping(ReadTimeoutError, "dottxt", APITimeoutError)
+    def test_read_timeout(self, urllib3_exceptions):
+        _check_mapping(urllib3_exceptions.ReadTimeoutError, "dottxt", APITimeoutError)
 
-    def test_max_retry_error(self):
-        from urllib3.exceptions import MaxRetryError
-        _check_mapping(MaxRetryError, "dottxt", APIConnectionError)
+    def test_max_retry_error(self, urllib3_exceptions):
+        _check_mapping(urllib3_exceptions.MaxRetryError, "dottxt", APIConnectionError)
 
-    def test_new_connection_error(self):
-        from urllib3.exceptions import NewConnectionError
-        _check_mapping(NewConnectionError, "dottxt", APIConnectionError)
+    def test_new_connection_error(self, urllib3_exceptions):
+        _check_mapping(urllib3_exceptions.NewConnectionError, "dottxt", APIConnectionError)
 
 
 # vLLM and SGLang both reuse the OpenAI SDK client, so they share the OpenAI
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -26,6 +26,29 @@ def test_schema_type_to_python_simple_types():
     assert schema_type_to_python({"type": "object"}, "foo") is Any
     assert schema_type_to_python({}, "pydantic") is Any
 
+    assert schema_type_to_python(
+        {"anyOf": [{"type": "integer"}, {"type": "string"}]}, "pydantic"
+    ) == Union[int, str]
+    assert schema_type_to_python(
+        {"oneOf": [{"type": "boolean"}, {"type": "null"}]}, "pydantic"
+    ) == Optional[bool]
+
+
+def test_json_schema_dict_to_pydantic_union_keywords():
+    schema = {
+        "type": "object",
+        "properties": {
+            "value": {"anyOf": [{"type": "integer"}, {"type": "string"}]},
+        },
+        "required": ["value"],
+    }
+
+    result = json_schema_dict_to_pydantic(schema, "UnionValue")
+    assert result.model_fields["value"].annotation == Union[int, str]
+    assert result.model_json_schema()["properties"]["value"]["anyOf"]
+    assert result(value=1).model_dump()["value"] == 1
+    assert result(value="one").model_dump()["value"] == "one"
+
 
 def test_schema_type_to_python_enum():
     schema = {"enum": ["red", "green", "blue"]}
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -50,6 +50,25 @@ def test_json_schema_dict_to_pydantic_union_keywords():
     assert result(value="one").model_dump()["value"] == "one"
 
 
+def test_schema_type_to_python_union_keyword_with_outer_type():
+    """Sibling ``type`` constraints must not be widened by union branches."""
+
+    assert (
+        schema_type_to_python(
+            {"type": "string", "anyOf": [{"minLength": 1}, {"const": "x"}]},
+            "pydantic",
+        )
+        is str
+    )
+    assert (
+        schema_type_to_python(
+            {"type": "string", "oneOf": [{"const": "x"}, {"const": 1}]},
+            "pydantic",
+        )
+        is str
+    )
+
+
 def test_schema_type_to_python_enum():
     schema = {"enum": ["red", "green", "blue"]}
     result = schema_type_to_python(schema, "pydantic")
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -69,6 +69,33 @@ def test_schema_type_to_python_union_keyword_with_outer_type():
     )
 
 
+def test_schema_type_to_python_union_keyword_boolean_subschemas():
+    """Boolean JSON Schema branches must not reach the dict converter."""
+
+    assert (
+        schema_type_to_python(
+            {"anyOf": [{"type": "integer"}, False]},
+            "pydantic",
+        )
+        is int
+    )
+    assert (
+        schema_type_to_python(
+            {"oneOf": [{"type": "string"}, False]},
+            "pydantic",
+        )
+        is str
+    )
+    assert (
+        schema_type_to_python(
+            {"anyOf": [{"type": "integer"}, True]},
+            "pydantic",
+        )
+        is Any
+    )
+    assert schema_type_to_python({"oneOf": [True]}, "pydantic") is Any
+
+
 def test_schema_type_to_python_enum():
     schema = {"enum": ["red", "green", "blue"]}
     result = schema_type_to_python(schema, "pydantic")


src/outlines/types/json_schema_utils.py
```diff
@@ -42,6 +42,14 @@ def schema_type_to_python(
         # widening the value back to its bare type.
         return Literal[schema["const"]]
 
+    for keyword in ("anyOf", "oneOf"):
+        if keyword in schema:
+            members = tuple(
+                schema_type_to_python(member, caller_target_type)
+                for member in schema[keyword]
+            )
```
Sanjays2402, line 58: The early return drops sibling constraints, but JSON Schema combines them with `anyOf`. For `{"type":"string","anyOf":[{"const":"ok"},{"const":1}]}`, this produces `Literal["ok", 1]` and Pydantic accepts `1`, even though the schema rejects it. Can the outer constraints be preserved here?

Solaris-star:
Good catch, and you're right that the sibling `type` is dropped. I confirmed it locally:

```python
schema_type_to_python({"type": "string", "anyOf": [{"const": "ok"}, {"const": 1}]}, "pydantic")
# -> Union[Literal["ok"], Literal[1]]   # the outer "type": "string" is lost, so 1 slips through
```

The reason I kept the early return is that combining `type` with `anyOf`/`oneOf` is an **intersection** (AND) in JSON Schema — the value must satisfy *both* the bare `type` and one of the union members. Python's typing system has no general intersection type, so there's no faithful annotation for the combined constraint: `Union[...]` alone is too loose (your `1` case), and narrowing to just `type` would be too strict (drops the enum-like refinement).

This PR is scoped to #1950, which is specifically about `anyOf`/`oneOf` collapsing to `Any` when they appear on their own — the common Pydantic-emitted shape. The `type` + `anyOf` intersection is a real but separate limitation that predates this change and can't be resolved without either a runtime validator on top of the generated type or an intersection representation.

Happy to open a follow-up issue documenting the intersection gap so it's tracked rather than silently lost. Would you prefer I add a short comment in the code flagging that sibling scalar constraints alongside `anyOf`/`oneOf` are not intersected, so the limitation is visible at the call site?

src/outlines/types/json_schema_utils.py
```diff
@@ -42,6 +42,14 @@ def schema_type_to_python(
         # widening the value back to its bare type.
         return Literal[schema["const"]]
 
+    for keyword in ("anyOf", "oneOf"):
```
Sanjays2402, line 45: this returns before `type` is consulted, so the constraint-only form of anyOf/oneOf loses the declared type: `{"type": "string", "anyOf": [{"minLength": 1}, {"const": "x"}]}` has members with no `type`, each maps to `Any`, and `Union[Any, Any]` collapses to plain `Any` so the field stops being a `str`. Same widening the `const` comment above is guarding against. Would it be safer to only take this branch when `type` is absent, or to intersect with the outer type the way the `isinstance(t, list)` branch reuses `{**schema, ...}`?

Solaris-star: You were right that the early union return widened sibling `type` constraints. I pushed `692f64e`: `anyOf`/`oneOf` are converted to a Python union only when no outer `type` is present; otherwise the existing outer-type mapping remains the conservative approximation of the JSON Schema intersection. Added regressions for untyped branches and a conflicting `const`. The type-utils module is 25/25 green, and the repo-pinned pre-commit/mypy/ruff hooks pass.

ErenAta16 (commented):
The core of this works. Ran it against a clone of `main` at `be2cd15` and then against `692f64e`:

```
                                   main    this PR
anyOf [str, int]                   Any     Union[str, int]
oneOf [str, null]                  Any     Optional[str]
anyOf [int, null]                  Any     Optional[int]
nested anyOf                       Any     int
plain type / enum / array          unchanged
```

Guarding on `"type" not in schema` is the right precedence choice, since a schema carrying both is already ambiguous and deferring to `type` keeps the existing path intact.

There is a regression I would want fixed before this lands, though. The members go straight into the recursive call with no guard on what they are, and JSON Schema allows a subschema to be a boolean, not just an object. Draft 6 onward defines `true` as the always-valid schema and `false` as the never-valid one, so `{"anyOf": [{"type": "integer"}, true]}` is a legal document. Today it degrades to `Any`; on this branch it raises:

```
schema_type_to_python({"anyOf": [{"type": "integer"}, True]}, "pydantic")
# main:     typing.Any
# this PR:  TypeError: argument of type 'bool' is not iterable
```

It comes through the realistic entry point too, not just the helper:

```python
json_schema_dict_to_pydantic({
    "type": "object",
    "properties": {"x": {"anyOf": [{"type": "integer"}, True]}},
    "required": [],
})
# main:     x -> Optional[Any]
# this PR:  TypeError: argument of type 'bool' is not iterable
```

A non-dict member reaches `"enum" in schema` and `schema.get(...)` inside the recursive call, so a string member fails a bit differently but just as hard (`AttributeError: 'str' object has no attribute 'get'`). Turning a permissive fallback into an exception is the part that concerns me more than the specific input, because callers converting third-party schemas currently get a loose model and would now get a traceback.

Filtering the members to dicts before the comprehension is enough, and it also keeps the `Any` fallback meaningful when everything is filtered out.

The other thing worth raising is overlap. #1963 covers the same two keywords and I ran both branches through the same matrix:

```
                        this PR                 #1963
anyOf [str, int]        Union[str, int]         Union[str, int]
oneOf [str, null]       Optional[str]           Optional[str]
anyOf [int, null]       Optional[int]           Optional[int]
allOf single            Any                     int
$ref to nested model    Any                     resolved model
anyOf with boolean      TypeError               int
```

On `anyOf` and `oneOf` the two are behaviourally identical, so this is a subset of #1963 rather than a complementary change, and #1963 already filters non-dict members so it does not have the crash above. They also touch the same function in the same file, so they will conflict.

I do not think that makes this PR redundant on the process side, it is the smaller and easier change to review, and if #1963 stalls this is the one that should go in. But the two should be decided together rather than reviewed as independent contributions. Flagging for the maintainers.

Tests are fine as far as they go. If this stays separate, a case with a boolean subschema member would pin the behaviour the fix above restores.

ErenAta16 (commented):
Confirming this fix and adding the round-trip case, since it's the one that makes the priority clear.

Pydantic emits `anyOf` for `Optional[T]` and `Union[...]`, so this isn't only a hand-written-schema concern — any model going through `json_schema_dict_to_pydantic` loses those fields:

```python
class Source(BaseModel):
    maybe_int: Optional[int]
    either: Union[int, str]
    plain: int
```

| field | on `main` | with this PR |
|---|---|---|
| `maybe_int` | `Any` | `Optional[int]` |
| `either` | `Any` | `Union[int, str]` |
| `plain` | `int` | `int` |

On `main` the regenerated schema for `maybe_int` is just `{"title": "Maybe Int"}` — nothing left to constrain generation with. With the fix it round-trips back to `{"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "Maybe Int"}`, identical to what Pydantic produced.

The `"type" not in schema` guard is the right call. I had written the same fix without it (#1973, now closed in favour of this one) and it diverges on `{"type": "object", "anyOf": [...]}`: without the guard the declared `type` is dropped in favour of the branches, which is a widening. JSON Schema treats both as constraints that apply together, so letting `type` win is the conservative reading.

```
{"type": "object", "anyOf": [{"type":"integer"},{"type":"string"}]}
  without the guard -> Union[int, str]     (type discarded)
  this PR           -> falls through to the type dispatch
```

Empty branch lists agree either way: `{"anyOf": []}` gives `Any` here, and `Any` via the `type` fallthrough without the guard.

If it's useful, the tests I had written cover a few cases beyond the ones here — both keywords parametrised over the same assertion, `null`-branch to `Optional`, `anyOf` agreeing with the equivalent `{"type": ["integer","string"]}` spelling, recursive branch mapping (`anyOf` of `array`/`string` to `Union[List[int], str]`), and the round trip above. Happy to open them as a follow-up against this branch rather than a competing PR.

percymcn:
I reimplemented this diff against `outlines` on PyPI and ran it before writing anything, so first the fair verdict: **it works, and it fixes what it says it fixes.** All four of the assertions this PR adds pass with the patched function, and the sibling-`type` guard is the right call — `{"type": "string", "oneOf": [{"const": "x"}, {"const": 1}]}` correctly stays `str` rather than widening to `Union[str, int]`.

The additive part is about **which spellings the fix reaches**, because the failure mode you describe in the summary — *"silently fell through to `Any` … structured generation could then accept arbitrary values"* — survives this PR for the single most common schema shape there is.

### `$ref` still falls through to `Any`, and that is every nested model

Measured with this diff applied:

| subschema | post-PR |
|---|---|
| `{"anyOf": [{"type":"integer"},{"type":"string"}]}` | `Union[int, str]` ✅ (this PR) |
| `{"oneOf": [{"type":"boolean"},{"type":"null"}]}` | `Optional[bool]` ✅ (this PR) |
| `{"allOf": [{"type":"integer"}]}` | `Any` |
| `{"allOf": [{"type":"integer"},{"minimum":0}]}` | `Any` |
| `{"$ref": "#/$defs/S"}` | `Any` |
| `{"$ref": "#/$defs/S", "description": "d"}` | `Any` |
| `{"allOf": [{"$ref": "#/$defs/S"}]}` | `Any` |
| `{"type": "string"}` (control) | `str` ✅ |

`{"$ref": "#/$defs/Inner"}` is verbatim what pydantic v2 (2.13.4) emits for **any** nested model, and `{"title": ..., "description": ..., "allOf": [{"$ref": "#/definitions/Inner"}]}` is verbatim what pydantic v1 (1.10.22) emits for the same thing — so both major versions' nested-model spelling lands in the `Any` branch.

End to end, with the PR applied:

```python
class Inner(BaseModel): b: str
class M(BaseModel):
    a: str
    inner: Inner

Model = json_schema_dict_to_pydantic(M.model_json_schema(), "RoundTrip")
Model.model_fields["inner"].annotation
# -> typing.Any
Model.model_json_schema()["properties"]["inner"]
# -> {"title": "Inner"}          # constrains nothing

Model(a="x", inner=123)                       # accepted
Model(a="x", inner="not-an-object")           # accepted
Model(a="x", inner={"COMPLETELY": "unrelated"})  # accepted
Model(a="x", inner=None)                      # accepted
```

The source schema requires `inner` to be an object with a required string `b`. The emitted schema for that field is `{"title": "Inner"}`, so for a structured-generation backend the field is unconstrained — the same outcome this PR is closing for unions.

(Control, so this isn't my harness: on the same run `{"anyOf": [{"type":"integer"},{"type":"string"}]}` becomes `Union[int, str]` and correctly rejects a dict. The `anyOf` fix is real and the harness can see it.)

### Why `$ref` can't just be another `for keyword in (...)` line

`schema_type_to_python(details, "pydantic")` receives only the subschema. `json_schema_dict_to_pydantic` iterates `properties` and never looks at `$defs`, so by the time the pointer is seen there is nothing to resolve it against. Adding `"$ref"` to the loop above would need the **root** threaded through as a parameter (and a `seen` set, since a self-referential model is a cycle and the recursion would not terminate on `{"$defs": {"Node": {"properties": {"child": {"$ref": "#/$defs/Node"}}}}}`).

`allOf` is cheaper and independent: a **single-member** `allOf` is the same intersection as its member, so it can be forwarded exactly the way this PR forwards a one-branch union. A **multi-member** `allOf` genuinely has no Python-type representation — but the summary's own principle ("retaining the allowed type union is strictly safer than widening to `Any`") argues for at least taking the first typed member rather than dropping to `Any`, or raising, rather than silently accepting everything.

### One test-design note

Every fixture added here uses inline branches with no `$ref` and no `allOf`:

```python
{"anyOf": [{"type": "integer"}, {"type": "string"}]}
{"oneOf": [{"type": "boolean"}, {"type": "null"}]}
{"type": "string", "anyOf": [{"minLength": 1}, {"const": "x"}]}
```

so a regression suite built on them stays green while the `$ref` hole ships. The discriminating fixture is one line — a schema whose property is `{"$ref": "#/$defs/Inner"}` with `Inner` in `$defs` — and the assertion has to be that an **illegal** value is *rejected*, since over-acceptance is precisely what goes wrong here; asserting on a legal value passes either way.

ErenAta16 (commented):
Re-measured at `692f64ee`, because the thread has moved on and the regression I raised on 31 July is still open. No commit since 25 July, and it still reproduces:

```
                       main    this branch
anyOf [int, true]      Any     TypeError: argument of type 'bool' is not iterable
anyOf [int, false]     Any     TypeError
anyOf [true]           Any     TypeError
oneOf [str, true]      Any     TypeError
anyOf [int, str]       Any     Union[int, str]    <- the actual fix, working
```

The fix is sound. The only problem is that members go into the recursive call unguarded, and draft 6 onward allows a subschema to be a boolean, so a legal document now raises where it used to degrade.

On @percymcn's `$ref` analysis: I agree with the measurements, but I do not think this PR is the place for it. #1963 already implements `$ref` resolution, root threading and a cycle guard in this same function, at 207 added lines against the 8 here. Two open PRs rewriting `schema_type_to_python` will conflict on merge and the smaller one gets redone regardless of which lands first.

Which points at the cheap resolution, because #1963 already carries the exact guard this PR is missing:

```python
for subschema in schema[keyword]
if isinstance(subschema, dict)
```

So the minimal path is to add that one clause here, keep this PR at nine lines, and let #1963 carry `$ref`. If maintainers would rather take #1963 whole, this can close in its favour instead: it subsumes the `anyOf`/`oneOf` behaviour and keeps the sibling-`type` guard, which as @percymcn notes is the right call and worth not losing.

One caveat if that clause gets copied across. Dropping non-dict members is not correct for all three combinators, and applying it uniformly introduces a different bug. I have written that up on #1963 rather than duplicate it here, since that is the branch it currently affects.

Solaris-star:
Addressed the boolean-subschema regression raised in the 14 Aug review and pushed it as `73e6e7b83284cd1599a90bd9cf14e34b0c7367b6`.

The converter now:

- ignores `false` branches instead of passing a boolean into the dict-oriented recursive converter;
- returns the conservative `Any` fallback when an always-valid `true` branch is present;
- keeps the existing `Any` fallback if no dictionary branches remain.

Added coverage for `anyOf`/`oneOf` with `true` and `false` boolean subschemas. Local validation on the pushed head:

- `PYTHONPATH=src uv run --no-project --python 3.12 --with pytest --with pydantic --with typing-extensions --with jinja2 --with cloudpickle --with diskcache --with jsonschema --with pillow --with outlines-core==0.2.14 --with genson python -m pytest tests/types/test_json_schema_utils.py -q` — **26 passed**.
- `uv run --no-project --python 3.12 --with pre-commit pre-commit run --files src/outlines/types/json_schema_utils.py tests/types/test_json_schema_utils.py` — all configured hooks **passed**.

The commit is authored as `Solaris-star <820622658@qq.com>`, includes `Signed-off-by`, and GitHub reports its SSH signature as `verified: true` / `reason: valid`. At the time of this note the five GitHub checks had started and were still in progress; I have not treated them as passing. I did not run the full repository suite locally.

ErenAta16 (commented):
Verified at `73e6e7b8`. The regression is gone and the boolean handling is correct, including the part I only described rather than specified.

```
case                   main       previous head    73e6e7b8    correct
anyOf [int, true]      Any        TypeError        Any         Any
anyOf [int, false]     Any        TypeError        int         int
anyOf [true]           Any        TypeError        Any         Any
anyOf [false]          Any        TypeError        Any         Any
oneOf [str, true]      Any        TypeError        Any         Any
oneOf [str, false]     Any        TypeError        str         str
anyOf [int, str]       Any        Union[int, str]  Union[int, str]
sibling type guard     str        str              str         str
```

You got the asymmetry right, which is the part that is easy to get wrong: in a union `true` is absorbing so the answer has to widen to `Any`, while `false` contributes nothing and is simply dropped. Treating them alike would have been the obvious reading of "ignore non-dict members" and it would have been wrong in one direction.

`branch is True` rather than a truthiness test is also the right call. `1` and `1.0` are equal to `True` under `==`, and neither is a valid subschema, so identity is what you want here.

The two spellings now agree, which was the property I actually cared about:

```
anyOf int + always-valid    {} spelling -> Union[int, Any]     true spelling -> Any
oneOf str + always-valid    {} spelling -> Union[str, Any]     true spelling -> Any
```

Both collapse to "accepts anything", so a schema author gets the same constraint whichever way they write the always-valid branch.

One thing for whoever sequences these two PRs. #1963 is still at `0d43be2a` and has the inverse of this bug: it filters non-dict members uniformly across `anyOf`, `oneOf` and `allOf`, so `{"anyOf": [{"type": "integer"}, true]}` comes out as `int` there and silently rejects values the source schema allows. I wrote it up on that thread. It matters here because both PRs rewrite `schema_type_to_python`, so whichever lands second will be rebased onto the other, and the boolean handling in this PR is the version worth keeping.

Nothing further from me on this one.

Solaris-star:
CI follow-up for live head `73e6e7b83284cd1599a90bd9cf14e34b0c7367b6`:

- The targeted regression command remains green: `PYTHONPATH=src uv run --no-project --python 3.12 --with pytest --with pydantic --with typing-extensions --with jinja2 --with cloudpickle --with diskcache --with jsonschema --with pillow --with outlines-core==0.2.14 --with genson python -m pytest tests/types/test_json_schema_utils.py -q` — 26 passed.
- Scoped pre-commit on the two changed files passed all configured hooks.
- The GitHub `Check the code style` job failed only in the repository-wide mypy hook with three `import-not-found` errors for optional `urllib3` in the unchanged `src/outlines/exceptions.py` and `tests/test_exceptions.py`; ruff passed. The PR diff does not touch those files or the pre-commit configuration.
- The other head checks passed: Python 3.10/3.13 tests, coverage, wheel/sdist, and documentation.
- The latest successful main-branch Tests run before this head was `31185764335` on `7d068478851f7ba76cb53997673d57f77b2d6f84`, so this looks like a style-environment dependency issue rather than a regression in the PR. I attempted to rerun the failed job, but GitHub rejected it because this account lacks repository admin rights.

I have not added an unrelated dependency/config change to this focused PR.

ErenAta16 (commented):
Coming back on the CI point, because I could not tell from the outside whether it was yours and it turns out it is not, and there is already a fix sitting in the queue.

Pulled the job log for `73e6e7b8`. The whole failure is:

```
mypy.....................................................................Failed
src/outlines/exceptions.py:352: error: Cannot find implementation or library stub for module named "urllib3.exceptions"  [import-not-found]
src/outlines/exceptions.py:352: error: Cannot find implementation or library stub for module named "urllib3"  [import-not-found]
tests/test_exceptions.py:714: error: Cannot find implementation or library stub for module named "urllib3.exceptions"  [import-not-found]
```

Three errors, all in files this PR does not touch. The job runs `pre-commit run --all-files`, so mypy sees the whole tree regardless of the diff, and the hook runs in its own isolated environment:

```yaml
- id: mypy
  additional_dependencies: [types-tqdm, types-Pillow]
```

`urllib3` is not in that list, so the import in `exceptions.py` cannot resolve. Nothing about the change here reaches it.

**It is recent rather than permanent.** Every PR merged up to 7 August has a green style check without touching the config:

```
#1984  merged 2026-08-07  success
#1966  merged 2026-08-06  success
#1979  merged 2026-08-05  success
#1972  merged 2026-08-03  success
```

so something in the resolved environment changed after that and made the import unresolvable where it previously was not.

**The fix already exists, twice.** #2003 and #2009 are both a single line:

```diff
- additional_dependencies: [types-tqdm, types-Pillow]
+ additional_dependencies: [types-tqdm, types-Pillow, urllib3]
```

and both are green, with their logs showing the hook initialising as `types-tqdm,types-Pillow,urllib3`. That is the cleanest proof that the red is the config and not anyone's code: the only PRs currently passing this job are the two that patch it.

Worth someone noticing that those two are the same one-line change by different authors, so one of them can close.

Nothing for you to do here. Rebasing will not help, since the config is the same on `main`, and I would not add the dependency to this PR either, because that would tangle a types fix with a CI fix and duplicate a third time. This just needs #2003 or #2009 to land first.

@RobinPicard the style job is red on everything that does not carry that one line, which is most of the open queue right now.

ErenAta16 (commented):
Your reading of that style failure was right, and it has since been fixed on main,
so this is resolvable with a rebase rather than anything on your branch.

The commit is `c52af84`, "Give the mypy hook urllib3 so the style check passes",
landed 14 August, which is the day before you posted the analysis above. Its
message describes the same chain you worked out from the outside:

```
5527d3e3 added `import urllib3.exceptions` to the dottxt branch of
exceptions.py, but the pre-commit mypy hook runs in an isolated env
built from additional_dependencies, which does not include urllib3.
mypy therefore cannot resolve the import and the style job fails on
main and on every PR branched from it.
```

The fix is one line, `additional_dependencies: [types-tqdm, types-Pillow, urllib3]`.

GitHub currently reports this PR as `behind`, with the head still at `da308f28`,
so the branch predates that commit and is carrying the broken hook environment.
Rebasing onto current main should take the style job green without touching your
diff, which also settles the thing you were careful about: you will not need to add
an unrelated config change to a focused PR, because someone else already made it in
the right place.

Worth noting for the record that you diagnosed a repo-wide failure correctly from a
failing job you could not rerun, and said so without touching anything outside your
scope. That is the right call and it is the reason this is a one-command fix now
rather than a config change nobody could review.

Solaris-star:
Rebased onto current main in cc312b88d5fc0efc3b5d9518c826ae8023b1c862. The repo-wide urllib3/mypy hook fix is now included from main; no unrelated config change was added to this focused PR.

Verified with PYTHONPATH=src python3.11 -m pytest -q tests/types/test_json_schema_utils.py: 26 passed. The branch retains the boolean-subschema handling and the outer-type guard discussed in review.

Solaris-star (commented):
Rebased onto current main and verified the focused JSON-schema suite: 26 passed. The outer-type guard and boolean-subschema handling are present in the current head.

src/outlines/types/json_schema_utils.py
Solaris-star: Addressed by limiting the anyOf/oneOf shortcut to schemas without an outer type. The existing type-dispatch path now preserves the sibling constraint, including the constraint-only branch case. The focused suite passes after rebase.

src/outlines/types/json_schema_utils.py
Solaris-star: Yes: the guard is now `keyword in schema and "type" not in schema`, so a declared outer type is not widened or dropped. The focused suite passes after rebase, including the constraint-only anyOf/oneOf cases.

PR #1954, opened 2026-07-23
feat(types): add UK and Canadian locale types (phone, postcode/postal_code)

### LINKED ISSUE #1303: Add more custom types

Goutham-Annem:
## Summary

Adds locale-specific regex patterns for the United Kingdom and Canada, following the same pattern as the existing `us` locale module.

**UK (`outlines.types.locale.uk`):**
- `phone_number` — supports `+44` international prefix and `0`-prefixed local numbers, covering mobile (07xxx), London (020x), and generic landline formats
- `postcode` — matches standard UK postcode format (e.g., `SW1A 2AA`, `EC1A1BB`)

**Canada (`outlines.types.locale.ca`):**
- `phone_number` — NANP format with optional `+1` prefix, parenthesized or bare area codes, dot/dash/space separators
- `postal_code` — alternating letter-digit format with restricted first-letter set (e.g., `K1A 0A6`, `K1A0A6`)

Closes #1303

## Changes

- `src/outlines/types/locale/uk.py` — new file
- `src/outlines/types/locale/ca.py` — new file
- `src/outlines/types/locale/__init__.py` — export `ca` and `uk`
- `tests/types/test_custom_types.py` — 12 new parametrized test cases covering valid and invalid inputs for all four new patterns

## Testing

All existing tests continue to pass. New test cases verify:
- Valid UK phone numbers with `+44` and `07xxx` prefixes
- Valid UK postcodes with and without space
- Invalid inputs correctly rejected
- Valid Canadian phone numbers (parenthesized area code, `+1` prefix)
- Valid Canadian postal codes with and without space
- Invalid inputs correctly rejected

Note: This PR was developed with AI assistance.


ErenAta16 (commented):
Ran the four patterns at `497f80f` against real-world cases rather than only the ones in the tests. The phone patterns hold up and the UK postcode handles the shapes it needs to, including the short `M1 1AE` form and the unspaced `EC1A1BB`. A couple of things are looser than they should be, and one of them I think is a genuine slip rather than a judgement call.

**The Canadian postal code is inconsistent with itself.**

```
postal_code = Regex(r"[ABCEGHJKLMNPRSTVXY]\d[A-Z]\s?\d[A-Z]\d")
```

The first character class is exactly right. It is the 18-letter set Canada Post uses, leaving out D, F, I, O, Q and U because they are confusable with digits or other letters, and also leaving out W and Z which are never used in the leading position. But the other two letter positions are plain `[A-Z]`, and D, F, I, O, Q and U are excluded from *every* position, not just the first. So the pattern rejects them where it already knows better and accepts them everywhere else:

```
D1A 0A6   rejected   correct, D is not a valid leading letter
K1D 0A6   accepted   should be rejected
K1F 0A6   accepted   should be rejected
K1I 0A6   accepted   should be rejected
K1O 0A6   accepted   should be rejected
K1Q 0A6   accepted   should be rejected
K1U 0A6   accepted   should be rejected
K1A 0D6   accepted   should be rejected
K1W 0A6   accepted   correct, W is fine outside the first position
```

`[A-CEGHJ-NPR-TV-Z]` in positions three and five covers it: same exclusions, but W and Z stay allowed since only the leading position bars them.

This matters more here than it would in a validation library. These patterns drive constrained generation, so a loose class is not a permissive fallback, it is a set of characters the model is actively allowed to emit. A generated `K1O 0A6` looks like a postal code and is not one, and nothing downstream will catch it.

**Two smaller ones, more arguable.**

UK postcode, `[A-Z]{1,2}\d[A-Z\d]?\s?\d[A-Z]{2}`:

```
QQ1 1AA   accepted   Q, V and X are not used as the first letter
SW1A 2CC  accepted   the final pair excludes C, I, K, M, O and V
```

Canadian phone, `(\+1\s?)?(\([0-9]{3}\)|[0-9]{3})...`:

```
012 555 1234   accepted   NANP area codes are [2-9]\d\d, they cannot start with 0 or 1
```

I am less exercised about these two. Simplified UK postcode regexes that accept `QQ1 1AA` are extremely common and tightening the full set makes the pattern considerably harder to read, so if you want to leave it that is a defensible call. The NANP one is a two-character change (`[2-9][0-9]{2}` in both branches) and the same restriction applies to the exchange code, so it is cheap if you want it. The Canadian postal code is the one I would fix before merging, because it is not a deliberate simplification, it is the same rule applied in one position and not the others.

Worth adding tests for whichever you tighten. The current cases (`K1A 0A6`, `SW1A 2AA`, and a bare `"12"`) confirm the happy path and reject something obviously malformed, but nothing pins the character sets, which is where these patterns will drift.

src/outlines/types/locale/uk.py
```diff
@@ -0,0 +1,6 @@
+"""Locale-specific regex patterns for the United Kingdom."""
+
+from outlines.types.dsl import Regex
+
+phone_number = Regex(r"(\+44\s?|0)((7\d{3}|\d{4})[\s-]?\d{6}|(20\d[\s-]?\d{4}[\s-]?\d{4})|\d{3}[\s-]?\d{3}[\s-]?\d{4})")
```
RobinPicard, line 5: I think this branch is wrong `(20\d[\s-]?\d{4}[\s-]?\d{4})`. It matches a total of 11 characters without the leading 0 while numbers should be only 10

RobinPicard (changes requested):
Thanks for opening a PR. I just think there's an issue in a regex

ErenAta16 (commented):
@RobinPicard is right about that branch, and the consequence is a bit worse than the character count suggests: as written it rejects the standard way of writing a London number.

Measured against the current head:

```
020 7946 0958      current=False   with the fix=True
+44 20 7946 0958   current=False   with the fix=True
02079460958        current=True    with the fix=True
020 79460 9581     current=False   with the fix=False    <- 11-digit NSN, rejected either way
```

The fix is dropping the stray `\d`:

```
(20\d[\s-]?\d{4}[\s-]?\d{4})    ->    (20[\s-]?\d{4}[\s-]?\d{4})
```

**The reason the unseparated form still passes is that it never reaches this branch.** `\d{3}[\s-]?\d{3}[\s-]?\d{4}` already accepts `02079460958` by reading it as 207-946-0958:

```
02079460958        generic branch alone = True
020 7946 0958      generic branch alone = False
```

So the only thing the London branch contributes is the 2-4-4 grouping, which is precisely what the extra digit breaks. In its current form it contributes nothing valid at all: with separators it matches only 11-digit numbers that should be rejected, and the 10-digit numbers it exists to accept fall through to a branch that cannot take the spacing.

**Worth adding a fixture, because the tests here cannot see any of this.** Both phone cases in the diff are mobiles:

```python
(types.locale.uk.phone_number, "+44 7911 123456", True),
(types.locale.uk.phone_number, "07911 123456", True),
```

Those resolve through `(7\d{3}|\d{4})[\s-]?\d{6}`, so the London branch is untested in both directions and would stay untested after the fix. `("020 7946 0958", True)` fails today and passes once the `\d` goes. Ofcom reserves 020 7946 0xxx for drama and examples, so it is a safe number to hard-code in a test suite.

For what it is worth, the Canadian patterns check out against Canada Post's letter rules: the first-position set excludes D, F, I, O, Q, U, W and Z, and the interior set excludes D, F, I, O, Q and U while still allowing W and Z. That asymmetry is easy to get wrong and it is right here.

ErenAta16 (commented):
`e060ed45` fixes it. Dropping the `\d` from the London branch is exactly the
change, and it does what @RobinPicard's comment and my 14 August measurement
asked for. Re-ran the same cases against the new pattern:

```
input                current   e060ed45
020 7946 0958        False     True
+44 20 7946 0958     False     True
+4420 7946 0958      False     True
02079460958          True      True
07700 900123         True      True
0161 496 0123        True      True
020 7946 09581       False     False     one digit too many
020 7946 095         False     False     one digit too few
```

The three rows that flipped are the ones that mattered, and the over- and
under-length London numbers are still refused, so the bound tightened in the
right direction rather than the branch simply becoming permissive. Mobile and
other geographic numbers are untouched.

Worth recording why this was easy to miss: `02079460958` matched before the fix
too, because the generic `\d{3}[\s-]?\d{3}[\s-]?\d{4}` branch happens to accept
the unspaced form. So the defect only showed when a London number was written the
way people actually write it, with the spaces, and a test using the compact form
would have passed throughout.

One housekeeping thing: GitHub reports the branch as `behind`. A merge from `main`
would clear that before anyone looks at it again.

Goutham-Annem:
Branch is now synced with `main` (b530052). @ErenAta16's Sep 11 analysis confirmed the fix resolves all three London number formats. Thank you both for the thorough review.

ErenAta16:
Ran all four new types on the synced branch. Seventeen cases, everything as expected, including the three London formats:

```
uk.phone_number   020 7946 0958     match     07911 123456   match
                  +44 20 7946 0958  match     123            rejected
                  02079460958       match
uk.postcode       SW1A 1AA          match     XYZ            rejected
                  EC1A 1BB          match
                  W1A 0AX           match
ca.postal_code    K1A 0B1           match     D1A 0B1        rejected
                  M5V3L9            match     K1A 0B         rejected
ca.phone_number   (416) 555-1234    match     555-1234       rejected
                  +1 416 555 1234   match
                  4165551234        match
```

`D1A 0B1` rejected is the one worth pointing at, since `D` is one of the letters Canada Post never issues as a first character and a naive `[A-Z]` would have let it through.

One thing I want to put on the record before a reviewer raises it as an inconsistency: `uk.postcode` next to `ca.postal_code` looks like a naming slip, and it is not. The only locale on `main` is `us.py`, which exposes `zip_code`, so the established convention here is that each locale uses its own domestic term rather than a normalised one. `postcode` for the UK is the right call under that precedent, not a deviation from it.

Nothing further from me.

PR #1955, opened 2026-07-24
feat(types): add Australian locale types (phone, postcode)

### LINKED ISSUE #1303: Add more custom types

Goutham-Annem:
Closes #1303

## What

Adds an `au` (Australia) locale module under `outlines.types.locale`, providing:

- `phone_number` — matches Australian mobile numbers (`04XX XXX XXX`, `+61 4XX XXX XXX`) and landline numbers (`0X XXXX XXXX`, `+61 X XXXX XXXX`) with optional separators (space, dot, hyphen).
- `postcode` — matches valid 4-digit Australian postcodes (`0[289]XX` for ACT/NT, `1XXX`–`9XXX` for all other states and territories).

## Pattern details

| Type | Example valid inputs |
|------|---------------------|
| `au.phone_number` | `+61 412 345 678`, `0412 345 678`, `+61 2 9876 5432`, `02 9876 5432` |
| `au.postcode` | `2000` (Sydney), `3000` (Melbourne), `0800` (Darwin NT) |

## Files changed

- `src/outlines/types/locale/au.py` — new locale module
- `src/outlines/types/locale/__init__.py` — export `au`
- `tests/types/test_custom_types.py` — parametrized test cases for both types


ErenAta16 (commented):
Ran both patterns at `ebe2866` against real allocations rather than just the cases in the tests, and they hold up on everything I threw at them.

Postcode, `(0[289]\d{2}|[1-9]\d{3})`:

```
2000  accept   Sydney            0000  reject   not allocated
0200  accept   ACT               0100  reject   not allocated
0800  accept   Darwin, NT        0300  reject   not allocated
0900  accept   NT                0700  reject   not allocated
3000  accept   Melbourne         999   reject   three digits
```

Splitting the leading-zero range into `0[289]` rather than allowing a blanket `0\d{3}` is the detail I was checking for. The only allocated ranges starting with zero are 02xx for the ACT and 08xx/09xx for the Northern Territory, and this gets that exactly right while rejecting 00xx, 01xx and 03xx through 07xx, which are not issued. That is more careful than most postcode regexes bother to be.

Phone, `(\+61\s?|0)([2378][\s.-]?\d{4}[\s.-]?\d{4}|4\d{2}[\s.-]?\d{3}[\s.-]?\d{3})`:

```
+61 2 1234 5678   accept    02 1234 5678   accept
+61 412 345 678   accept    0412 345 678   accept
05 1234 5678      reject    (5 is not an area code)
```

The part worth calling out is that `(\+61\s?|0)` is an alternation rather than an optional prefix followed by an optional zero. That is the correct model of how Australian numbers work, the trunk `0` is dropped in international form, and it means the pattern rejects the mixed spelling:

```
+61 0412 345 678   reject
+610412345678      reject
```

A `(\+61)?0?` construction would have accepted both, and that is the single most common way this gets written wrong. Good that it does not.

Nothing blocking from me. The one thing I would suggest is adding a couple of the rejection cases above to the test table, specifically an unallocated leading-zero postcode like `0100` and the `+61 0412...` trunk-zero form. Both currently pass by construction rather than by assertion, and they are exactly the two properties a later "simplification" of these patterns would quietly break.

PR #1956, opened 2026-07-24
feat(types): add French locale types (phone, postal_code)

### LINKED ISSUE #1303: Add more custom types

Goutham-Annem:
Closes #1303

## What

Adds an `fr` (France) locale module under `outlines.types.locale`, providing:

- `phone_number` — matches French 10-digit phone numbers including mobile (`06`/`07` prefixes), landlines (`01`–`05`), and service numbers (`08`/`09`), with an optional `+33` international prefix and optional space separators between digit pairs.
- `postal_code` — matches valid 5-digit metropolitan France postal codes (`01000`–`95999`), excluding overseas territories.

## Pattern details

| Type | Example valid inputs |
|------|---------------------|
| `fr.phone_number` | `0612345678`, `+33 6 12 34 56 78`, `0123456789`, `+33 1 23 45 67 89` |
| `fr.postal_code` | `75001` (Paris), `01000` (Ain), `59000` (Lille) |

## Files changed

- `src/outlines/types/locale/fr.py` — new locale module
- `src/outlines/types/locale/__init__.py` — export `fr`
- `tests/types/test_custom_types.py` — parametrized test cases for both types


ErenAta16 (commented):
Ran both patterns at `e346770`. The phone number one is correct on everything I tried:

```
+33 6 12 34 56 78   accept      0612345678    accept
01 42 68 53 00      accept      0012345678    reject
```

`[1-9]` after the prefix is the right guard, since no French subscriber number starts with a second zero, and `(\s?\d{2}){4}` handles both the grouped and unspaced spellings without also accepting a half-spaced mess.

The postal code has a real gap though. `0[1-9]\d{3}|[1-8]\d{4}|9[0-5]\d{3}` stops at 95999, which covers metropolitan France and nothing else:

```
75001  accept   Paris 1er
01000  accept   Bourg-en-Bresse
95999  accept   top of the metropolitan range
97100  reject   Basse-Terre, Guadeloupe
97400  reject   Saint-Denis, La Reunion
98000  reject   Monaco
98800  reject   Noumea, New Caledonia
```

97xxx and 98xxx are not edge cases or legacy allocations, they are the live ranges for the overseas departments and collectivities: Guadeloupe, Martinique, Guyane, La Reunion and Mayotte in 97xxx, and Monaco, French Polynesia and New Caledonia in 98xxx. La Poste operates them and they appear on ordinary French addresses. Anything generated under this type simply cannot produce one.

`9[0-8]\d{3}` in place of `9[0-5]\d{3}` covers it and keeps 99xxx out, which is the range you do want to exclude. That does also let in 96xxx, which is unallocated, so if you would rather stay exact then `9[0-5]\d{3}|9[78]\d{3}` is the precise version. Either is fine, I would take the simpler one unless the module is aiming to be an allocation table.

Worth deciding deliberately rather than by omission, because the same question applies to the `us` module this is modelled on and to the other locale PRs in flight. If the intent is "metropolitan France only" then that is a legitimate scope, it just needs to be in the docstring so someone in Fort-de-France does not file a bug against it later.

One thing I checked and it is fine, so ignore this as a concern. `postal_code` is the only pattern in these locale modules written without an enclosing group, unlike `au.postcode` which wraps its alternation. That would matter if the bare alternation could leak into a surrounding expression, but `to_regex` parenthesises each term on the way out:

```
to_regex(String("code: ") + locale.fr.postal_code)
'code:\ (0[1-9]\d{3}|[1-8]\d{4}|9[0-5]\d{3})'

"code: 75001"  accept      "code: 7500"  reject      "75001"  reject
```

so concatenation binds correctly. Adding the parentheses anyway would make it consistent with the neighbouring modules, but it is cosmetic rather than a fix.

RobinPicard (changes requested):
I think we should also add support for oversea territories for postal codes to be exhaustive

ErenAta16 (commented):
@RobinPicard I went to check which codes were missing before suggesting branches, and I think the exhaustiveness you are after is already there.

`9[0-8]\d{3}` spans 90000 to 98999, which contains every overseas department and collectivity:

```
Guadeloupe                 97110   covered
Martinique                 97200   covered
Guyane                     97300   covered
La Reunion                 97400   covered
Saint-Pierre-et-Miquelon   97500   covered
Mayotte                    97600   covered
Saint-Barthelemy           97133   covered
Saint-Martin               97150   covered
Wallis-et-Futuna           98600   covered
Polynesie francaise        98700   covered
Nouvelle-Caledonie         98800   covered
Monaco                     98000   covered
```

All twelve match the pattern as it stands, so an explicit DOM-TOM alternative would overlap the third branch rather than extend it.

What the three branches span in total:

```
0[1-9]\d{3}   01000-09999
[1-8]\d{4}    10000-89999
9[0-8]\d{3}   90000-98999
```

If tightening these is worthwhile, the discrepancies run the other way. 96xxx falls inside `9[0-8]\d{3}` but no department uses it, and 00100 through 00480, the armed forces sectors, are excluded because `0[1-9]` requires a nonzero second digit. Both are judgement calls about how strict a generation constraint should be, and neither is about overseas territories.

One thing that would be worth adding either way is a test fixture from the 97/98 range. Every postal fixture in the diff is metropolitan (`75001`, `01000`, `59000`), so nothing currently pins the overseas behaviour that this discussion is about, and a later tightening of the third branch could drop it without a test noticing.

I built that list from the department numbering rather than a postal export, so if you have a specific code in mind that you know fails, I would rather see it than assume the pattern is complete.

Goutham-Annem:
Added test fixtures for the 97/98 range in eebd1ad: `97100` (Basse-Terre, Guadeloupe) and `98000` (Monaco) both pass, and `99000` is rejected as expected. This pins the behaviour @ErenAta16 described on Aug 14 — `9[0-8]\d{3}` already covers all twelve overseas departments and collectivities.

ErenAta16:
Ran the branch rather than reading the fixtures, and the 97/98 behaviour holds.

The type composes to `(0[1-9]\d{3}|[1-8]\d{4}|9[0-8]\d{3})`. I specifically wanted to see that grouping before agreeing the fixtures were enough: a bare three-way alternation is the kind of thing that binds loosely once the type is nested inside a container, but `to_regex` wraps it, so the same pattern holds standalone and inside a `List`/`Dict`.

Measured on `eebd1ad`:

```
75001   match      99000    rejected
01000   match      00000    rejected
59000   match      123456   rejected
97100   match      9800     rejected
98000   match
```

The one that earns its place is `99000 -> False`. The five positives would all still pass if the last alternative were widened to `9\d{4}`, so that row is the only one that pins the upper bound where you actually put it.

Nothing further from me.

PR #1963, opened 2026-07-27
fix: resolve $ref and combinator keywords in JSON Schema to Python conversion

DerMayer1:
### What

`schema_type_to_python` handles `enum`, `const`, `type`, arrays and objects. Everything else falls through to `return Any`. `$ref`, `anyOf`, `oneOf` and `allOf` are all unhandled, so a schema using them loses its constraints. For `$ref` that loss is not silent: it makes the converted model unusable.

Pydantic emits `$ref` for a nested model, so a round trip drops the nesting. This is not a
quiet widening of the type: the converted model is then unusable, because the regex backend
refuses a property with no `type`.

```python
to_regex(JsonSchema(json_schema_dict_to_pydantic(Person.model_json_schema(), "Person")))
ValueError: Unsupported JSON Schema structure {"title":"Address"}
```

So on `main` the converter cannot round-trip any pydantic model that contains another model.

```python
class Address(BaseModel):
    street: str
    city: str

class Person(BaseModel):
    name: str
    address: Address

json_schema_dict_to_pydantic(Person.model_json_schema(), "Person")
# before: address -> typing.Any
# after:  address -> Address(street: str, city: str)
```

The three converters also read `properties` directly off the document. Pydantic emits a *top-level* `$ref` for a self-referential model — `{"$defs": {...}, "$ref": "#/$defs/Node"}` — so `properties` is empty and the generated model has no fields at all:

```python
class Node(BaseModel):
    value: int
    children: Optional[List["Node"]] = None

json_schema_dict_to_pydantic(Node.model_json_schema(), "Node")
# before: {}   (no fields)
# after:  value: int, children: Optional[List[Any]]
```

### Why it's user-visible

Gemini is the one backend that converts a JSON schema into a Python type and ships it as `response_schema` — `models/gemini.py:181`, `JsonSchema.convert_to(output_type, ["dataclass", "typeddict", "pydantic"])`. So passing a schema with a nested object to Gemini sends a constraint that doesn't contain the nested structure, with no warning.

Other backends are unaffected: OpenAI, Ollama and LMStudio convert to `dict`, dottxt to `str`, and the local backends compile the raw schema.

### Scope

**In.** Local `$ref` (`#`, `#/$defs/…`, `#/definitions/…`), `anyOf`/`oneOf` → `Union`, and a one-element `allOf` → its subschema (the form Pydantic emits for a `$ref` with sibling metadata).

**Out.** Remote refs, dangling pointers and multi-element `allOf` stay `Any` rather than raising, so an unsupported schema still degrades instead of breaking. Merging two or more `allOf` subschemas means reconciling conflicting keywords and competing `required` sets; that's a separate change.

Reference cycles resolve to `Optional[Any]`. A recursive schema has no finite Python type
here, and the alternative is recursing until the stack ends. It is `Optional[Any]` rather
than a bare `Any` so the back edge round-trips to `{"anyOf": [{}, {"type": "null"}]}`, which
the regex backend accepts; a bare `Any` serializes to a property with no `type` and raises
the same `ValueError` as above when the recursive property is required. Null is also the
only way an instance of a required self-reference can terminate.

One incidental change: `items` given as a list (the draft-7 tuple form) previously raised `AttributeError` from `schema.get` on a list. It is now `List[Any]`. Happy to drop that if you'd rather keep the PR to the `$ref` path.

### Why this is a fix rather than a feature

`tests/types/test_json_schema_utils.py` already covers nested objects and asserts full recursion into them (`test_json_schema_dict_to_pydantic_nested_object` and the typeddict/dataclass equivalents). `$ref` is the same intent expressed through `$defs` rather than inline, so the existing tests describe behaviour that `$ref` schemas were silently not getting.

### Tests

17 added, covering local `$ref`, the draft-07 `definitions` keyword, unresolvable refs, root-level `$ref`, recursive `$ref`, `anyOf`/`oneOf`/`allOf`, the typeddict and dataclass paths, the end-to-end Pydantic round trip, and `to_regex` over
converted recursive and nested models rather than their annotations alone.

`tests/types/`: 302 passed, 2 failed. Both failures are `test_dsl.py::test_dsl_cfg_from_file` and `test_dsl_json_schema_from_file`, which fail identically on a clean checkout of `main` on Windows (`NamedTemporaryFile` cannot be reopened while held). Unrelated to this change.

ruff 0.9.1 with `--config=pyproject.toml` and mypy 1.14.1 with `--allow-redefinition` are both clean on the changed files.

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -115,6 +115,15 @@ def schema_type_to_python(
     Any
         The Python type
 
+    Notes
+    -----
+    The back edge of a recursive schema is intentionally left unconstrained,
+    as ``Optional[Any]``. It is the designed stopping point rather than a
+    missed conversion: a finite regex cannot express an unbounded tree, and
+    null is the only way an instance of a required self-reference can
+    terminate. Narrowing it to a bare ``Any`` would produce a property with no
+    ``type``, which the regex backend rejects.
+
     """
     # ``$defs`` live on the top-level schema, so the root has to travel with
     # the recursion; without it a nested ``$ref`` has nothing to resolve against.
@@ -123,11 +132,20 @@ def schema_type_to_python(
 
     if "$ref" in schema:
         ref = schema["$ref"]
-        # A recursive schema (a model referencing itself, directly or through
-        # a cycle) has no finite Python type here, so the cycle is broken with
-        # ``Any`` rather than recursing until the stack runs out.
+        # A recursive schema (a model referencing itself, directly or through a
+        # cycle) has no finite Python type here, so the back edge is left
+        # deliberately unconstrained rather than recursing until the stack runs
+        # out. This is the designed stopping point, not a missed conversion: a
+        # finite regex cannot express an unbounded tree.
+        #
+        # ``Optional[Any]`` rather than ``Any``, because the back edge has to
+        # round-trip to a schema the regex backend accepts. A bare ``Any``
+        # serializes to a property with no ``type`` (``{"title": "Next"}``),
+        # which ``to_regex`` rejects outright; ``Optional[Any]`` serializes to
+        # ``{"anyOf": [{}, {"type": "null"}]}``, which it accepts. Null is also
+        # the only way an instance of a *required* self-reference can terminate.
         if ref in _seen:
-            return Any
+            return Optional[Any]
         resolved = _resolve_ref(ref, root)
         if resolved is None:
             return Any
@@ -135,6 +153,11 @@ def schema_type_to_python(
             resolved, caller_target_type, root=root, _seen=_seen | {ref}
         )
 
+    # A member that is not a dict is not a subschema. Every combinator drops
+    # those and falls back to ``Any`` once nothing valid is left, rather than
+    # raising: an unusable member should not make the whole schema unusable.
+    # This is separate from a well-formed ``allOf`` of several object
+    # subschemas, which is left as ``Any`` for the reason given below.
     for keyword in ("anyOf", "oneOf"):
         if keyword in schema:
             members = tuple(
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -12,9 +12,88 @@ else:  # pragma: no cover
     from typing_extensions import _TypedDictMeta, NotRequired, TypedDict  # type: ignore
 
 
+def _resolve_ref(ref: str, root: dict) -> Optional[dict]:
+    """Resolve a local JSON Pointer against the root schema.
+
+    Only local references are resolvable: ``#``, ``#/$defs/Name``,
+    ``#/definitions/Name``. Remote references (a URI, or a pointer into
+    another document) return ``None`` and the caller falls back to ``Any``.
+
+    Parameters
+    ----------
+    ref: str
+        The value of a ``$ref`` keyword
+    root: dict
+        The schema the pointer is resolved against
+
+    Returns
+    -------
+    Optional[dict]
+        The referenced subschema, or None if it is not a resolvable local ref
+
+    """
+    if not ref.startswith("#"):
+        return None
+
+    pointer = ref[1:].lstrip("/")
+    if not pointer:
+        return root
+
+    target: Any = root
+    for token in pointer.split("/"):
+        # JSON Pointer escapes, RFC 6901: ~1 is '/', ~0 is '~'.
+        token = token.replace("~1", "/").replace("~0", "~")
+        if not isinstance(target, dict) or token not in target:
+            return None
+        target = target[token]
+
+    return target if isinstance(target, dict) else None
+
+
+def _deref_root(
+    schema: dict, root: dict, _seen: frozenset
+) -> tuple[dict, frozenset]:
+    """Follow top-level ``$ref`` keywords to the subschema holding the properties.
+
+    Pydantic emits ``{"$defs": {...}, "$ref": "#/$defs/Model"}`` for a
+    self-referential model, so the object being converted is one indirection
+    away from the document root. Reading ``properties`` off the document
+    directly yields nothing.
+
+    Parameters
+    ----------
+    schema: dict
+        The schema to dereference
+    root: dict
+        The schema the pointer is resolved against
+    _seen: frozenset
+        The ``$ref`` pointers already followed
+
+    Returns
+    -------
+    tuple[dict, frozenset]
+        The dereferenced schema and the updated set of followed pointers
+
+    """
+    while "$ref" in schema:
+        ref = schema["$ref"]
+        if ref in _seen:
+            break
+        resolved = _resolve_ref(ref, root)
+        if resolved is None:
+            break
+        schema = resolved
+        _seen = _seen | {ref}
+
+    return schema, _seen
+
+
 def schema_type_to_python(
     schema: dict,
-    caller_target_type: Literal["pydantic", "typeddict", "dataclass"]
+    caller_target_type: Literal["pydantic", "typeddict", "dataclass"],
+    *,
+    root: Optional[dict] = None,
+    _seen: frozenset = frozenset(),
 ) -> Any:
     """Get a Python type from a JSON Schema dict.
 
@@ -24,6 +103,12 @@ def schema_type_to_python(
         The JSON Schema dict to convert to a Python type
     caller_target_type: Literal["pydantic", "typeddict", "dataclass"]
         The type of the caller
+    root: Optional[dict]
+        The top-level schema, against which ``$ref`` pointers are resolved.
+        Defaults to ``schema`` itself, so a top-level call needs no argument.
+    _seen: frozenset
+        The ``$ref`` pointers currently being resolved, used to stop a
+        self-referential schema from recursing forever.
 
     Returns
     -------
@@ -31,6 +116,48 @@ def schema_type_to_python(
         The Python type
 
     """
+    # ``$defs`` live on the top-level schema, so the root has to travel with
+    # the recursion; without it a nested ``$ref`` has nothing to resolve against.
+    if root is None:
+        root = schema
+
+    if "$ref" in schema:
+        ref = schema["$ref"]
+        # A recursive schema (a model referencing itself, directly or through
+        # a cycle) has no finite Python type here, so the cycle is broken with
+        # ``Any`` rather than recursing until the stack runs out.
+        if ref in _seen:
+            return Any
+        resolved = _resolve_ref(ref, root)
+        if resolved is None:
+            return Any
+        return schema_type_to_python(
+            resolved, caller_target_type, root=root, _seen=_seen | {ref}
+        )
+
+    for keyword in ("anyOf", "oneOf"):
+        if keyword in schema:
+            members = tuple(
+                schema_type_to_python(
+                    subschema, caller_target_type, root=root, _seen=_seen
+                )
+                for subschema in schema[keyword]
+                if isinstance(subschema, dict)
+            )
+            return Union[members] if members else Any  # type: ignore
+
+    if "allOf" in schema:
+        subschemas = [s for s in schema["allOf"] if isinstance(s, dict)]
+        # A single-element ``allOf`` is just an alias for its one subschema —
+        # the form Pydantic emits for a ``$ref`` carrying sibling metadata.
+        # Merging two or more subschemas is a different problem (conflicting
+        # keywords, competing ``required`` sets) and is left as ``Any``.
+        if len(subschemas) == 1:
+            return schema_type_to_python(
+                subschemas[0], caller_target_type, root=root, _seen=_seen
+            )
+        return Any
+
     if "enum" in schema:
         values = schema["enum"]
         return Literal[tuple(values)]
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Ran this against a clone of `main` at `be2cd15` and then against `d986758`, using the `Person`/`Address` case from the description plus a set of schemas I picked to probe the edges. It does what it says, and the headline case holds up end to end:

```
Person.address    main: typing.Any        this PR: Address(street: str, city: str)
```

Generation follows through, which is the part I cared about most:

```
{"name": "a", "address": {"street": "s", "city": "c"}}   -> True
{"name": "a", "address": "not an object"}                -> False
{"name": "a", "address": 5}                              -> False
```

One thing you should put in the description, because it makes the case stronger than what is written there now. On `main` this is not a silent loss of constraints for nested models, it is a hard failure. `Any` round-trips to a property with no `type`, and the regex builder refuses it:

```
to_regex(JsonSchema(json_schema_dict_to_pydantic(Person.model_json_schema(), "Person")))
ValueError: Unsupported JSON Schema structure {"title":"Address"}
```

So today the converter cannot round-trip any pydantic model that contains another model, it raises rather than generating something loose. "Silently loses its constraints" undersells it and might be why this has sat unreviewed.

Combinators, all `Any` on `main`:

```
anyOf [str, int]      Union[str, int]
oneOf [str, null]     Optional[str]
anyOf [int, null]     Optional[int]
allOf single          int
allOf multi           Any
enum / array / plain  unchanged
```

The recursion question was the one I expected to break, and it does not. Both a self-referential schema and a mutually recursive pair terminate:

```
Node   value -> int,  child -> Optional[Any]
A      b     -> Optional[AnonymousPydanticModel]
```

No `RecursionError`, and generation still works on the converted recursive model because the back edge lands as `Optional[Any]`, which round-trips to `{"anyOf": [{}, {"type": "null"}]}` and the regex builder accepts that shape even though it rejects a bare untyped property. That is a nice accident of the two representations and I would not want it to stay accidental. There is no test covering a recursive schema, so if the cycle guard is ever tightened to return something other than `Optional[Any]`, or loosened to plain `Any`, this goes straight back to the `ValueError` above and nothing in the suite catches it. Worth one test that converts a self-referential schema and calls `to_regex` on the result, not just on the annotations.

It is also worth a line in the docstring saying the back edge is intentionally unconstrained. It is the correct call, a finite regex cannot express an unbounded tree, but a reader hitting `child -> Optional[Any]` will otherwise read it as the bug this PR set out to fix rather than the designed stopping point.

Cross-PR note, since these will collide. #1949 changes `field_definitions[property] = (Optional[typ], None)` to `(typ, None)`, and that exact line sits inside one of this PR's hunks as context, so merge order matters textually. Semantically they fit together rather than fight: this PR is what makes an explicitly nullable schema produce `Optional`, and #1949 stops a merely non-required field from producing it. Landing #1963 first and #1949 second reads cleanest, because after this PR the `Optional` that #1949 removes is genuinely redundant for the `anyOf`/`oneOf` null cases rather than being the only thing providing them. Flagging so they are not reviewed as independent.

One smaller thing, take it or leave it. Malformed subschemas are handled two different ways:

```
{"anyOf": [{"type": "integer"}, "junk"]}    ->  int     (non-dict member dropped)
{"allOf": [{"type": "integer"}, {"type": "string"}]}  ->  Any   (bails out)
```

Dropping the junk member is defensible, and so is bailing, but right now which one you get depends on the keyword. If that is deliberate a short comment saying so would help, since a schema with a typo'd member currently narrows to `int` without complaint.

Otherwise this is a solid change and the test additions cover the parts that matter.

DerMayer1:
Thanks for actually running it, and for the `to_regex` point. You were right that the recursion held up by accident, and chasing that turned up a case that is broken today rather than only after some future change.

The back edge is safe only when the recursive property is optional, because then `json_schema_dict_to_pydantic` wraps it and you get `{"anyOf": [{}, {"type": "null"}]}`. Make it required and there is no wrapper:

```
{"$defs": {"Node": {"properties": {"value": {...}, "next": {"$ref": "#/$defs/Node"}},
                    "required": ["value", "next"]}}, "$ref": "#/$defs/Node"}

next -> typing.Any  ->  {"title": "Next"}
ValueError: Unsupported JSON Schema structure {"title":"Next"}
```

Same for a mutually recursive pair where both references are required. So the exact error this PR removes for nested models was still reachable through recursion.

The cycle now returns `Optional[Any]` directly instead of relying on the caller to wrap it. Null is also the only way an instance of a required self-reference can terminate, so it is the honest type rather than a patch. Four shapes now round-trip: required back edge, optional back edge, mutual pair, and list-based recursion.

Tests call `to_regex` on the converted model rather than only checking annotations, which is the part that would have caught this. Reverting the guard to a bare `Any` fails three of them. Added the `Notes` section saying the back edge is deliberately unconstrained and why narrowing it breaks the regex backend.

On the description, you are right and I have updated it. Calling it a silent loss of constraints was wrong: `to_regex` refuses the converted model outright, so today the converter cannot round-trip any pydantic model containing another model.

On the malformed members, I looked into it and the behaviour is already consistent, though my examples in the code did not say so. All three combinators drop a non-dict member and fall back to `Any` when none are left:

```
anyOf [{"type": "integer"}, "junk"]   ->  int
oneOf [{"type": "integer"}, "junk"]   ->  int
allOf [{"type": "integer"}, "junk"]   ->  int
anyOf ["junk"]                        ->  Any
allOf ["junk"]                        ->  Any
```

The `Any` you saw from `allOf` was the multi-member case, `allOf [{"type": "integer"}, {"type": "string"}]`, which is a well-formed intersection rather than a malformed member. Merging those means reconciling conflicting keywords and competing `required` sets, so it stays `Any`. Added a comment stating both rules, and a parametrized test over the three keywords so the malformed-member behaviour is pinned rather than incidental.

Agreed on #1949 and the ordering. This PR is what makes an explicitly nullable schema produce `Optional`, so landing it first leaves #1949 removing an `Optional` that is genuinely redundant. Happy to rebase on whichever lands first.

`tests/types`: 302 passed, 2 failed, both `test_dsl.py` file-handling tests that fail the same way on a clean checkout of `main` on Windows. ruff 0.9.1 and mypy 1.14.1 clean.

ErenAta16 (commented):
Reproduced your required-versus-optional split, and the asymmetry comes from two adjacent lines in `json_schema_dict_to_pydantic`:

```python
if property not in required:
    field_definitions[property] = (Optional[typ], None)
else:
    field_definitions[property] = (typ, ...)
```

With `typ` degraded to `Any` by the unresolved back edge, those two branches serialize very differently:

```
required  (typ, ...)              -> {"title": "Next"}
optional  (Optional[typ], None)   -> {"anyOf": [{}, {"type": "null"}], "default": null, "title": "Next"}
```

and that is the whole story:

```
required   type=False  anyOf=False   -> no structural keyword to dispatch on
optional   type=False  anyOf=True    -> dispatchable
```

The optional spelling smuggles in an `anyOf` that the container path is willing to walk, so the cycle appeared to work while carrying no type information at all. The required spelling has nothing to dispatch on and surfaces as the `Unsupported JSON Schema structure` you hit. So the recursion was not merely fragile against a future change, it was already broken for the more common way to write a linked structure, and the optional case was masking it.

**The consequence for the tests is the part I would make sure lands.** A recursive fixture written with an optional `next` passes on `main` today and would keep passing under a change that reintroduces this, because the `anyOf` wrapper does the dispatching rather than the fix. Any regression test for the cycle needs the **required** spelling, and ideally both, with a comment saying the optional one is a compatibility case rather than a guard. Otherwise the suite looks like it covers recursion and covers only the accident.

The mutually-recursive pair with both references required is the same shape one step out, and worth including for the reason you give: it is the case where neither side has a wrapper to hide behind.

Thanks for chasing it rather than taking the "it happens to work" answer. I would not have found the required case from where I was looking.

percymcn:
Reproduced this and ran the patch against a battery before commenting. It works, and I think the case for it is **stronger than the PR body claims** — there is a second, silent failure path that the round-trip example doesn't show.

**Verified on 1.3.3 + the patch** (9 shapes through `json_schema_dict_to_dataclass`):

| shape | `main` | with this PR |
|---|---|---|
| pydantic nested model (`$ref`) | `Any` | resolved dataclass |
| `anyOf` of scalars | `Any` | `str \| int` |
| `anyOf` object + null | `Any` | `dataclass \| None` |
| single-member `allOf` | `Any` | `str` |
| recursive `$ref` | `Any` | resolved, terminates |
| `$ref` to an RFC 6901 escaped name (`#/$defs/a~1b`) | `Any` | resolved |
| dangling `$ref` | `Any` | `Any` (fails soft — correct) |
| plain typed property (control) | `str` | `str` |
| 2-member `allOf` | `Any` | `Any` — deliberate, and pinned by your own assertion |

The escaped-name row is worth calling out: `~1` is a spelling that several constrained-decoding libraries get wrong, so handling it here is not incidental.

**The part I'd add to the motivation.** The PR says the `$ref` loss "is not silent … the converted model is then unusable, because the regex backend refuses a property with no `type`". That's true for the `to_regex` round trip you show, but that path is one a user assembles by hand. In shipped code there is exactly one caller that reaches this converter at all:

```
models/gemini.py:197   JsonSchema.convert_to(t, ["dataclass","typeddict","pydantic"])   <- converts
models/openai.py:175   ... ["dict"]      models/ollama.py:141   ... ["dict"]
models/lmstudio.py:155 ... ["dict"]      models/dottxt.py:73    ... ["str"]
anthropic / vllm / transformers / llamacpp: no convert_to at all
```

Every other backend forwards the schema verbatim. So the converter's real audience is Gemini — and on that path the failure is **not** loud. On `main`, an ordinary nested pydantic model reaches google-genai as:

```json
"inner": { "title": "Inner" }
```

a property with no `type`, and the live `v1beta` endpoint **accepts** it (a bogus key in the same slot is rejected with `Unknown name … Cannot find field`, so the payload validator is discriminating there). The request succeeds and the field is simply unconstrained — the model can return anything, or nothing, for a required nested object. That's the failure mode a constrained-decoding library exists to prevent, and it is invisible.

**Your patch fixes that end to end.** Same model, same call, with the patch applied:

```json
"inner": { "type": "OBJECT", "title": "Inner",
           "properties": { "kind": {"type":"STRING"}, "score": {"type":"INTEGER"} },
           "required": ["kind","score"] }
```

So this isn't only a round-trip nicety — on `main`, `outlines` + Gemini + any pydantic model containing another model silently ships an unconstrained field. Worth prioritising the review accordingly.

One test-design note, since none of the added tests exercises `models.gemini`: a regression test asserting the *converted Python type* catches everything above, but the silent-acceptance property is a different assertion, and the cheap version needs no API key — build the dataclass, run it through `google.genai._transformers.t_schema`, and assert the emitted property has a `type`. On `main` that assertion fails; with the patch it passes.

ErenAta16 (commented):
Went back through the combinator handling at `0d43be2a` and found one case where the non-dict filter goes the wrong way. Everything below is measured on that head.

The guard is a real improvement over #1951, which raises `TypeError` on all of these inputs. But `true` and `false` are opposite schemas, and dropping both alike is only right for one of them:

```
                       this branch    correct
anyOf [int, true]      int            Any      true accepts everything, so the union does too
anyOf [int, false]     int            int      ok
allOf [int, true]      int            int      ok
```

The sharpest way to see it is that `{}` and `true` are the same schema, both accept every instance, and this branch answers differently depending on which spelling the document uses:

```
                              dict spelling     bool spelling
anyOf  int + always-valid     Union[int, Any]   int        diverges
oneOf  str + always-valid     Union[str, Any]   str        diverges
allOf  int + always-valid     Any               int        diverges
```

The rule underneath is that in a union the always-valid schema is absorbing, so the answer has to be `Any`, while in an intersection it is the identity, so the answer is the other member. A single filter applied to all three keywords gets `allOf` right and `anyOf`/`oneOf` wrong. `allOf` diverges the other way for the same reason, and there the boolean path is the better answer: the dict spelling counts `{}` as a second subschema and bails to `Any`.

**Why I would fix this rather than note it.** Every other gap in this area widens to `Any`. This one narrows, and narrowing is the direction that hurts a constrained-decoding library, because it makes legal outputs unreachable instead of merely permitting extra ones:

```python
{"type": "object", "properties": {"x": {"anyOf": [{"type": "integer"}, True]}}, "required": ["x"]}

emitted x schema:      {"title": "X", "type": "integer"}
x=7                    accepted
x='legal-under-true'   REJECTED     (legal under the source schema)
```

A schema that permits any value for `x` generates as integer-only, and nothing raises anywhere along the way.

Your comment at the filter says the drop is there so "an unusable member should not make the whole schema unusable", which is exactly right for `false` and for `allOf`. It just needs the union case to treat `True` separately, returning `Any` as soon as a member is `True` rather than filtering it out. That is about two lines and it makes the `{}` and `true` spellings agree.

Boolean members are rare in hand-written schemas and I would not hold the PR for this. But it is the one place in the diff where the failure direction flips, and right now the comment reads as deliberate, so it would otherwise survive review as intended behaviour.

Separately, on @percymcn's point about the Gemini path being the converter's only real consumer: that matches what I found when I first went through this, and it is the strongest argument in the thread for prioritising the review. Worth pulling into the PR description, since the `to_regex` example currently in there is a path users assemble by hand while the Gemini one is what they hit without asking for it.

PR #1965, opened 2026-07-28
fix(llguidance): raise a clear error for CFG literals matching special tokens

### LINKED ISSUE #1771: Using a CFG with a <think>.+</think> section, when there is a special token <think>, breaks the CFG with "ParserTooComplex"

Use case: I want to constrain my output with a CFG, and I want some arbitrary thinking to happen beforehand.
How I am solving this: pass a CFG with an explicit <think> section in the beginning, and then use my grammar.
What I have found: when my LLM has a tokenizer that includes a <think> token, this breaks; when the tokenizer doesn't have that in its vocabulary, everything is fine.
Potential workaround: run inference once, extract the thinking section, run inference again with CFG with the thinking section pre-stuffed in the assistant's response.

This is distinct from #1627 .

The attached code breaks on Qwen3-4B-Thinking, but works fine on SmolLM2. Crucially, there is a ParserTooComplex error when the tokenizer vocabulary includes <think>, and there is no error when the vocabulary doesn't.

```python
"""
Minimal reproducible example for outlines CFG bug with <think> special tokens.

This demonstrates that when a model has special tokens for <think> and </think>,
outlines CFG grammar fails to parse them correctly.

Expected behavior: Grammar should constrain output to have <think>...</think> followed by yes|no
Actual behavior: Parser error when trying to match special tokens against literal strings

Model: Qwen/Qwen3-4B-Thinking-2507 (has <think> token ID 151667, </think> token ID 151668)
"""

import transformers
from outlines import Transformers
from outlines.types import CFG


def main():
    print("=== Outlines CFG Bug: Special Tokens in Grammar ===\n")

    print(f"Loading model...")
    pipe = transformers.pipeline(
        "text-generation",
        # "HuggingFaceTB/SmolLM2-1.7B-Instruct",
        "Qwen/Qwen3-4B-Thinking-2507",
    )

    # Show that <think> and </think> are special tokens
    print("\n--- Tokenizer Analysis ---")
    vocab = pipe.tokenizer.get_vocab()
    think_start_id = vocab.get('<think>')
    think_end_id = vocab.get('</think>')

    print(f"<think> token ID: {think_start_id}")
    print(f"</think> token ID: {think_end_id}")

    # Show how they encode
    encoded_start = pipe.tokenizer.encode('<think>', add_special_tokens=False)
    encoded_end = pipe.tokenizer.encode('</think>', add_special_tokens=False)
    print(f"<think> encodes to: {encoded_start} (single token)")
    print(f"</think> encodes to: {encoded_end} (single token)")

    # Create outlines model
    print("\n--- Setting up Outlines ---")
    model = Transformers(pipe.model, pipe.tokenizer)

    # Define a grammar that includes <think> tags
    # This SHOULD work but DOESN'T due to special token handling
    grammar_with_thinking = '''
?start: thinking_section answer
thinking_section: "<think>" /[^<]*/ "</think>" /[\\r\\n\\t ]*/
answer: "yes" | "no"
'''

    print("Grammar:")
    print(grammar_with_thinking)

    cfg_type = CFG(grammar_with_thinking)
    prompt = "Is the sky blue?"

    print(f"\n--- Attempting Generation ---")
    print(f"Prompt: {prompt}")
    print("Expected: <think>reasoning here</think>\\nyes")
    print("\nGenerating...")

    try:
        response = model(prompt, cfg_type, max_new_tokens=10000)
        print(f"\nSuccess! Response: {response}")
    except Exception as e:
        print(f"\n❌ ERROR: {type(e).__name__}: {e}")
        print("\nThis demonstrates the bug: outlines cannot match special tokens")
        print("in the grammar against the tokenizer's single-token representation.")

    # Show that a grammar without <think> tags works fine
    print("\n\n--- Testing Grammar Without Special Tokens ---")
    grammar_without_thinking = '''
?start: answer
answer: "yes" | "no"
'''

    print("Grammar (no special tokens):")
    print(grammar_without_thinking)

    cfg_type_simple = CFG(grammar_without_thinking)

    try:
        response = model(prompt, cfg_type_simple, max_new_tokens=10)
        print(f"\n✓ Success! Response: {response}")
        print("\nThis works because there are no special tokens in the grammar.")
    except Exception as e:
        print(f"\n❌ ERROR: {type(e).__name__}: {e}")

if __name__ == "__main__":
    main()
```

### Expected result:

```shell
By uncommenting the SmolLM2 model specification and commenting the Qwen3 model specification, the code runs through with two successes, constrained and unconstrained.
```

### Error message:

```shell
.venv/lib/python3.13/site-packages/outlines/backends/llguidance.py:175: UserWarning: Error in LLMatcher: Parser Error: token "�[151667]" doesn't satisfy the grammar; forced bytes: got '<'; applying 'ÿ'
<state>
Tokens: ⟦<think>⟧
1 tokens, 0 bytes; grm_prefix: ""
Flags:
Parser: {
  "compute_time_us": 0,
  "rows": 2,
  "cached_rows": 0,
  "all_items": 4,
  "lexer_cost": 3271,
  "slices_applied": 0,
  "trie_nodes_walked": 0,
  "definitive_bytes": 7,
  "lexer_ops": 0,
  "num_lex_errors": 0,
  "num_lexemes": 0
}
Stop: ParserTooComplex
Error: Parser Error: token "�[151667]" doesn't satisfy the grammar; forced bytes: got '<'; applying 'ÿ'
</state><grammar>

?start: thinking_section answer
thinking_section: "<think>" /[^<]*/ "</think>" /[\r\n\t ]*/
answer: "yes" | "no"

</grammar>
```

Version information
<details>
```
% python -c "from outlines import _version; print(_version.version)"; python -c "import sys; print('Python', sys.version)"; uv pip freeze;
1.2.7
Python 3.13.3 (main, Apr  8 2025, 13:54:08) [Clang 17.0.0 (clang-1700.0.13.3)]
accelerate==1.10.1
aiofiles==24.1.0
aiohappyeyeballs==2.6.1
aiohttp==3.13.0
aiosignal==1.4.0
annotated-types==0.7.0
anyio==4.11.0
attrs==25.4.0
audioop-lts==0.2.2
brotli==1.1.0
certifi==2025.10.5
charset-normalizer==3.4.4
click==8.3.0
cloudpickle==3.1.1
datasets==4.2.0
dill==0.4.0
diskcache==5.6.3
fastapi==0.119.0
ffmpy==0.6.3
filelock==3.20.0
frozenlist==1.8.0
fsspec==2025.9.0
genson==1.3.0
gradio==5.49.1
gradio-client==1.13.3
groovy==0.1.2
h11==0.16.0
hf-xet==1.1.10
httpcore==1.0.9
httpx==0.28.1
huggingface-hub==0.35.3
idna==3.11
iniconfig==2.1.0
jinja2==3.1.6
joblib==1.5.2
jsonpath-ng==1.7.0
jsonschema==4.25.1
jsonschema-specifications==2025.9.1
llguidance==1.2.0
markdown-it-py==4.0.0
markupsafe==3.0.3
mdurl==0.1.2
mpmath==1.3.0
multidict==6.7.0
multiprocess==0.70.16
networkx==3.5
ninja==1.13.0
numpy==2.3.4
optimum-quanto==0.2.7
orjson==3.11.3
outlines==1.2.7
outlines-core==0.2.11
packaging==25.0
pandas==2.3.3
pillow==11.3.0
pluggy==1.6.0
ply==3.11
propcache==0.4.1
psutil==7.1.0
pyarrow==21.0.0
pydantic==2.11.10
pydantic-core==2.33.2
pydub==0.25.1
pygments==2.19.2
pytest==8.4.2
python-dateutil==2.9.0.post0
python-multipart==0.0.20
pytz==2025.2
pyyaml==6.0.3
referencing==0.37.0
regex==2025.9.18
requests==2.32.5
rich==14.2.0
rpds-py==0.27.1
ruff==0.14.0
safehttpx==0.1.6
safetensors==0.6.2
scikit-learn==1.7.2
scipy==1.16.2
semantic-version==2.10.0
sentence-transformers==5.1.1
sentencepiece==0.2.1
setuptools==80.9.0
shellingham==1.5.4
six==1.17.0
sniffio==1.3.1
starlette==0.48.0
sympy==1.14.0
threadpoolctl==3.6.0
tokenizers==0.22.1
tomlkit==0.13.3
torch==2.9.0
tqdm==4.67.1
transformers==4.57.1
typer==0.19.2
typing-extensions==4.15.0
typing-inspection==0.4.2
tzdata==2025.2
urllib3==2.5.0
uvicorn==0.37.0
websockets==15.0.1
xxhash==3.6.0
yarl==1.22.0
```
</details>

### Context for the issue:

_No response_

VenkateswarluNagineni:
## Problem

Fixes #1771.

A CFG literal that exactly matches one of the tokenizer's special tokens (e.g. `<think>` on reasoning-model tokenizers like Qwen3-4B-Thinking) makes llguidance's parser fail with an opaque `ParserTooComplex` error at generation time:

```
Parser Error: token "<think>[151667]" doesn't satisfy the grammar; forced bytes: got '<'; applying 'ÿ'
Stop: ParserTooComplex
```

Root cause: the grammar's literal terminal (`"<think>"`) is written for byte-level matching, but the tokenizer emits `<think>` as a single atomic token rather than as decomposable byte fragments. llguidance's parser can't reconcile an atomic token against a literal-string terminal, and bails out with a hard-to-diagnose error deep in generation.

## What this does

This doesn't attempt to fix llguidance's parser itself (that lives in a separate compiled dependency, `guidance-ai/llguidance`, and reasoning-model support there looks like a bigger design effort — the maintainer's comment on the issue mentions this is being worked on). Instead, it detects the conflict at grammar-compile time, where we have both the grammar and the tokenizer's vocabulary available, and raises a clear `ValueError` naming the conflicting literal(s) instead of letting the confusing runtime parser failure surface later during generation.

Scoped to the `Transformers` backend (where the tokenizer's vocabulary is available) since that's what's reported and what I could verify. `LlamaCpp` and `MLXLM` tokenizers expose this differently and aren't covered here.

**Update:** the initial version of this fix only checked `all_special_tokens`, which misses tokens registered with `special=False`. Reasoning-model tokenizers commonly register `<think>`/`</think>` exactly this way (verified against the real `Qwen/Qwen3-4B-Thinking-2507` tokenizer — `<think>` is absent from `all_special_tokens`) specifically so `skip_special_tokens=True` decoding doesn't strip them mid-generation, but the tokenizer still emits them atomically, bypassing ordinary BPE merging exactly like a true special token does. Without this, the guard would have silently let the original issue's exact grammar through to the same opaque `ParserTooComplex` failure it was meant to catch. The check now covers the union of `all_special_tokens` and every token in `added_tokens_decoder`, since the `special` flag only governs decode-time stripping, not which tokens bypass BPE merging.

## Testing

- Reproduced the original issue locally: built the llguidance matcher directly from the real `Qwen/Qwen3-4B-Thinking-2507` tokenizer (no model weights needed for the grammar-compile-time check — confirmed separately end-to-end with the actual model loaded too) and confirmed the exact `ParserTooComplex` failure from the issue, then confirmed the fix raises a clear `ValueError` instead.
- Added `test_cfg_logits_processor_rejects_literal_matching_special_token`, using the existing lightweight `erwanf/gpt2-mini` fixture already in this test file (its `<|endoftext|>` special token stands in for `<think>`, so the test doesn't need network access to a 4B+ model).
- Added `test_cfg_logits_processor_allows_literal_not_matching_special_token` to confirm ordinary grammars are unaffected.
- Added `test_cfg_logits_processor_rejects_literal_matching_added_non_special_token`, reproducing the added-but-not-special gap via `tokenizer.add_tokens(["<think>"], special_tokens=False)` on the same lightweight fixture — no network access to the 4B model needed here either.
- Ran the full `test_llguidance_backend[model_transformers-torch]` parametrization (the one exercising `get_cfg_logits_processor` with the existing `cfg_lark`/`cfg_ebnf` fixtures end-to-end through real generation) to confirm no regression.
- `ruff check` clean on both changed files (matches the pinned pre-commit ruff version, 0.9.1). `pre-commit run` mypy reports two pre-existing, unrelated `urllib3` stub errors in `src/outlines/exceptions.py` present identically without this change (confirmed by re-running against the base commit) — not introduced by this PR.

### DIFF
--- a/src/outlines/backends/llguidance.py
+++ b/src/outlines/backends/llguidance.py
@@ -226,9 +226,13 @@ class LLGuidanceBackend(BaseBackend):
             Every token string the tokenizer treats as atomic.
 
         """
+        # `added_tokens_decoder` was added in transformers 4.34; guard against
+        # older installations so a CFG-only feature doesn't break JSON-schema
+        # and regex users who happen to have an older tokenizer.
+        added_tokens_decoder = getattr(hf_tokenizer, "added_tokens_decoder", {})
         added = {
             added_token.content
-            for added_token in hf_tokenizer.added_tokens_decoder.values()
+            for added_token in added_tokens_decoder.values()
         }
         return frozenset(hf_tokenizer.all_special_tokens) | frozenset(added)
 
@@ -401,7 +405,11 @@ class LLGuidanceBackend(BaseBackend):
         -------
         set[str]
             The decoded text of every double- or single-quoted literal found
-            outside of a `//` comment.
+            outside of a `//` comment. Note: the scanner matches quoted runs
+            by syntax alone and does not distinguish Lark string literals from
+            quoted content embedded inside regexp terminals (e.g. the `"[^"`
+            portion of `STR: /"[^"]*"/`). Such fragments cannot equal a real
+            token string, so they cannot produce false positives.
 
         """
         literals = set()
--- a/src/outlines/backends/llguidance.py
+++ b/src/outlines/backends/llguidance.py
@@ -1,5 +1,6 @@
 """Backend class for LLGuidance."""
 
+import re
 import warnings
 from typing import TYPE_CHECKING
 
@@ -193,6 +194,11 @@ class LLGuidanceBackend(BaseBackend):
         self.llg = llg
         self.tensor_library_name = model.tensor_library_name
         self.llg_tokenizer = self._create_llg_tokenizer(model)
+        self._special_tokens: frozenset[str] = (
+            frozenset(model.hf_tokenizer.all_special_tokens)
+            if isinstance(model, Transformers)
+            else frozenset()
+        )
 
     def _create_llg_tokenizer(self, model: SteerableModel) -> "LLGTokenizer":
         """Create an llg tokenizer from the Outlines model's tokenizer.
@@ -300,6 +306,7 @@ class LLGuidanceBackend(BaseBackend):
             The logits processor to use to constrain the generation.
 
         """
+        self._check_grammar_literals_against_special_tokens(grammar)
         # We try both lark and ebnf
         try:
             grammar_spec = self.llg.grammar_from("grammar", grammar)
@@ -308,3 +315,37 @@ class LLGuidanceBackend(BaseBackend):
         return LLGuidanceLogitsProcessor(
             grammar_spec, self.llg_tokenizer, self.tensor_library_name
         )
+
+    def _check_grammar_literals_against_special_tokens(self, grammar: str) -> None:
+        """Raise a clear error if a grammar literal matches a special token.
+
+        When a quoted literal in the grammar exactly matches one of the
+        tokenizer's special tokens (e.g. `<think>`), llguidance's parser can
+        fail with an opaque `ParserTooComplex` error at generation time: the
+        model may emit the literal as a single atomic special token, which
+        the parser cannot match against a literal-string terminal the way it
+        matches ordinary, byte-decomposable tokens. Surfacing this at grammar
+        compilation time, with the actual conflicting token named, is far
+        more actionable than the downstream parser failure.
+
+        Parameters
+        ----------
+        grammar: str
+            The context-free grammar to check.
+
+        """
+        if not self._special_tokens:
+            return
+        literals = set(re.findall(r'"((?:[^"\\]|\\.)*)"', grammar))
+        conflicts = sorted(literals & self._special_tokens)
+        if conflicts:
+            raise ValueError(
+                "The grammar contains the literal(s) "
+                f"{conflicts} which exactly match special token(s) in the "
+                "tokenizer's vocabulary. The llguidance backend cannot "
+                "reliably constrain generation around a special token used "
+                "as ordinary grammar text, as the model may emit it as a "
+                "single atomic token that the parser cannot match against a "
+                "literal-string terminal. Avoid embedding the text of a "
+                "special token as a literal in your grammar."
+            )
--- a/src/outlines/backends/llguidance.py
+++ b/src/outlines/backends/llguidance.py
@@ -348,6 +348,72 @@ class LLGuidanceBackend(BaseBackend):
             grammar_spec, self.llg_tokenizer, self.tensor_library_name
         )
 
+    # Matches a Lark grammar's double- or single-quoted string literals, or a
+    # `//`-to-end-of-line comment. Alternation order matters: a string
+    # alternative always starts matching at its opening quote, so a `//`
+    # occurring inside a literal (e.g. `"http://foo"`) is consumed as part of
+    # that literal rather than being mistaken for the start of a comment, and
+    # a literal written inside a real comment is never extracted at all,
+    # since only the `dq`/`sq` groups are inspected by the caller.
+    _LITERAL_OR_COMMENT_RE = re.compile(
+        r'"(?P<dq>(?:[^"\\]|\\.)*)"'
+        r"|'(?P<sq>(?:[^'\\]|\\.)*)'"
+        r"|//[^\n]*"
+    )
+    # The backslash escapes a grammar author can realistically write inside a
+    # Lark string literal: \\, \", \', \n, \t, \r, and \uXXXX.
+    _ESCAPE_RE = re.compile(r"\\u([0-9a-fA-F]{4})|\\(.)")
+    _SIMPLE_ESCAPES = {"n": "\n", "t": "\t", "r": "\r", "\\": "\\", '"': '"', "'": "'"}
+
+    @classmethod
+    def _decode_literal_escapes(cls, literal: str) -> str:
+        """Decode backslash escapes in a raw grammar string literal.
+
+        Parameters
+        ----------
+        literal: str
+            The literal's source text, as captured between its quotes.
+
+        Returns
+        -------
+        str
+            The string the literal denotes, so it can be compared against
+            actual token text rather than against its escaped source form.
+
+        """
+        def _replace(match: "re.Match[str]") -> str:
+            hex_digits, other = match.groups()
+            if hex_digits is not None:
+                return chr(int(hex_digits, 16))
+            return cls._SIMPLE_ESCAPES.get(other, other)
+
+        return cls._ESCAPE_RE.sub(_replace, literal)
+
+    def _extract_grammar_literals(self, grammar: str) -> set[str]:
+        """Extract every quoted string literal's decoded text from a grammar.
+
+        Parameters
+        ----------
+        grammar: str
+            The context-free grammar to scan.
+
+        Returns
+        -------
+        set[str]
+            The decoded text of every double- or single-quoted literal found
+            outside of a `//` comment.
+
+        """
+        literals = set()
+        for match in self._LITERAL_OR_COMMENT_RE.finditer(grammar):
+            raw = match.group("dq")
+            if raw is None:
+                raw = match.group("sq")
+            if raw is None:
+                continue  # a `//` comment match; its contents are not grammar text
+            literals.add(self._decode_literal_escapes(raw))
+        return literals
+
     def _check_grammar_literals_against_special_tokens(self, grammar: str) -> None:
         """Raise a clear error if a grammar literal matches an atomic token.
 
@@ -364,6 +430,13 @@ class LLGuidanceBackend(BaseBackend):
         conflicting token named, is far more actionable than the downstream
         parser failure.
 
+        The check considers both `"..."` and `'...'` literals (Lark accepts
+        either), decodes standard backslash escapes before comparing, and
+        ignores literal-looking text inside `//` comments. It only catches an
+        *exact* match between a literal and an atomic token — a literal that
+        merely embeds the token's text as a substring (e.g. `"<think>: go"`)
+        is not detected.
+
         Parameters
         ----------
         grammar: str
@@ -372,7 +445,7 @@ class LLGuidanceBackend(BaseBackend):
         """
         if not self._atomic_literal_tokens:
             return
-        literals = set(re.findall(r'"((?:[^"\\]|\\.)*)"', grammar))
+        literals = self._extract_grammar_literals(grammar)
         conflicts = sorted(literals & self._atomic_literal_tokens)
         if conflicts:
             raise ValueError(
--- a/src/outlines/backends/llguidance.py
+++ b/src/outlines/backends/llguidance.py
@@ -194,11 +194,43 @@ class LLGuidanceBackend(BaseBackend):
         self.llg = llg
         self.tensor_library_name = model.tensor_library_name
         self.llg_tokenizer = self._create_llg_tokenizer(model)
-        self._special_tokens: frozenset[str] = (
-            frozenset(model.hf_tokenizer.all_special_tokens)
+        self._atomic_literal_tokens: frozenset[str] = (
+            self._get_atomic_literal_tokens(model.hf_tokenizer)
             if isinstance(model, Transformers)
             else frozenset()
         )
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
The diagnosis reads right to me, and the docstring explaining why `added_tokens_decoder` is included alongside `all_special_tokens` — that `special` only governs decode-time stripping, not whether the tokenizer splits the text out before BPE — is the part I'd have got wrong. Worth keeping.

The literal extraction is where I'd look again. Running the regex over a few grammar shapes:

| grammar | literals found | conflict raised |
|---|---|---|
| `start: "<think>" WORD` | `['<think>']` | yes |
| `start: '<think>' WORD` | `[]` | **no** |
| `start: "<think>" WORD` | `['\\u003cthink\\u003e']` | **no** |
| `// example: "<think>"` + rule | `['<think>']` | **yes, from a comment** |
| `start: "say\"hi" WORD` | `['say\\"hi']` | **no** |

Four things fall out of that.

**Single-quoted literals are missed.** Lark accepts `'<think>'` and `"<think>"` interchangeably, and the pattern only looks at double quotes. Someone hitting the original `ParserTooComplex` and switching quote style to see if it helps would get the opaque error back with no explanation. Adding `'((?:[^'\\]|\\.)*)'` as an alternation covers it.

**Escapes aren't decoded before comparison.** The capture keeps the raw text, so `"<think>"` and `"say\"hi"` are compared as their source form rather than the string they denote. Both are ways to write a conflicting literal that the check waves through. The escapes a grammar author would realistically use here are `\"` and `\uXXXX`, so decoding those two before the set intersection would close most of it.

**A literal inside a comment raises.** That's the direction that costs the user something — a valid grammar refuses to compile because a token name appears in a `//` comment. Stripping comments before the scan, or at least mentioning the limitation in the error, would avoid a confusing hard failure.

**The check is exact-match only.** `"<think>abc"` embeds the atomic token without equalling it, so it passes the check and presumably still confuses the parser. That may be deliberate scoping — matching substrings would produce false positives on any grammar containing `<` — but if so it's worth a line saying the check covers exact literals only, since the error message reads like a general guarantee.

None of these break the reported case; they're all ways the same bug reaches a user without the new error firing. Given the point of the PR is turning an opaque failure into an actionable one, the quote-style gap in particular seems worth closing before merge — it's the first thing someone debugging this would try.

VenkateswarluNagineni:
Thanks for the close read — the regex table made the gaps concrete. Pushed `269a7f87` to address all three:

- **Single-quoted literals**: the scanner now matches `'...'` as well as `"..."`.
- **Undecoded escapes**: literal text is now escape-decoded (`\\`, `\"`, `\'`, `\n`, `\t`, `\r`, `\uXXXX`) before comparing against token text, so `"<think>"` is caught same as `"<think>"`.
- **Literal inside a comment**: rather than stripping comments as a separate pass, I combined literal and comment matching into one regex scan — a `//` that occurs *inside* a string (e.g. `"http://foo"`) is consumed as part of that string because the string alternative starts matching at the opening quote, while a real `// comment` outside any string is matched by the comment branch and its contents are never extracted as literals. That also sidesteps the risk of a separate comment-stripping pass mangling a literal that legitimately contains `//`.

On exact-match-only: kept that scoping as-is (matching substrings would indeed produce false positives on grammars containing `<` etc.), but added a line to the docstring making clear the check only catches exact literal matches, not embedded occurrences.

Added regression tests for each of the four cases from your table (single-quote, `\u`-escape, comment false-positive, plus a same-shape check that a literal legitimately containing `//` still compiles). Full `test_llguidance.py` CFG-literal suite passes locally; ruff and mypy clean on the touched files.

ErenAta16 (commented):
Re-ran the same table against `269a7f87`. All five rows flip, and the `//`-inside-a-literal case the new comment mentions holds too:

| grammar | literals found | conflict |
|---|---|---|
| `start: "<think>" W` | `['<think>']` | yes |
| `start: '<think>' W` | `['<think>']` | **yes** (was missed) |
| `start: "<think>" W` | `['<think>']` | **yes** (was missed) |
| `// ornek: "<think>"` + rule | `[]` | **none** (was a false positive) |
| `start: "say\"hi" W` | `['say"hi']` | **yes** (was missed) |
| `start: "http://foo" W` | `['http://foo']` | none |

The alternation ordering is the part worth keeping the comment on. Putting the string alternatives before `//[^\n]*` means a `//` inside a literal gets consumed by the string match, so `"http://foo"` survives intact instead of having the rest of the line swallowed as a comment — and because only `dq`/`sq` are read, a literal written inside a real comment never reaches the set. Two problems solved by one ordering decision, which isn't obvious from reading the regex alone.

Decoding `\uXXXX` alongside the simple escapes covers the case I'd worry about most in practice, since a grammar author working around the original `ParserTooComplex` is fairly likely to try escaping the angle brackets.

Substring matching still isn't covered (`"<think>abc"` passes), but you've said so explicitly in the docstring now, which was the ask — the error message no longer reads like a guarantee it can't make.

ErenAta16:
Ran the new scanner against the cases I was worried about rather than eyeballing it, and all three gaps are genuinely closed. Extracting with `_LITERAL_OR_COMMENT_RE` + `_decode_literal_escapes` at `269a7f87`:

| grammar | extracted |
|---|---|
| `start: "<think>"` | `<think>` |
| `start: '<think>'` | `<think>` |
| `start: "\u003cthink\u003e"` | `<think>` |
| `// avoid "<think>"` + `start: "ok"` | `ok` |
| `start: "http://x" \| "<think>"` | `http://x`, `<think>` |
| `// don't` + `start: "<think>"` | `<think>` |
| `// stray "` + `start: "<think>"` | `<think>` |
| `start: "don't"` | `don't` |
| `start: 'say "hi"'` | `say "hi"` |
| `start: "a\"b"` | `a"b` |

The two that could plausibly have gone wrong both hold up. An unpaired `'` or `"` inside a comment doesn't desynchronise the scan, because the comment alternative is tried at the `//` before the scan ever reaches the stray quote. And `"http://x"` survives, since the string alternative starts matching at the opening quote and swallows the `//`. Combining the two into one pass was the right call — a separate comment-stripping pass is exactly where that class of bug lives.

Two observations, neither blocking:

**Lark regexp terminals aren't distinguished from string literals.** `STR: /"[^"]*"/` extracts `[^` as a literal. It's harmless in practice — a fragment like that can't equal a token string, so it can't produce a false error — but the extracted set isn't strictly "grammar literals", it's "quoted runs". Might be worth a word in the docstring so nobody later builds something stricter on top of it.

**`added_tokens_decoder` is read unguarded in `__init__`.** `pyproject.toml` doesn't pin `transformers`, and the attribute only exists from 4.34 onward. Because the call sits in the constructor rather than in `get_cfg_logits_processor`, a tokenizer without it takes down `LLGuidanceBackend` construction entirely, so JSON-schema and regex users would hit an `AttributeError` from a CFG-only feature. `getattr(hf_tokenizer, "added_tokens_decoder", {})` would keep the blast radius on the code path that actually needs it.

The `special=False` reasoning in the docstring is the part I'd most want kept — that added-but-not-special tokens still bypass BPE is the non-obvious fact that makes this check correct, and it's the sort of thing that gets "simplified" away later without it written down.

VenkateswarluNagineni:
Pushed `bf148779` addressing both follow-ups:

1. **`added_tokens_decoder` guard**: changed to `getattr(hf_tokenizer, "added_tokens_decoder", {})` — falls back to an empty dict on `transformers < 4.34`, so the atomic-token set degrades to `all_special_tokens` only rather than raising `AttributeError` for JSON-schema and regex users.

2. **Regexp terminal note**: added a sentence to the `_extract_grammar_literals` docstring acknowledging that the scanner matches quoted runs without regard to surrounding Lark syntax, so content like the `"[\^"` fragment inside `STR: /"[\^"]*"/` may be extracted — but can't equal any real token string and can't produce a false positive.

The special/added token distinction in the docstring is unchanged, as requested.

PR #1966, opened 2026-07-29
Add e164, latitude, and longitude custom types

uditmahato:
## Summary

Adds three custom types to `outlines.types`, continuing #1303:

- `e164`: ITU-T E.164 international phone numbers in canonical `+` form, complementing the per-country patterns in the `locale` submodule
- `latitude` / `longitude`: geographic coordinates in signed decimal degrees, with the `[-90, 90]` / `[-180, 180]` bounds encoded in the pattern

Each type follows the structure of the previously merged type additions (#1863, #1934, #1945): a `Regex(...)` definition with a comment stating the reference and the intentional scope limits, an `__all__` entry, and parametrized positive and negative test cases.

## Scope notes

- `e164` accepts only the canonical form: a leading `+`, a country code starting with a non-zero digit, and at most fifteen digits in total. Spaced and hyphenated national formats are intentionally excluded, and number assignment is not validated.
- `latitude` / `longitude` accept an optional sign and fractional part; the bounds themselves (`90` / `180`) admit only a zero fraction, and leading zeros in the integer part are excluded. Degree-minute-second and cardinal-direction notations are out of scope.

## Testing

- `pytest tests/types/test_custom_types.py`: 203 passed (30 new cases)
- `ruff check --config=pyproject.toml` and `mypy --allow-redefinition` pass on both changed files

### DIFF
--- a/src/outlines/types/__init__.py
+++ b/src/outlines/types/__init__.py
@@ -91,6 +91,9 @@ __all__ = [
     "credit_card",
     "iban",
     "bic",
+    "e164",
+    "latitude",
+    "longitude",
     # Document-specific types
     "sentence",
     "paragraph",
@@ -215,6 +218,20 @@ bic = Regex(
     r"[A-Z2-9][A-NP-Z0-9]"  # location code
     r"(?:[A-Z0-9]{3})?"  # optional branch code, "XXX" for the head office
 )
+
+# ITU-T E.164 international phone numbers: https://www.itu.int/rec/T-REC-E.164
+# Canonical text form only: a leading plus, a country code starting with a
+# non-zero digit, and at most fifteen digits in total. Spacing, hyphens, and
+# national dialing prefixes are intentionally excluded, and number assignment
+# is not validated. For national conventions see the `locale` submodule.
+e164 = Regex(r"\+[1-9]\d{1,14}")
+
+# Geographic coordinates in signed decimal degrees. Latitude is bounded to
+# [-90, 90] and longitude to [-180, 180]; the bounds themselves admit only a
+# zero fractional part. Leading zeros in the integer part and the
+# degree-minute-second and cardinal-direction notations are excluded.
+latitude = Regex(r"[+-]?(?:90(?:\.0+)?|[1-8]?\d(?:\.\d+)?)")
+longitude = Regex(r"[+-]?(?:180(?:\.0+)?|(?:1[0-7]\d|[1-9]?\d)(?:\.\d+)?)")
 
 # Document-specific types
 sentence = Regex(r"[A-Z].*\s*[.!?]")
--- a/src/outlines/types/__init__.py
+++ b/src/outlines/types/__init__.py
@@ -224,12 +224,17 @@ bic = Regex(
 # non-zero digit, and at most fifteen digits in total. Spacing, hyphens, and
 # national dialing prefixes are intentionally excluded, and number assignment
 # is not validated. For national conventions see the `locale` submodule.
+# Like the other built-in string types, this is meant for standalone use: inside
+# a JSON container it is currently generated unquoted (#1962).
 e164 = Regex(r"\+[1-9]\d{1,14}")
 
 # Geographic coordinates in signed decimal degrees. Latitude is bounded to
 # [-90, 90] and longitude to [-180, 180]; the bounds themselves admit only a
 # zero fractional part. Leading zeros in the integer part and the
 # degree-minute-second and cardinal-direction notations are excluded.
+# These are text forms meant for standalone use: inside a JSON container they
+# are currently generated unquoted (#1962), which parses as a number rather
+# than the string the schema declares.
 latitude = Regex(r"[+-]?(?:90(?:\.0+)?|[1-8]?\d(?:\.\d+)?)")
 longitude = Regex(r"[+-]?(?:180(?:\.0+)?|(?:1[0-7]\d|[1-9]?\d)(?:\.\d+)?)")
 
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Went through the three patterns properly rather than eyeballing them, since bounded numeric ranges expressed as regex are the kind of thing that is usually subtly wrong. These are not. Ran the branch at `2aa696c` against a set of cases beyond the ones in the PR and every single one landed where it should:

`e164`, `\+[1-9]\d{1,14}`:

```
+14155552671        accept      +1                  reject  (one digit)
+12                 accept      +1234567890123456   reject  (16 digits)
+123456789012345    accept      +0155552671         reject  (zero country code)
                                +                   reject
                                ++1415              reject
                                +1415555267a        reject
```

Fifteen digits is the E.164 maximum and `[1-9]` plus `\d{1,14}` lands exactly on it, which is the boundary most implementations get wrong by one.

`latitude`, `[+-]?(?:90(?:\.0+)?|[1-8]?\d(?:\.\d+)?)`:

```
0, 9, 89, 89.999999, 90, 90.0, 90.00, -90, +0, -0    accept
90.1, -90.1, 91, 05, .5, 5., 1e5                     reject
```

`longitude`, `[+-]?(?:180(?:\.0+)?|(?:1[0-7]\d|[1-9]?\d)(?:\.\d+)?)`:

```
0, 18, 100, 179.999, 180, 180.0, -180                accept
180.1, 181, 189, 099, 1e5                            reject
```

The two things I specifically tried to break are both handled. Restricting the fractional part to `\.0+` on the bound itself means `90.1` and `180.1` cannot sneak in through the general branch, and the `[1-8]?` / `[1-9]?` optional first digit rejects leading zeros without also rejecting single-digit values. `189` is the case a naive `1\d\d` would have let through and `1[0-7]\d` catches it. Nice work on the boundaries.

Separate from the patterns, one thing worth knowing before these land, because it affects how they behave in the most common usage rather than standalone. All three declare themselves as strings:

```
e164       declared JSON Schema type: 'string'
latitude   declared JSON Schema type: 'string'
longitude  declared JSON Schema type: 'string'
```

but inside a container they generate unquoted, so what comes out does not match that declaration:

```
list[e164]        [+14155552671]   matches, and json.loads rejects it
                  ["+14155552671"] does not match
list[latitude]    [45.5231]        matches, parses as a JSON number
                  ["45.5231"]      does not match
list[longitude]   [-122.6765]      matches, parses as a JSON number
                  ["-122.6765"]    does not match
```

`e164` is the plain version of the problem, `[+14155552671]` is not parseable at all. `latitude` and `longitude` are the sneakier version: they happen to produce something a JSON parser accepts, but as a number rather than the string the schema promised, so a consumer reading the field as a string gets the wrong type without any error to tell them.

This is not something this PR introduces. `types.email` behaves identically today (`[a@b.co]` matches, `["a@b.co"]` does not) and it is the gap tracked in #1962, built-in terms reach `_ensure_json_quoted` through the `isinstance(ptype, Term)` shortcut in `python_types_to_terms` and stay bare. So I would not block on it. It is worth flagging here only because three more string-shaped types widens the surface, and because the numeric-looking ones fail more quietly than the existing cases do, which makes them a worse thing to discover later.

If you want to pre-empt it, a line in the docstrings saying these are canonical text forms intended for standalone use, with #1962 linked, would set expectations without waiting on that fix.

Tests are well chosen, particularly `+04155552671` and the spacing and hyphen cases, since those are the forms a model is most likely to produce if the pattern were loose. Adding `+123456789012345` at exactly fifteen digits next to the existing sixteen-digit rejection would pin the upper bound from both sides rather than just above it.

uditmahato:
Thanks for going through the patterns case by case, that is a genuinely useful review. I reproduced the container behaviour on my side before acting on it, and it is exactly as you describe:

```
list[e164]       [+14155552671]  matches -> json.loads raises
list[latitude]   [45.5231]       matches -> parses as float, not the declared string
list[email]      [a@b.co]        matches -> json.loads raises   (pre-existing, same shape)
```

The `latitude` / `longitude` case being silently wrong rather than loudly wrong is the part worth flagging, agreed. Since `email` behaves identically today and the fix belongs in `python_types_to_terms` (#1962) rather than in these patterns, I have not tried to work around it here.

Both suggestions applied in de6a58f:

- Added a line to the comments on all three types noting they are canonical text forms meant for standalone use, with #1962 linked, so the expectation is set at the definition site.
- Added `+123456789012345` at exactly fifteen digits, so the upper bound is now pinned from both sides rather than only from above.

204 tests pass, ruff and mypy clean.

ErenAta16 (commented):
Reproducing the container behaviour independently before acting on it is the right instinct, and your table matches mine exactly, including the part that matters most:

```
list[latitude]  [45.5231]  matches -> parses as a JSON number, not the declared string
```

Agreed on scope. `email` behaves identically today, the cause is in `python_types_to_terms` rather than in these three patterns, and working around it here would mean putting a fix in the wrong layer where the next person would not find it. Leaving it to #1962 is correct.

The patterns themselves were the part I checked hardest and they held up on every boundary I threw at them, so with the two suggestions applied I have nothing left. The fifteen-digit E.164 case pinned from both sides is the one I would have most expected to be off by one, and it is not.

uditmahato:
Thanks for the careful review. Testing the boundaries case by case rather than eyeballing them, and digging up the #1962 context, made this a better change than it would have been otherwise.

uditmahato:
Rebased on main after #1979 (iban/bic touched the same two files), conflicts resolved, all 232 type tests green.

RobinPicard (approved):

PR #1967, opened 2026-07-29
fix(gemini): pass the system message as a system instruction

Kaif10:
## Problem

`GeminiTypeAdapter.format_input` sends every `Chat` message into the `contents` argument, mapping `assistant` to `model` but leaving `system` untouched. Gemini's `Content.role` only accepts `"user"` or `"model"` (per the `google-genai` field description: *"The producer of the content. Must be either 'user' or 'model'."*); a system prompt belongs in `GenerateContentConfig.system_instruction`. As a result, `Chat([{"role": "system", ...}, ...])` produces a request containing `{"role": "system", ...}` in `contents`.

The other adapters that need a role translation already handle this: `MistralTypeAdapter` maps `system` to `SystemMessage` (`mistral.py`), and `LMStudioTypeAdapter` hoists a leading system message into the chat's system prompt (`lmstudio.py`). Gemini was the outlier — the Gemini documentation page even works around it by putting the system prompt in a `user` message.

## Fix

`format_chat_model_input` hoists a leading system message out of `contents` and returns it as a `system_instruction` argument; `generate` and `generate_stream` merge the adapter's non-`contents` arguments into the `config` passed to the client. Inputs without a system message are unaffected (no extra key is returned), and user-supplied `inference_kwargs` still take precedence.

A system message that is not the first message keeps its current behaviour: Gemini exposes a single `system_instruction` field, so a mid-conversation or repeated system message has no faithful representation, and `Chat` places the system message first in practice. Happy to raise a `ValueError` there instead if you prefer, as mistral/lmstudio do for unsupported roles.

## Tests

Added `test_gemini_type_adapter_input_chat_system`, asserting the system message lands in `system_instruction` and no longer in `contents`. The existing `test_gemini_type_adapter_input_chat` already pins the no-system-message case (it asserts exact equality against a dict containing only `contents`, so a stray `system_instruction` key would fail it). `pytest tests/models/test_gemini_type_adapter.py` passes (19); the new test fails on `main` with an `AssertionError` showing the `system` role inside `contents`. The `tests/models/test_gemini.py` failures in my environment are the pre-existing `api_call` tests that require an API key, unchanged by this patch. `ruff` and `mypy` are clean on the changed files.

Note: the client types `role` as a plain `str` and does not validate it, so the invalid role is only rejected server-side — the evidence here is structural (the SDK's documented role contract) and at the type-adapter level rather than a live API call.

*AI assistance: this change was written with the help of an AI coding assistant; I reviewed the diff, the design choice and the tests, and can explain every line.*


ErenAta16 (commented):
Confirmed the bug on `main` — `Chat([{"role": "system", ...}, {"role": "user", ...}])` really does put a `system` role into `contents`:

```
main: {'contents': [{'role': 'system', 'parts': [...]}, {'role': 'user', 'parts': [...]}]}
```

and `_create_message` only rewrites `assistant` to `model`, so `system` passes through untouched. Lifting it into `system_instruction` is the right shape.

Two inputs still produce the same bad output on this branch, though, because the lift only looks at `messages[0]`:

| input | `system_instruction` | resulting `contents` roles |
|---|---|---|
| `system, user` | `'S'` | `['user']` |
| `user` | `None` | `['user']` |
| `user, system, user` | `None` | **`['user', 'system', 'user']`** |
| `system, system, user` | `'S1'` | **`['system', 'user']`** |

The third and fourth rows land back in the case the PR is fixing — a `role: "system"` entry inside `contents`, which is what Gemini rejects.

How to handle those is a design call rather than an oversight, so worth deciding explicitly:

- **A system message that isn't first.** Some callers append policy reminders mid-conversation. Gemini has one `system_instruction` slot and no ordering to preserve it in, so the options are folding it in with the leading one or refusing the input. Silently forwarding it as a `contents` role is the one option that's definitely wrong.
- **More than one system message.** Same slot problem. Joining them (newline-separated) matches how the other adapters that collapse system prompts behave; raising is also defensible since the caller can't express two system prompts to Gemini anyway.

Either way, a filter over all messages rather than a check on `messages[0]` closes both, and the current tests would keep passing.

One smaller thing: `config={**input_args, **generation_config, **inference_kwargs}` gives `generation_config` precedence over `system_instruction`. Nothing in `format_output_type` emits that key today so there's no collision, but the ordering means a future output-type key of the same name would silently drop the user's system prompt. Putting `input_args` after `generation_config`, or asserting no overlap, would make that explicit — not a blocker.

The `contents = input_args.pop("contents")` split reads fine, and doing it identically in `generate` and `generate_stream` keeps the two paths from drifting.

ErenAta16 (commented):
The direction is right. Gemini takes the system prompt as `system_instruction` on the config rather than as a turn in `contents`, and `Content.role` genuinely does not accept `"system"`, so the previous code could only ever produce an API error for a chat that had one.

**The guard is positional, and the failure it fixes is not.** It only inspects `messages[0]`:

```python
if messages and messages[0]["role"] == "system":
    formatted["system_instruction"] = messages[0]["content"]
    messages = messages[1:]
```

Everything after that goes through `_create_message`, which rewrites exactly one role:

```python
# Gemini uses "model" instead of "assistant"
if role == "assistant":
    role = "model"
```

so a `system` message anywhere other than index 0 still arrives as `{"role": "system", ...}` and fails the same way it does today. Two shapes reach that: a chat with more than one leading system message, where only the first is lifted out and the second stays in `contents`, and a system message re-injected part-way through a conversation, which some agent frameworks do between turns.

Neither is exotic enough to leave, given the fix for both is small. Collecting every system message rather than only a leading one, and joining them for `system_instruction`, makes the behaviour total instead of depending on position. If multiple system messages should instead be an error, raising a clear one beats letting it through to a Gemini-side failure whose message will not mention `system`.

Related, and cheap while you are here: `_create_message` silently passes through any unrecognised role. Anything that is not `assistant` or `user` becomes an invalid `Content.role` and surfaces as a provider error rather than a validation error at the boundary. An explicit check there would turn every one of these into a local, readable failure, and it would have caught the original bug too.

The docstring addition is accurate and worth having. I would extend it to say what happens to a non-leading system message, whichever way you resolve the above, since that is the question the next reader will have.

Kaif10:
Thanks — you're right on both counts, and the positional guard was the real gap: `user, system, user` and `system, system, user` both still put a `system` role into `contents`, which is exactly what this PR set out to prevent.

Pushed in 564609c:

- **The lift is no longer positional.** All system messages are collected wherever they appear and joined with a newline for `system_instruction`; everything else goes to `contents`. That makes the two shapes you listed produce `system_instruction='...'` with no `system` role in `contents`. I went with joining rather than raising: Gemini has one slot and no ordering to preserve, and joining matches how the other adapters collapse system prompts, so it keeps working input working rather than turning it into an error.
- **Config precedence** is now `{**generation_config, **input_args, **inference_kwargs}`, so a future output-type key named `system_instruction` can't silently drop the user's system prompt. `inference_kwargs` still wins, as before.
- **Docstring** updated to state that system messages are collected regardless of position and joined.

Two tests added alongside the existing one: a non-leading system message, and two system messages. Both fail on the previous commit and pass now; the file is at 21 passing, ruff clean.

On validating unrecognised roles in `_create_message` — I agree it's the right boundary and it would have caught this class of bug. I've left it out of this PR to keep the diff to the reported defect, since it changes behaviour for inputs that currently pass through silently. Happy to add it here if you'd prefer, or as a follow-up.

ErenAta16 (commented):
Ran `564609c` against the two shapes I raised plus a few neighbours, and the lift is total now:

```
case                   system_instruction   contents roles
leading system         'S1'                 ['user']
user, system, user     'S1'                 ['user', 'user']
system, system, user   'S1\nS2'             ['user']
trailing system        'S1'                 ['user']
no system              <absent>             ['user', 'model']
```

No `system` role reaches `contents` in any of them, and `assistant` still maps to `model`. The detail I was not expecting and think is right: when there are no system messages the key is omitted entirely rather than set to an empty string. An empty `system_instruction` is a different request from no system instruction, and sending one would have been easy to do by accident with a `"\n".join([])`.

Joining rather than raising is the right call for the reason you gave. Gemini has one slot, the caller has already expressed intent by supplying several, and dropping all but one silently would be worse than concatenating them. Raising would also make outlines stricter than the provider for no gain.

The unknown-role point is still open and still not a blocker, just noting it so it does not look forgotten:

```
[{"role": "tool", ...}, {"role": "user", ...}]  ->  contents roles ['tool', 'user']
```

Anything that is neither `assistant` nor `system` passes through and becomes an invalid `Content.role` at the provider. That is unchanged by this PR and was already true before it, so it belongs in its own change if you want it at all.

Docstring reads accurately against the behaviour now, including the "regardless of their position" wording, which is the part someone will check first. Looks good to me.

RobinPicard (approved):
Looks good, thanks!

PR #1970, opened 2026-07-29
fix(types): map additionalProperties objects to Dict instead of a generated model

### LINKED ISSUE #1969: Dict[str, V] round-trips to a field-named model instead of a mapping (additionalProperties ignored)

`Dict[str, V]` doesn't survive a JSON-Schema round trip. `schema_type_to_python` treats any `{"type": "object"}` as a nested model and ignores `additionalProperties`, so a mapping field comes back as a generated model instead of `Dict[str, V]`. The generated class is named from the schema `title`, which for a field like `by_user` is `"By User"` — a name with a space in it, not a valid Python identifier.

That means the reconstructed type accepts a fixed (empty) set of keys rather than arbitrary string keys, which is the opposite of what the original annotation meant.

```python
from typing import Dict
from outlines.types.json_schema_utils import (
    json_schema_dict_to_pydantic,
    json_schema_dict_to_dataclass,
)

class Scores(BaseModel):
    by_user: Dict[str, int] = {}

schema = Scores.model_json_schema()
# {"properties": {"by_user": {"additionalProperties": {"type": "integer"},
#                             "default": {}, "title": "By User", "type": "object"}}, ...}

json_schema_dict_to_pydantic(schema).model_fields["by_user"].annotation
json_schema_dict_to_dataclass(schema)
```

### Expected result:

```
Optional[Dict[str, int]]
```

### Error message:

No error — it degrades quietly:

| converter | result |
|---|---|
| `json_schema_dict_to_pydantic` | `Optional[outlines.types.json_schema_utils.By User]` |
| `json_schema_dict_to_dataclass` | `<class 'outlines.types.json_schema_utils.By User'>` |
| `json_schema_dict_to_typeddict` | `NotRequired[outlines.types.json_schema_utils.By User]` |

```python
"By User".isidentifier()   # False
```

The generated model has no fields, so nothing in the original mapping's value type survives either — `int` is gone.

`List[str]` in the same model round-trips correctly to `Optional[List[str]]`, so this is specific to the `additionalProperties` shape.

`outlines` @ 13fe8dd2 (main), Python 3.10.12, pydantic 2.12.5

### Context for the maintainers:

`schema_type_to_python`'s `t == "object"` branch goes straight to `json_schema_dict_to_{pydantic,typeddict,dataclass}` without looking at `additionalProperties`. Reading it first — mapping `{"type": "object", "additionalProperties": S}` with no `properties` to `Dict[str, schema_type_to_python(S)]` — would cover it.

Filing this separately because #1908, #1949, #1951 and #1963 are all open against `schema_type_to_python` and none of them touch `additionalProperties`; I checked each of them against this repro.

ErenAta16:
Closes #1969.

`schema_type_to_python` routes every `{"type": "object"}` to one of the `json_schema_dict_to_*` builders, so the shape Pydantic emits for `Dict[str, V]` is turned into a model:

```python
class Scores(BaseModel):
    by_user: Dict[str, int] = {}

Scores.model_json_schema()["properties"]["by_user"]
# {"additionalProperties": {"type": "integer"}, "default": {}, "title": "By User", "type": "object"}
```

| converter | before | after |
|---|---|---|
| `json_schema_dict_to_pydantic` | `Optional[...By User]` | `Optional[Dict[str, int]]` |
| `json_schema_dict_to_typeddict` | `NotRequired[...By User]` | `NotRequired[Dict[str, int]]` |
| `json_schema_dict_to_dataclass` | `<class '...By User'>` | `Dict[str, int]` |

The generated class had no fields, so `int` was gone as well, and its name came from `title` — `"By User".isidentifier()` is `False`.

The change detects the mapping shape (`additionalProperties` present, `properties` absent) and returns `Dict[str, V]`. A boolean `additionalProperties` carries no value type, so `True`/`False` both give `Dict[str, Any]`.

Objects that declare `properties` keep going down the model path, including the `properties` + `additionalProperties: False` combination Pydantic emits for `extra="forbid"` — there's a test pinning that so the branch order can't be flipped later.

Tests in `tests/types/test_json_schema_utils.py`: one per converter, a parametrised pair for boolean `additionalProperties`, an end-to-end round trip from a real `BaseModel`, and the `properties`-wins case. Stashing only `json_schema_utils.py` gives `6 failed, 24 passed`; with the change, `30 passed`.

`ruff --config=pyproject.toml` on ruff 0.16.0 reports 35 findings on these two files before the change and 46 after — the 11 new ones are all `UP006`/`UP035` for `Dict[...]`, the same rules the file already trips on its existing `Dict`/`List`/`Optional` usage. I kept the existing style rather than mixing `dict[...]` into it; happy to switch both if you'd rather move the file over.

Separately, the four open PRs against `schema_type_to_python` (#1908, #1949, #1951, #1963) all leave `additionalProperties` alone — I ran this repro against each of them — so this shouldn't collide with any of them.

### DIFF
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -78,12 +78,14 @@ def schema_type_to_python(
         # class with no fields and drop the value type, so map it to `Dict[str, V]`.
         if "properties" not in schema and "additionalProperties" in schema:
             additional = schema["additionalProperties"]
-            if additional is True:
-                return Dict[str, Any]  # type: ignore
-            if additional is False:
-                return Dict[str, Any]  # type: ignore
-            value_type = schema_type_to_python(additional, caller_target_type)
-            return Dict[str, value_type]  # type: ignore
+            # `additionalProperties: false` is the opposite of a free-form mapping: it
+            # forbids every key, so the only valid instance is `{}`. Fall through to the
+            # model path, which builds a field-less model, rather than widening it.
+            if additional is not False:
+                if additional is True:
+                    return Dict[str, Any]  # type: ignore
+                value_type = schema_type_to_python(additional, caller_target_type)
+                return Dict[str, value_type]  # type: ignore
 
         name = schema.get("title")
         if caller_target_type == "pydantic":
--- a/src/outlines/types/json_schema_utils.py
+++ b/src/outlines/types/json_schema_utils.py
@@ -73,6 +73,18 @@ def schema_type_to_python(
             item_type = Any
         return List[item_type]  # type: ignore
     elif t == "object":
+        # A mapping is an object with `additionalProperties` and no `properties`; that's
+        # what Pydantic emits for `Dict[str, V]`. Building a model out of it would give a
+        # class with no fields and drop the value type, so map it to `Dict[str, V]`.
+        if "properties" not in schema and "additionalProperties" in schema:
+            additional = schema["additionalProperties"]
+            if additional is True:
+                return Dict[str, Any]  # type: ignore
+            if additional is False:
+                return Dict[str, Any]  # type: ignore
+            value_type = schema_type_to_python(additional, caller_target_type)
+            return Dict[str, value_type]  # type: ignore
+
         name = schema.get("title")
         if caller_target_type == "pydantic":
             return json_schema_dict_to_pydantic(schema, name)
--- a/tests/types/test_json_schema_utils.py
+++ b/tests/types/test_json_schema_utils.py
@@ -1,7 +1,9 @@
+import dataclasses
 import sys
 from dataclasses import is_dataclass
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union, get_type_hints
 
+import pytest
 from pydantic import BaseModel, TypeAdapter
 from pydantic_core import PydanticUndefined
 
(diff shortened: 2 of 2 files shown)


ErenAta16:
Force-pushed: the branch was stacked on the commits from #1917 (closed), so the diff was carrying `dsl.py` / `test_dsl.py` changes that don't belong here, and an earlier push had swept a `.mypy_cache` directory into the commit. Rebased onto `be2cd151`; the diff is now the two intended files only.

Also fixed the style failure — `assert nested.model_fields["v"].annotation == int` tripped `E721` on the pinned ruff 0.9.1, now `is int`. `ruff check --config=pyproject.toml` is clean on both files, and `tests/types/test_json_schema_utils.py` is 30 passed.

Sorry for the noise in the earlier revisions.

src/outlines/types/json_schema_utils.py
```diff
@@ -73,6 +73,18 @@ def schema_type_to_python(
...
         return List[item_type]  # type: ignore
     elif t == "object":
+        # A mapping is an object with `additionalProperties` and no `properties`; that's
+        # what Pydantic emits for `Dict[str, V]`. Building a model out of it would give a
+        # class with no fields and drop the value type, so map it to `Dict[str, V]`.
+        if "properties" not in schema and "additionalProperties" in schema:
+            additional = schema["additionalProperties"]
+            if additional is True:
+                return Dict[str, Any]  # type: ignore
+            if additional is False:
```
Sanjays2402: `additionalProperties: false` with no `properties` is the exact opposite of a free-form mapping: it says the object has no keys at all, and mapping it to `Dict[str, Any]` widens it to accept anything. Either treat it as a plain empty model or at least skip the mapping branch when `additional is False`.

ErenAta16:
You're right and I've pushed the fix. `{"type": "object", "additionalProperties": false}` accepts `{}` and nothing else:

```
{}           gecerli: True
{'a': 1}     gecerli: False
{'x': 'y'}   gecerli: False
```

while `Dict[str, Any]` serialises back to `{"type": "object", "additionalProperties": true}` — so I had it mapping to the exact inverse of the input schema.

The `additional is False` branch now falls through to the model path instead of returning early, which builds a field-less model:

```
schema_type_to_python({"type":"object","additionalProperties":False}, "pydantic")
  before: typing.Dict[str, typing.Any]
  after:  <class 'Empty'>   # model_fields == []
```

`additionalProperties: true` still returns `Dict[str, Any]` and a typed `additionalProperties` still returns `Dict[str, V]`.

I split the old parametrised test, because `[True, False]` asserting the same annotation for both was what let this through in the first place. `test_additional_properties_false_is_not_a_mapping` fails on the previous revision with `assert Dict[str, Any] != Dict[str, Any]` and passes now; the file is at 30 passed.

One thing worth recording for whoever reads this later: Pydantic never emits this shape. `extra="forbid"` on a field-less model produces `{"additionalProperties": false, "properties": {}, ...}` — `properties` is present, so the `"properties" not in schema` guard doesn't fire and it goes to the model path regardless. The branch is only reachable from a hand-written schema, which is presumably why the round-trip tests didn't catch it.

`ruff check` is clean on both files. `ruff format --check` wants to reformat them, but it does so on the unmodified branch head too and `.pre-commit-config.yaml` only runs `ruff`, not `ruff-format`, so I left that alone.

src/outlines/types/json_schema_utils.py
ErenAta16: @C:\Users\erena\AppData\Local\Temp\claude\C--Users-erena-Desktop-Ferrari\21e02430-bf99-467a-9c89-cca4dc31ce6d\scratchpad\o1970_reply.md

src/outlines/types/json_schema_utils.py
```diff
@@ -73,6 +73,20 @@ def schema_type_to_python(
             item_type = Any
         return List[item_type]  # type: ignore
     elif t == "object":
+        # A mapping is an object with `additionalProperties` and no `properties`; that's
+        # what Pydantic emits for `Dict[str, V]`. Building a model out of it would give a
+        # class with no fields and drop the value type, so map it to `Dict[str, V]`.
+        if "properties" not in schema and "additionalProperties" in schema:
```
Sanjays2402, line 79: this only kicks in for a nested schema, since dsl.py's to_target_type calls json_schema_dict_to_pydantic/typeddict/dataclass on the root dict directly and never goes through schema_type_to_python. so a top-level `{"type": "object", "additionalProperties": {"type": "integer"}}` still comes out as a field-less AnonymousPydanticModel, which is the same thing you're fixing here. worth routing the root through the mapping check too?

ErenAta16:
@Sanjays2402 you're right, and it is worse at the root than the case this PR fixes. Sorry for the slow reply, I missed the second comment.

Confirmed against this branch. `dsl.py:424` calls `json_schema_dict_to_pydantic(schema_dict)` on the root directly, and that function reads only `properties`, so the mapping check in `schema_type_to_python` never runs:

```
schema: {"type": "object", "additionalProperties": {"type": "integer"}}

nested (schema_type_to_python)   -> typing.Dict[str, int]        <- fixed here
root   (json_schema_dict_to_*)   -> AnonymousPydanticModel, fields=[]
```

And the value type is not merely dropped, the model silently swallows the data:

```
M = json_schema_dict_to_pydantic(schema)
M().model_dump()            -> {}
M(a=1).model_dump()         -> {}          # accepted, then discarded
M(a="x").model_dump()       -> {}          # no validation at all
M.model_json_schema()       -> {"properties": {}, "title": "...", "type": "object"}
```

So the round trip turns `additionalProperties: {"type": "integer"}` into an object with no properties and no constraint on values. All three target types have it, not just pydantic:

```
typeddict -> AnonymousTypedDict  keys=[]
dataclass -> AnonymousDataclass  fields=[]
```

**Where it gets awkward, and why I don't think it belongs in this PR.** Routing the root through `schema_type_to_python` returns `typing.Dict[str, int]`, and `to_target_type` declares

```python
) -> Union[str, dict, type[BaseModel], _TypedDictMeta, type, SchemaBuilder]:
```

A `_GenericAlias` is none of those, so a caller asking for `"pydantic"` would get something that is not a model class. That is a contract change rather than a bug fix.

For pydantic there is a clean answer that keeps the contract, `RootModel`:

```python
>>> M = RootModel[Dict[str, int]]
>>> issubclass(M, BaseModel)
True
>>> M.model_json_schema()
{"additionalProperties": {"type": "integer"}, "type": "object"}   # exact round trip
>>> M({"a": 2, "b": 3}).model_dump()
{"a": 2, "b": 3}
>>> M({"a": "x"})
ValidationError
```

and `RootModel[Dict[str, Any]]` emits `{"additionalProperties": true, "type": "object"}` for the free-form case, also exact.

For typeddict and dataclass there is no equivalent. A root-level mapping is not expressible as either, so the honest options are to raise and let the existing handler in `to_target_type` do its thing:

```python
except Exception as e:  # pragma: no cover
    warnings.warn(f"Cannot convert schema type {type(schema)} to {target_type}: {e}")
    continue
```

which already exists for exactly this "this target cannot represent this schema" situation, or to leave them as they are.

My preference is to keep this PR on the nested case it already fixes and tests, and open a separate issue for the root path, since it needs a decision on the `RootModel` return and on what typeddict/dataclass should do. Happy to write that issue up with the measurements above, or to extend this PR instead if @RobinPicard would rather see it in one change. Say which and I'll do it.

Issue #1971, opened 2026-07-31
Optional[T] generates Python None instead of JSON null inside DSL containers


ErenAta16:
### Describe the issue as clearly as possible

`Optional[T]` and `Union[T, None]` inside a DSL container generate the Python spelling `None` rather than the JSON spelling `null`, so the resulting regex accepts documents `json.loads` rejects and rejects the ones it accepts.

This is the same class of bug as #1942 (booleans generating `True`/`False` instead of `true`/`false`) and #1960 (date/time not being quoted), and it is the last of the three JSON literals still using Python syntax.

### Steps/code to reproduce the bug

```python
import json, re
from typing import List, Optional
from outlines.types.dsl import python_types_to_terms, to_regex

rx = to_regex(python_types_to_terms(List[Optional[int]]))
print(rx)

for s in ["[1]", "[None]", "[null]", "[1, None]", "[1, null]"]:
    matches = re.fullmatch(rx, s) is not None
    try:
        json.loads(s)
        valid_json = True
    except Exception:
        valid_json = False
    print(f"{s:12} matches={matches!s:5} valid_json={valid_json}")
```

### Expected result

The generated pattern accepts `[null]` and rejects `[None]`, matching what a JSON parser accepts.

### Error message

No exception. The generation is silently wrong:

```
\[(([+-]?(0|[1-9][0-9]*))|(None))(,\ (([+-]?(0|[1-9][0-9]*))|(None)))*\]

[1]          matches=True  valid_json=True
[None]       matches=True  valid_json=False
[null]       matches=False valid_json=True
[1, None]    matches=True  valid_json=False
[1, null]    matches=False valid_json=True
```

Every row where the two columns disagree is a case where constrained generation is steering the model toward output that cannot be parsed. `[null]`, the only correct spelling, is the one the pattern forbids.

The same holds for the other containers and for the explicit union spelling:

```
List[Optional[str]]        ["a"]=ok    [null]=rejected   [None]=accepted
List[Optional[bool]]       [True]=ok   [null]=rejected   [None]=accepted
Dict[str, Optional[int]]   {"k":1}=ok  {"k":null}=rejected  {"k":None}=accepted
List[Union[int, None]]     [1]=ok      [null]=rejected   [None]=accepted
```

### Outlines/Python version information

Reproduced on `main` at `be2cd151855c64a81262c4daace2428400b109ff`, Python 3.12, Windows.

### Context for solving the bug

[...]

PR #1974, opened 2026-08-01
fix(types): render Optional/Union None as JSON null inside DSL containers

feiiiiii5:
## Root Cause

Inside a DSL container (`list[Optional[int]]`, `dict[str, Optional[int]]`, `tuple[Optional[int], ...]`, or the explicit `Union[T, None]` spelling), the `None` member is emitted as a bare `Regex("None")` term. The generated regex therefore accepts the Python spelling `[None]` and **rejects the only valid JSON spelling `[null]`** — constrained generation steers the model toward output that `json.loads` rejects (issue #1971):

```
\[(([+-]?(0|[1-9][0-9]*))|(None))(,\ (([+-]?(0|[1-9][0-9]*))|(None)))*\]

[1]          matches=True  valid_json=True
[None]       matches=True  valid_json=False
[null]       matches=False valid_json=True
```

This is the `None` counterpart of #1942 (booleans → `True`/`False`) and #1960 (temporal values unquoted): the JSON literal spelling inside containers.

## Fix

Introduce a dedicated `NoneLiteral` term (mirroring the `BooleanLiteral` approach used for booleans):

- `_handle_union` appends `NoneLiteral()` instead of `Regex("None")`, keeping the standalone spelling `None` (top-level `Optional[int]`) for backward compatibility.
- `_ensure_json_quoted` renders `NoneLiteral` as `Regex("null")` when the union is nested inside a list/dict/tuple container, so containers emit valid JSON.
- `to_regex` renders the standalone form as `None`.

A genuine `Literal["None"]` string is unaffected and stays JSON-quoted as before.

## Test

- `test_e2e_optional_none_renders_json_null_in_containers` — list/dict/tuple now match `null` and reject `None`; standalone `Optional[int]` still matches the bare `None` keyword and refuses `null` (pinned by a negative assertion added in `5b4c1f9`); `Literal["None"]` still quoted; explicit `Union[int, None]` behaves the same.
- Updated the three `_handle_union` unit tests to expect `NoneLiteral()`.
- Full `tests/types` suite: **286 passed** (incl. `test_custom_types.py` 173).

## Diff scope

2 files, +52/-15: `src/outlines/types/dsl.py`, `tests/types/test_dsl.py`.

## Related PRs (worth landing as a set)

- #1947 — bool literals render as JSON `true`/`false` in containers.
- #1961 — temporal values JSON-quoted in containers.

This PR covers only the `None` literal. Until #1947/#1961 merge, container output can mix Python and JSON spellings across different value types, so any changelog wording about "None renders as JSON null" should stay scoped to containers and be read together with those two.

## AI Disclosure

AI-assisted root-cause analysis, initial draft, and test scaffolding; human review of the term-design decision, backward-compatibility scope, and test coverage.

## Not PR-related failures

None; the full types suite passes locally.


ErenAta16 (commented):
I filed #1971, so this is the fix for my report. Ran it against `6d04c86` rather than reading the diff, and it does what it says:

```
                              [1]    [null]   [None]   [1, null]
List[Optional[int]]            Y       Y        n          Y
List[Optional[str]]            Y       Y        n
Dict[str, Optional[int]]       {"k":1} Y   {"k":null} Y   {"k":None} n
List[Union[int, None]]         Y       Y        n
```

and the standalone case is preserved, which was the part I was most concerned about when I wrote the issue:

```
to_regex(python_types_to_terms(Optional[int]))   (([+-]?(0|[1-9][0-9]*))|None)
```

Introducing `NoneLiteral` rather than special-casing a `Regex("None")` is the better call and I want to say so explicitly, because the alternative was on the table. Matching a bare `Regex` by pattern would also catch a user's own term that happens to spell `None`, and `Regex.__eq__` being pattern-based makes that a real collision rather than a theoretical one. A distinct node carries the intent instead of trying to recover it downstream, and it means the container-context decision lives in one place.

Nesting and the variadic tuple both work:

```
List[List[Optional[int]]]     [[null]] Y    [[None]] n
Tuple[Optional[int], ...]     (null)   Y    (None)   n
```

**One thing worth knowing before this lands, not a defect in this PR.** It makes the emitted document half-JSON while the sibling spelling fixes are still open. On this branch:

```
List[Optional[bool]]    [null] accepted    [None] rejected
                        [True] accepted    [true] rejected
```

so a `list[Optional[bool]]` now generates things like `[True, null]`, mixing the Python spelling of one keyword with the JSON spelling of another in the same array. `json.loads` still rejects it, just for a different reason than before. Same story for temporal values, since `List[Optional[date]]` rejects `["2024-01-01"]` here.

That is #1947 and #1961 respectively and neither is merged yet, so nothing to change here. It does mean the three are worth landing as a set rather than spread out, and that a changelog entry saying "None now renders as JSON null" would be misleading on its own until the other two are in. Worth a line in the description pointing at both.

**Smaller note on the `quote_regex` branch.** Wrapping `null` in quotes for the dict-key path gives:

```
Dict[Optional[int], str]    {"null":"a"} accepted    {null:"a"} rejected
```

Quoting is right, JSON keys are always strings. But it means `Optional[int]` as a key type produces the literal key `"null"`, which is a slightly surprising thing for a schema to permit and I could not tell from the diff whether it was deliberate. If it is, a one-line comment would save the next reader the same detour. If the intended answer is that `None` should not be a valid key at all, that is a different change and not one I would ask for here.

Test coverage looks right for the container cases. If you add anything, the assertion I would want is a negative on the standalone path, pinning that `Optional[int]` outside a container still refuses `null`, since that is the backward-compatibility promise in the docstring and nothing currently fails if it regresses.

Thanks for picking this up.

feiiiiii5:
Thanks for running the matrix against `6d04c86` — glad the behavior matches the report. #1975 has been un-stacked so it no longer carries this PR's `NoneLiteral` block; happy to adjust anything else if the review turns up issues.

feiiiiii5:
Thanks for the detailed verification — both follow-ups are in:

1. **Standalone backward-compat pin** (commit `5b4c1f9`): `test_e2e_optional_none_renders_json_null_in_containers` now asserts `not _re.fullmatch(standalone, "null")`, so a regression that lets JSON `null` leak into the standalone spelling fails the suite.
2. **quote_regex intent documented**: the dict-key branch in `_ensure_json_quoted` now has a one-line comment explaining that JSON keys must be strings, hence `null` is quoted as `"null"`.
3. **Description**: added a "Related PRs" section pointing at #1947/#1961 and noting they should land as a set (changelog wording about None→null stays scoped to containers).

Local: `pytest tests/types/test_dsl.py` → 61 passed. Full `tests/types` suite: 286 passed.

ErenAta16 (commented):
Both follow-ups check out.

The standalone pin is the one I cared about:

```python
assert _re.fullmatch(standalone, "None")
assert not _re.fullmatch(standalone, '"None"')
assert not _re.fullmatch(standalone, "null")
```

so the bare Python keyword stays the standalone spelling and `null` is explicitly rejected there. Without that third line, a later change that made the container spelling leak outward would pass the suite, since the container assertions alone cannot tell the two contexts apart.

Renaming to `test_e2e_optional_none_renders_json_null_in_containers` is worth more than it looks. The old name said the None was "not quoted", which described the old bug rather than the new contract, and the flip from `[None]` to `[null]` in the assertions is exactly the behaviour the name now states.

Nothing further from me.

feiiiiii5:
@ErenAta16 gentle ping — both follow-ups from your review are in (`5b4c1f9` pins the standalone backward-compat case, plus the email-parameterization alignment), and CI is green. Anything else you'd like adjusted before merge?

ErenAta16 (commented):
Nothing further, and nothing has changed since I said so: the head is still `5b4c1f92`, which is the commit I signed off on 3 August. Sorry for leaving the ping sitting.

One thing worth adding now that it is relevant. This is one of three PRs fixing the same family of bugs, the JSON literals that the DSL still spells in Python inside containers:

```
#1961  date / time / datetime   quoting
#1974  Optional / Union None    null
#1947  bool                     true / false
```

All three take the same shape, which is the part I think matters: a distinct term type, container context adapts it, standalone spelling preserved. Your `5b4c1f92` is the clearest statement of that contract, because it pins all three directions at once:

```python
assert _re.fullmatch(standalone, "None")
assert not _re.fullmatch(standalone, '"None"')
assert not _re.fullmatch(standalone, "null")
```

A fourth PR for the boolean half, #2005, landed today and takes the other route: it redefines `types.boolean` to lowercase globally and inverts the assertions in `test_custom_types.py` that pinned the Python spelling. I have said the same over there. Raising it here only because if #2005 goes in as it stands, the family stops being consistent: `None` and the date types would keep their standalone spelling and `bool` would not, which is a worse place to end up than either convention applied uniformly.

So from my side this is ready, and I would rather see it and #1961 merge on their current shape than have the boolean decision reshape them.

feiiiiii5:
Thanks for confirming and for the cross-PR consistency call-out — agreed that `None` and the date types keeping their standalone Python spelling while `bool` drops it would be the worst of both worlds. No changes needed on my side; happy to keep #1974 (and #1961) on the current shape. (The style-check CI failures on this PR are the mypy/urllib3 environment issue fixed by #2009.)

ErenAta16:
Glad that landed the way I meant it, and you read the CI right. I dug into that one on #1951 afterwards and it holds up: the job runs `pre-commit run --all-files`, the mypy hook's isolated env is `[types-tqdm, types-Pillow]` with no `urllib3`, and `src/outlines/exceptions.py:352` imports it, so it is red on every PR that does not carry the one-line config change. Everything merged through 7 August was green, and the only two PRs passing it now are #2003 and #2009, which are that change. So nothing to do on your side, it clears when either of those lands.

Nothing further from me here.

feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

feiiiiii5:
Reopening — the 08-28 close was a mistake on my side; this thread was still in review.

All three points from @ErenAta16 are addressed on-branch, and his 08-16 reply confirmed the CI reading on the `pre-commit run --all-files` job. It ended as COMMENTED, so this is waiting on a merge call rather than on more work from me.

Sorry for the noise.

ErenAta16 (commented):
No noise from my side, and nothing outstanding from the review. Both follow-ups
were confirmed on 08-03, including the standalone pin, so the three points are
closed and I agree this is waiting on a merge call rather than on you.

The one thing I should update is my own sequencing note, because it has gone stale
since I wrote it. On 08-01 I said this lands best alongside #1947 and #1961, since
otherwise a `list[Optional[bool]]` emits `[True, null]` and mixes the two
spellings. #1961 was closed on 08-22 by its author in favour of #2015, which covers
the temporal terms as part of a wider set, so the companion list today is:

```
booleans          #1947 (07-21, unchanged since 08-01)  or  #2005 (08-15)
string-shaped     #1975 (08-01)  or  #2015 (08-17)
None -> null      this PR
```

Both of the first two rows are duplicate pairs rather than dependencies, so nothing
here is blocked on picking a winner. It only matters for the changelog: "None now
renders as JSON null" is accurate for this branch on its own, and the broader claim
that DSL containers emit valid JSON is not true until one branch from each of the
other two rows is in.

For what it is worth I would not hold this one for them. It is the smallest of the
three changes, it is the only one in its row, and the `NoneLiteral` node it
introduces is the part the other branches would have had to invent anyway.

feiiiiii5:
Gentle ping @ErenAta16 - the three follow-ups from your review are addressed on-branch and CI is green. Let me know if anything else needs changing.

ErenAta16 (commented):
Nothing else needed from you. I tried to send this as an approving review rather
than a fourth comment, and learned something worth passing on: GitHub refused it
with `Can not approve a pull request without explicit repository access`. So no
amount of reviewing from outside the org can satisfy the gate this branch is
sitting behind. That is a maintainer action and always was, which is worth knowing
before you spend more time wondering what is still missing on your side.

Checked out `df2a84f` and ran it rather than re-reading the diff:

```
list[Optional[int]]        [null]=True    [None]=False
list[Optional[str]]        [null]=True    [None]=False
list[Union[int, None]]     [null]=True    [None]=False
tuple[Optional[int], ...]  (null)=True    (None)=False
```

The backward-compatibility pin I asked for on 1 August is intact, which was the
part I cared about most:

```
Optional[int] -> (([+-]?(0|[1-9][0-9]*))|None)
  None  True     null  False     "None"  False     42  True
```

And the dict-key path behaves the way the comment now says it does, with the
quoting correct because JSON keys are always strings:

```
{"null":"a"}  True     {null:"a"}  False     {"1":"a"}  True
```

`tests/types/test_dsl.py` gives 60 passed on this branch against 58 on `main`, so
the two new cases are the difference and nothing existing moved. The two
`PermissionError` failures in that file are present on `main` too, they are a
Windows tempfile artefact and nothing to do with you. CI here is 6 check runs,
all green.

So from my side this is done. `mergeable_state` is `blocked`, which is a
maintainer gate rather than anything on the branch, and the three points from my
first review were closed on 3 August. The sequencing note in my 1 September
comment still holds and is the only thing left worth reading before merge: this
one is the smallest of the DSL fixes, it is the only branch in its row, and the
`NoneLiteral` node it introduces is the piece the other branches would have had to
invent anyway.

Sorry for the four-round wait on a change that was ready in early August.

feiiiiii5:
@ErenAta16 gentle nudge: the three follow-ups from your review are addressed on-branch and CI is green. Is there anything else you would like changed, or is this ready for a merge call? Thanks!

ErenAta16:
Re-checked on the current head rather than answering from memory, since the branch moved after my 3 August sign-off. `df2a84fa` is a merge of `main`, and the PR's own diff is still only `src/outlines/types/dsl.py +30/-4` and `tests/types/test_dsl.py +36/-11`, so the merge did not quietly reshape what I looked at.

Same table I ran on 1 August, re-run against `df2a84fa`:

```
List[Optional[int]]        [1]=Y      [null]=Y      [None]=n      [1, null]=Y
List[Optional[str]]        ["a"]=Y    [null]=Y      [None]=n
Dict[str, Optional[int]]   {"k":1}=Y  {"k":null}=Y  {"k":None}=n
List[Union[int, None]]     [1]=Y      [null]=Y      [None]=n

standalone Optional[int]  ->  (([+-]?(0|[1-9][0-9]*))|None)
```

Identical to August, including the standalone spelling, which was the part I cared about most when I filed #1971. `tests/types/test_dsl.py` is 60 passed on the branch against 58 on `main`, the difference being the two tests you added. Two `PermissionError` failures show up in `test_dsl_cfg_from_file` and `test_dsl_json_schema_from_file`, but they fail identically on `main`, so they are a Windows temp-file quirk on my machine rather than anything of yours.

The CI point from August resolved itself the way I guessed it would: #2003 landed on 24 August, your 28 August merge of `main` picked it up, and `Check the code style` is green now along with the other five checks.

On "is this ready for a merge call", the honest answer is that it has been ready since August, and what stands between it and a merge is not something either of us can supply. The PR's `reviewDecision` is `REVIEW_REQUIRED`, which only a write-access approval clears. Mine cannot be one: this repo does not accept an approving review from an outside contributor, so everything I have left here is a `COMMENTED` review regardless of how positive it reads. That is worth knowing because it changes the ask from "please review this" to "please click approve on something already reviewed", which is a much smaller thing to put in front of a maintainer.

Nothing further from me on the code.

feiiiiii5:
Thanks for re-running the matrix against `df2a84fa` rather than replying from memory, @ErenAta16 — and for spelling out why the remaining gap is procedural rather than technical.

@RobinPicard this needs one click, not one more review:

- It fixes #1971, which the reporter filed and has now verified twice on this branch (1 August, and again today on the current head). Their note today: behaviour is identical to what they measured in August, including the standalone `Optional[int]` to `(([+-]?(0|[1-9][0-9]*))|None)` spelling that motivated the issue.
- `tests/types/test_dsl.py` is 60 passed on the branch vs 58 on `main`; the delta is the two tests added here.
- Diff is `src/outlines/types/dsl.py` +30/-4 and `tests/types/test_dsl.py` +36/-11. The merge of `main` did not change the diff.
- All checks are green. The `Check the code style` redness from August was the missing `urllib3` in the mypy hook env, cleared when #2003 landed and picked up by the 28 August merge of `main`.

`reviewDecision` is still `REVIEW_REQUIRED`, and an approving review from an outside contributor does not count toward it here, so @ErenAta16's sign-off cannot be what unblocks this. Happy to take any change you'd like, or to squash anything you prefer differently.

PR #1984, opened 2026-08-07
fix: include whitespace_pattern in JsonSchema equality

### LINKED ISSUE #1981: dsl: JsonSchema.__eq__ ignores whitespace_pattern, causing stale cache hits
`JsonSchema.__eq__` (lines 447–455) compares only the parsed schema JSON, never `whitespace_pattern`:

```python
def __eq__(self, other):
    if not isinstance(other, JsonSchema):
        return False
    try:
        self_dict = json.loads(self.schema)
        other_dict = json.loads(other.schema)
        return self_dict == other_dict   # whitespace_pattern not compared
    except json.JSONDecodeError:
        return self.schema == other.schema
```

`whitespace_pattern` materially changes the regex produced by `to_regex`. Any code that caches or deduplicates logits processors keyed on `JsonSchema` equality (e.g. via `@cache`, a `dict`, or a set) will serve a processor built for the first `whitespace_pattern` when the same schema is reused with a different pattern — silently producing incorrectly constrained output.

**Minimum example:**

```python

schema = '{"type": "string"}'
j1 = JsonSchema(schema, whitespace_pattern=r"[ ]?")
j2 = JsonSchema(schema, whitespace_pattern=r"[\n\t ]*")

assert j1 == j2   # True — wrong; patterns differ
```

`CFG.__eq__` correctly compares its single attribute. `JsonSchema.__eq__` should do the same for `whitespace_pattern`.

**Fix:** add `and self.whitespace_pattern == other.whitespace_pattern` to the equality check.

mikemikimike:
Fixes #1981

## Summary

- Include `whitespace_pattern` in `JsonSchema` equality comparisons.
- Add regression coverage for matching, differing, and `None` patterns.

## Validation

- `uv run --frozen --with genson --with pytest pytest tests/types/test_dsl.py -q -k "not test_dsl_cfg_from_file and not test_dsl_json_schema_from_file"` ? 58 passed.
- `uv run --frozen --with pre-commit pre-commit run ruff --files src/outlines/types/dsl.py tests/types/test_dsl.py` ? passed.
- `uv run --frozen --with mypy --with urllib3 mypy --allow-redefinition src/outlines/types/dsl.py tests/types/test_dsl.py` ? no issues found.
- `git diff --check` ? passed.

The two file-based tests in the full `tests/types/test_dsl.py` run remain blocked by the repository's existing Windows `NamedTemporaryFile` permission behavior; the changed-path tests pass.

### DIFF
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -450,9 +450,15 @@ class JsonSchema(Term):
         try:
             self_dict = json.loads(self.schema)
             other_dict = json.loads(other.schema)
-            return self_dict == other_dict
+            return (
+                self_dict == other_dict
+                and self.whitespace_pattern == other.whitespace_pattern
+            )
         except json.JSONDecodeError:  # pragma: no cover
-            return self.schema == other.schema
+            return (
+                self.schema == other.schema
+                and self.whitespace_pattern == other.whitespace_pattern
+            )
 
     @classmethod
     def from_file(cls, path: str) -> "JsonSchema":
--- a/tests/types/test_dsl.py
+++ b/tests/types/test_dsl.py
@@ -533,6 +533,18 @@ def test_dsl_json_schema_from_file():
         assert schema == JsonSchema(schema_content)
 
 
+def test_json_schema_equality_includes_whitespace_pattern():
+    schema = {"type": "string"}
+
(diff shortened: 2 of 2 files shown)


RobinPicard (approved):
Looks good, thanks!

PR #1987, opened 2026-08-07
fix(mlxlm): check if output_type is None instead of truthiness

### LINKED ISSUE #1980: mlxlm: MLXLMTypeAdapter.format_output_type uses falsy check instead of identity-None check
`MLXLMTypeAdapter.format_output_type` (line 97) uses `if not output_type` instead of `if output_type is None`:

```python
# mlxlm.py:97
def format_output_type(self, output_type=None):
    if not output_type:   # wrong: discards any processor where bool(processor) is False
        return None
    return [output_type]
```

If a logits processor subclass defines `__bool__` returning `False`, the processor is silently discarded and generation proceeds unconstrained — no error raised. Every other model file uses the correct pattern (`transformers.py`, `vllm.py`, `llamacpp.py`, etc.).

**Minimum example:**

```python
class FalsyProcessor:
    def __bool__(self): return False

def format_output_type_buggy(output_type=None):
    if not output_type:       # exact mlxlm.py code
        return None
    return [output_type]

p = FalsyProcessor()
print(p is None)                        # False
print(format_output_type_buggy(p))     # None — processor silently dropped
```

**Fix:** replace `if not output_type:` with `if output_type is None:`, consistent with every other model backend.

Aryan-Pardeshi:
Fixes #1980

### Context
In \src/outlines/models/mlxlm.py\, \ormat_output_type\ previously checked \if not output_type:\. This causes issue #1980 when an \output_type\ object evaluates to \False\ in boolean context (e.g., custom processors/callables with \__bool__\ returning \False\).

### Changes
- Replaced \if not output_type:\ with \if output_type is None:\ in \MLXLMTypeAdapter.format_output_type\, matching the standard pattern across all other model adapters.
- Added unit tests in \	ests/models/test_mlxlm_type_adapter.py\ for both \None\ and falsy processor objects.

### DIFF
--- a/src/outlines/models/mlxlm.py
+++ b/src/outlines/models/mlxlm.py
@@ -94,7 +94,7 @@ class MLXLMTypeAdapter(ModelTypeAdapter):
             The logits processor argument to be passed to the model.
 
         """
-        if not output_type:
+        if output_type is None:
             return None
         return [output_type]
 
--- a/tests/models/test_mlxlm_type_adapter.py
+++ b/tests/models/test_mlxlm_type_adapter.py
@@ -105,29 +105,6 @@ def test_mlxlm_type_adapter_format_input(adapter, image):
         ]))
 
 
-def test_mlxlm_type_adapter_format_output_type_falsy_processor():
-    tokenizer = MagicMock()
-    adapter = MLXLMTypeAdapter(tokenizer=tokenizer)
-
-    class FalsyProcessor:
-        def __bool__(self):
-            return False
-
-    processor = FalsyProcessor()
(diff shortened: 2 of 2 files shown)


RobinPicard (changes requested):
Thanks, the fix itself looks good as it makes `mlxlm` consistent with the other adapters, so happy to merge it.

Could you remove the two added tests though? A falsy processor cannot actually occur here: everything that reaches `format_output_type` is either `None` or a logits processor instance that defines neither `__bool__` nor `__len__`, so the `FalsyProcessor` case is synthetic. I'd rather keep the test suite focused on cases that can happen in practice.

PR #1995, opened 2026-08-13
fix: forward falsy logits processor in mlxlm format_output_type

### LINKED ISSUE #1980: mlxlm: MLXLMTypeAdapter.format_output_type uses falsy check instead of identity-None check

serhiizghama:
Closes #1980. `MLXLMTypeAdapter.format_output_type` used `if not output_type` to decide whether a logits processor was supplied. That's a truthiness check, so any processor whose `__bool__` (or `__len__`) evaluates to `False` gets silently dropped and generation runs unconstrained — no error, just wrong output. Only `None` should mean "no constraint".

Switched to the identity check `if output_type is not None`, matching `transformers.py`, `llamacpp.py` and the other backends.

Added two tests that don't need Apple Silicon (they build the adapter with a mock tokenizer): one for the `None` case and one for a falsy processor. The falsy-processor test fails on the old code and passes on the fix.

### DIFF
--- a/src/outlines/models/mlxlm.py
+++ b/src/outlines/models/mlxlm.py
@@ -94,9 +94,9 @@ class MLXLMTypeAdapter(ModelTypeAdapter):
             The logits processor argument to be passed to the model.
 
         """
-        if not output_type:
-            return None
-        return [output_type]
+        if output_type is not None:
+            return [output_type]
+        return None
 
 
 class MLXLM(Model):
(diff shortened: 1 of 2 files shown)


serhiizghama:
Closing in favour of #1987, which was opened first and covers the same fix.

PR #2000, opened 2026-08-14
fix(types): make JsonSchema and CFG hashable

### LINKED ISSUE #1996: JsonSchema and CFG are unhashable: defining __eq__ without __hash__ sets __hash__ = None
## Bug Description

`JsonSchema` and `CFG` both define `__eq__` without defining `__hash__`. Per the [Python data model](https://docs.python.org/3/reference/datamodel.html#object.__hash__), when a class defines `__eq__` but not `__hash__`, Python silently sets `__hash__ = None`, making instances **unhashable**. This means neither type can be used in a `set` or as a `dict` key.

## Affected files

- `src/outlines/types/dsl.py` — `JsonSchema.__eq__` at line 447 (no `__hash__` defined)
- `src/outlines/types/dsl.py` — `CFG` is a `@dataclass` that defines explicit `__eq__` at line 258, which suppresses the dataclass-generated `__hash__`

## Minimal reproducer

```python
from outlines.types.dsl import JsonSchema, CFG

js = JsonSchema('{"type": "string"}')
print(JsonSchema.__hash__)  # None
hash(js)                    # TypeError: unhashable type: 'JsonSchema'
{js}                        # TypeError: unhashable type: 'JsonSchema'

cfg = CFG("root ::= [0-9]+")
print(CFG.__hash__)         # None
hash(cfg)                   # TypeError: unhashable type: 'CFG'
{cfg: "value"}              # TypeError: unhashable type: 'CFG'
```

## Why this matters

Any code that stores output types in a set or uses them as dict keys will fail at runtime with a `TypeError`. This includes cache key construction, deduplication of output types, and any future caching layer over `JsonSchema` or `CFG` instances.

## Suggested fix

Add `__hash__` implementations consistent with the `__eq__` semantics already defined.

**`JsonSchema`** — hash on normalised JSON (sorted keys) so equivalent schemas hash equally:

```python
def __hash__(self):
    try:
        normalised = json.dumps(json.loads(self.schema), sort_keys=True)
    except json.JSONDecodeError:
        normalised = self.schema
    return hash((normalised, self.whitespace_pattern))
```

**`CFG`** — add `unsafe_hash=True` to the `@dataclass` decorator, or add:

```python
def __hash__(self):
    return hash(self.definition)
```

## Environment

Reproduced on Python 3.12 with Outlines `main` branch (2026-08-13).

eeshsaxena:
Closes #1996.

`JsonSchema` and `CFG` both define `__eq__` without `__hash__`, so Python sets `__hash__ = None` and instances cannot be used in a `set` or as a `dict` key:

```python
hash(JsonSchema('{"type": "string"}'))  # TypeError: unhashable type: 'JsonSchema'
```

`CFG` is a `@dataclass` whose explicit `__eq__` suppresses the dataclass-generated `__hash__`; `JsonSchema` is a plain class that defines only `__eq__`.

This adds `__hash__` to both, kept consistent with `__eq__`:

- `CFG` hashes its `definition` (what `__eq__` compares).
- `JsonSchema.__eq__` compares the *parsed* schema, so `__hash__` hashes the canonical (key-sorted) JSON together with `whitespace_pattern`. Two schemas that differ only in key order or whitespace are equal and now hash equal, so they collapse in a set; non-JSON input falls back to the raw string, matching the `JSONDecodeError` branch of `__eq__`.

I added a regression test in `tests/types/test_dsl.py` covering hashability, set/dict use, and eq/hash consistency for reordered schemas.

One note: I couldn't build `outlines_core` locally (no wheel for my Python version), so I verified the fix by exercising the real `JsonSchema`/`CFG` classes with `outlines_core` stubbed. All the assertions in the new test pass; CI runs the full suite.

### DIFF
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -466,12 +466,9 @@ class JsonSchema(Term):
     def __hash__(self):
         # __eq__ compares the *parsed* schema, so hash the canonical (key-sorted)
         # form: two schemas that differ only in key order or whitespace are equal
-        # and must hash equal. Non-JSON input falls back to the raw string, which
-        # matches the JSONDecodeError branch of __eq__.
-        try:
-            canonical = json.dumps(json.loads(self.schema), sort_keys=True)
-        except json.JSONDecodeError:
-            canonical = self.schema
+        # and must hash equal. __init__ validates self.schema as JSON, so
+        # json.loads cannot fail here.
+        canonical = json.dumps(json.loads(self.schema), sort_keys=True)
         return hash((canonical, self.whitespace_pattern))
 
     @classmethod
--- a/src/outlines/types/dsl.py
+++ b/src/outlines/types/dsl.py
@@ -260,6 +260,9 @@ class CFG(Term):
             return False
         return self.definition == other.definition
 
+    def __hash__(self):
+        return hash(self.definition)
+
     @classmethod
     def from_file(cls, path: str) -> "CFG":
         """Create a CFG instance from a file containing a CFG definition.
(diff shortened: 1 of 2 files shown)


src/outlines/types/dsl.py
```diff
@@ -460,6 +463,17 @@ def __eq__(self, other):
                 and self.whitespace_pattern == other.whitespace_pattern
             )
 
+    def __hash__(self):
+        # __eq__ compares the *parsed* schema, so hash the canonical (key-sorted)
+        # form: two schemas that differ only in key order or whitespace are equal
+        # and must hash equal. Non-JSON input falls back to the raw string, which
+        # matches the JSONDecodeError branch of __eq__.
+        try:
```
RobinPicard: I don't think we need that try/except block here as we already check the JSON validity of the schema at initialization

RobinPicard (changes requested):
Thanks for opening a PR, just a small comment

PR #2005, opened 2026-08-15
fix(types): emit JSON boolean literals (true/false) instead of Python True/False

### LINKED ISSUE #1942: DSL container types render bool as Python True/False, producing invalid JSON and diverging from the JSON-schema path

uuzzrm:
## Summary

The DSL emitted Python-style booleans instead of JSON literals. `types.boolean` matched `True`/`False`, and a bool value (e.g. from `Literal[True]`) became a `Regex` of `str(True)`. Inside container types like `list[bool]` or `list[Literal[True]]` that produced output such as `[True, False]`, which `json.loads` rejects — the JSON-schema path, by contrast, generates `(true|false)`.

This change makes both paths agree:

- `types.boolean` now matches `true`/`false` only.
- `python_types_to_terms` maps a bool value to the lowercase JSON literal, so `Literal[True]` -> `true`, `Literal[False]` -> `false`.

Examples:

```python
>>> from typing import List, Literal
>>> from outlines import types
>>> from outlines.types.dsl import python_types_to_terms, to_regex

>>> types.boolean.pattern
'(true|false)'

>>> to_regex(python_types_to_terms(List[bool]))
'\\[((true|false))(,((true|false)))*\\]'   # was (True|False) before

>>> to_regex(python_types_to_terms(List[Literal[True]]))
'\\[((true))(,((true)))*\\]'               # was (True) before
```

Updated the existing literal-bool test and the `types.boolean` parametrized cases, and added a regression test covering the type-level pattern, bool values, and container output.

## Verification

- `pytest tests/types/` — 346 passed; the only 2 failures are pre-existing Windows temp-file permission errors in `test_dsl_cfg_from_file` / `test_dsl_json_schema_from_file` (they also fail on a clean checkout of `main` here, and don't touch this change).
- `ruff check` on the changed files — clean (repo-pinned 0.9.1).

Closes #1942

Developed with AI assistance and reviewed before submission.


ErenAta16 (commented):
This fixes the container case, and I filed the issue behind it (#1942), so a couple of things worth having on the record before it goes further.

**It changes `types.boolean` itself, not just the container rendering.**

```diff
- boolean = Regex("(True|False)")
+ boolean = Regex("(true|false)")
```

and the existing assertions that pinned that behaviour are inverted to suit:

```diff
- (types.boolean, "True", True),
- (types.boolean, "False", True),
- (types.boolean, "true", False),
+ (types.boolean, "true", True),
+ (types.boolean, "false", True),
+ (types.boolean, "True", False),
+ (types.boolean, "False", False),
```

Those three lines on `main` are the contract, not an accident. I said so explicitly when filing #1942:

> To be clear up front: the standalone `types.boolean` being Python-style is intentional. `tests/types/test_custom_types.py` explicitly asserts `"True"`/`"False"` match and `"true"` does not, so I'm not suggesting changing that type on its own.

The bug I reported is narrower: inside a container, everything else is JSON-shaped (strings get quoted, brackets are JSON) while booleans stay Python-shaped, so the output is neither valid JSON nor consistent with the JSON-schema path. That is a container-context problem. Anyone today generating a bare `outlines` boolean and parsing it with `ast.literal_eval`, or matching it against `True`, breaks silently on upgrade.

I am not arguing lowercase is wrong in the abstract, it is obviously the JSON spelling. I am saying it is a public behaviour change that deserves to be a decision rather than a side effect, and rewriting the assertions that encoded the old contract is what makes it invisible in review.

**There is an existing PR for this.** #1947 has been open since 21 July and solves the same issue without the breakage, by introducing a `BooleanLiteral` term that renders Python-style standalone and JSON-style inside containers:

```python
@dataclass
class BooleanLiteral(Term):
    """A Python boolean literal with context-sensitive JSON rendering."""
```

so `_ensure_json_quoted` adapts it in container context and the standalone type keeps its documented behaviour. It carries two reviews already. It is also the same shape as #1961 and #1974, which handle the date/time and `None` halves of the same family, so the three land as a consistent set.

Which one a maintainer prefers is their call, and if the answer is that `types.boolean` should be lowercase everywhere then this PR is the right approach and #1942 was too conservative. I would just rather that be answered out loud, because the two PRs are not interchangeable: one is a container fix, the other is an API change that happens to fix containers.

@RobinPicard flagging the overlap since #1947, #1961, #1974 and this all touch the same dispatch.

PR #2036, opened 2026-09-14
fix(types): accept ISO 8601 T separator in the datetime type

### LINKED ISSUE #2035: `types.datetime` rejects the ISO 8601 `T` separator (`datetime.isoformat()` output does not match)
## Actual behavior

```python
import datetime
from outlines import types

types.datetime.matches("2022-03-23T10:15:30")  # False
types.datetime.matches(datetime.datetime(2022, 3, 23, 10, 15, 30).isoformat())  # False
types.datetime.matches("2022-03-23 10:15:30")  # True (space separator only)
```

The `datetime` regex requires a whitespace separator between the date and the time:

```python
datetime = Regex(rf"({date.pattern})(\s)({time.pattern})")
```

So the canonical ISO 8601 / RFC 3339 form `YYYY-MM-DDTHH:MM:SS` — what `datetime.isoformat()` produces by default and what models are most likely to emit — is rejected. When `types.datetime` is used as a field type, the generated JSON Schema `pattern` inherits the same restriction, so a constrained model cannot emit the standard form. The JSON Schema path disagrees: a `{"format": "date-time"}` field compiled via `build_regex_from_schema` requires the `T` separator, so the DSL type and the schema path constrain datetimes differently.

## Expected behavior

`types.datetime` should accept the `T` separator (ISO 8601 also permits lowercase `t`), in addition to the space form it already accepts.

## Proposed fix

Change the separator group from `(\s)` to `([Tt\s])`. Fractional seconds and timezone offsets could be added later; this fix only restores the canonical separator.

## Environment

outlines main @ c52af84, Python 3.12, Linux.

betacatsling:
## Summary

Fixes #2035.

`outlines.types.datetime` requires a whitespace character between the date and time components, so it rejects the canonical ISO 8601 / RFC 3339 form that uses `T` — including everything `datetime.datetime.isoformat()` produces by default. A model constrained by `types.datetime` (directly, or as a `pattern` in a generated JSON Schema) cannot emit `2022-03-23T10:15:30`; the same `date-time` handled through the JSON Schema path (`format: "date-time"` in outlines_core) does use the `T` separator, so the two paths disagree.

- `src/outlines/types/__init__.py`: change the `datetime` separator group from `(\s)` to `([Tt\s])`, accepting `T`, `t` and the whitespace forms that were already allowed.
- `tests/types/test_custom_types.py`: add parametrized cases covering the `T`/`t` separators, the previously accepted space form, and negative cases (missing separator, date-only, invalid time, invalid date).

Scope note: fractional seconds and timezone offsets (`Z` / `+hh:mm`) remain unsupported, matching the current `time` regex's fixed `HH:MM:SS` shape.

## Validation

- Before: `pytest tests/types/test_custom_types.py -k "custom_type49 or custom_type50"` (the two new `T`/`t` cases) → **2 failed** (`assert False is True` — `T` separator rejected); the 6 other new cases (space form + negatives) pass.
- After: all 8 new cases pass; `pytest tests/types/` → **354 passed**.
- `pytest tests/types tests/test_cache.py tests/test_templates.py tests/test_inputs.py tests/test_exceptions.py tests/test_utils` → **488 passed, 63 skipped** (env-gated skips; `tests/models`, `tests/backends`, `tests/processors`, `test_applications.py`, `test_generator.py` are uncollectable in this env — `numpy`/`transformers` not installed).
- `ruff check` (pinned v0.9.1) on changed files → clean; `ruff format --diff` shows only pre-existing reformat noise in unrelated sections. `mypy` reports only missing stub packages (PIL/urllib3) in unrelated files.


warter666:
Ran into the same thing via #2035 and opened #2040 with the same patch before seeing this. Closing mine --- one PR is better than two identical ones. Independent confirmation on the diagnosis: it's not just that isoformat() output was rejected, the two paths were mutually exclusive (the DSL regex took whitespace only, DATE_TIME in the JSON Schema path requires T). Verified ([Tt\s]) separately: 347 passed in tests/types, mypy 1.14.1 and ruff 0.9.1 clean, and reverting the regex does make the new cases fail. Nice addition for this PR: a case built from datetime.datetime(...).isoformat() rather than a literal.

